PR #3, opened 2021-01-07
Makefile Setup

kevinaer:
- Added Makefile
- Added doc scripts
- ran styling on .py files


markurtz (dismissed):
LGTM

bnellnm (commented):
Can you double check the merge on setup.py?  I think some code got duplicated but it is hard to tell from the raw diff view.

I'm also not sure how the build process should work considering that the binary bits have to come from somewhere else for the package to be complete.

bnellnm (dismissed):

markurtz (approved):

bnellnm (approved):

PR #10, opened 2021-01-15
Examples, tests, and utilities

mgoin:
First round of examples, integration tests, notebooks, and utilities


bfineran (changes requested):
Looks good overall, comments are mostly nits.  I like the notebooks.

bfineran (approved):
LGTM. I think you may have missed the nits on `src/deepsparse/utils/onnx.py`, but if intentional I'm also fine as is

markurtz (approved):

PR #11, opened 2021-01-16
Ignore failed import for sparsezoo in engine

mgoin:
The engine doesn't have a way to install sparsezoo so we're removing it from engine.py for the moment


markurtz (changes requested):
Just put a try catch around the import. The functionality of this code should 100% be in here for launch.

markurtz (dismissed):
LGTM as long as we update logging for the error with your examples diff

bfineran (approved):
LGTM

markurtz (approved):

PR #14, opened 2021-01-20
Flask server example

mgoin:
Provide example server integration for DeepSparse Engine

### DIFF
--- a/examples/benchmark/check_correctness.py
+++ b/examples/benchmark/check_correctness.py
@@ -33,6 +33,7 @@ python examples/benchmark/check_correctness.py \
 import argparse
 
 import onnxruntime
+
 from deepsparse import compile_model, cpu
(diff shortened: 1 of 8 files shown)


markurtz (commented):
I thought we were going to add in a script to show how to communicate with the server? At the minimum we should have a documented curl request on how to send.

Also, we had talked over potentially sending pb or some other serialization to the server to save network time. Is that still planned?

markurtz (commented):
Should update the isort dependencies so the imports sort correctly to separate python native, third party, and first party into their own sections. 

Other than that, LGTM

jeanniefinks (commented):
Editorial provided; two engineers will want to approve commits.

markurtz (approved):

bfineran (approved):

PR #47, opened 2021-02-21
Add support for batch splitting

bnellnm:


src/deepsparse/engine.py
```diff
@@ -553,7 +574,7 @@ def analyze_model(
     num_cores = _validate_num_cores(num_cores)
     batch_size = _validate_batch_size(batch_size)
     num_sockets = 1
-    eng_net = LIB.deepsparse_engine(model, batch_size, num_cores, num_sockets)
+    eng_net = LIB.deepsparse_engine(model, batch_size, num_cores, num_sockets, True)
```
mgoin, line 577: what is `True` here?

mgoin (commented):
so by default we will run on all sockets with all physical cores?

bnellnm:
This is a bit hacked up and not ready for review but currently, we hardcode
the number of sockets to 1.   I wanted to pass in the actual number of
sockets so the right number would get passed down to the engine
constructor.  The boolean is to force batch splitting (or not).  Ideally,
it would just be enabled whenever num sockets > 1 but that still needs some
work.

PR #48, opened 2021-02-22
add compile engine from SparseZoo stub and benchmarking progress bar

bfineran:
* enable one line SparseZoo integrations
* added an optional progress bar for benchmarking (tqdm should be a nested dependency)

```
benchmark_model(
    "zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenette/base-none",
    sample_input,
    show_progress=True,
)
```

### DIFF
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -62,10 +62,12 @@ def _model_to_path(model: Union[str, Model, File]) -> str:
         raise ValueError("model must be a path, sparsezoo.Model, or sparsezoo.File")
 
     if isinstance(model, str) and model.startswith("zoo:"):
+        # load SparseZoo Model from stub
         if sparsezoo_import_error is not None:
             raise sparsezoo_import_error
-        model = Zoo.load_model_from_stub(model).onnx_file.downloaded_path()
-    elif Model is not object and isinstance(model, Model):
+        model = Zoo.load_model_from_stub(model)
+
+    if Model is not object and isinstance(model, Model):
         # default to the main onnx file for the model
         model = model.onnx_file.downloaded_path()
     elif File is not object and isinstance(model, File):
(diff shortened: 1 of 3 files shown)


markurtz (approved):

mgoin (approved):

bfineran:
blocks https://github.com/neuralmagic/sparseml/pull/81

src/deepsparse/engine.py
```diff
@@ -23,13 +23,19 @@
 import numpy
 
 from deepsparse.benchmark import BenchmarkResults
+from tqdm.auto import tqdm
```
markurtz, line 26: tqdm needs to be added to known third party in the setup.cfg and since we're explicitly importing it here I think we should add it as an explicit dependency in the setup.py for a specific version range

src/deepsparse/engine.py
```diff
@@ -55,8 +61,10 @@ def _model_to_path(model: Union[str, Model, File]) -> str:
     if not model:
         raise ValueError("model must be a path, sparsezoo.Model, or sparsezoo.File")
 
-    if isinstance(model, str):
-        pass
+    if isinstance(model, str) and model.startswith("zoo:"):
+        if sparsezoo_import_error is not None:
+            raise sparsezoo_import_error
+        model = Zoo.load_model_from_stub(model).onnx_file.downloaded_path()
```
markurtz, line 67: NIT: might make sense to have this overwrite the model to the Model instance returned just from load_model_from_stub and then add a new line and start the if statement again. That way we keep from duplicating the onnx_file.downloaded_path() portion in the event we change the zoo model flow and resolving a runnable model in the future

markurtz (changes requested):

mgoin (approved):

markurtz (approved):

PR #50, opened 2021-02-23
Batch splitting

bnellnm:
This adds support for batch-splitting in the engine, i.e. if the number of the sockets on the system is > 1, then the model is split into batch_size/num_sockets smaller models which each run on their own socket.

Batch splitting will be enabled whenever num_sockets >1.


mgoin (dismissed):
LGTM, thanks Bill

markurtz (changes requested):
Need to run make style to format it, other than that looks good

markurtz (approved):

mgoin (approved):

PR #62, opened 2021-03-19
move numpy serialization from examples to API level

bfineran:
there are a few planned / ongoing server examples that can reuse this functionality


markurtz (approved):

mgoin (approved):
thanks for the test!

PR #63, opened 2021-03-22
Fix documentation related to number of sockets

bnellnm:
Some of the documentation wrt. number of sockets was out of date.  These changes should reflect the current state of things.

### DIFF
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -533,7 +533,7 @@ def benchmark_model(
     Gives defaults of batch_size == 1 and num_cores == None
     (will use all physical cores available on a single socket).
 
-    Note 1: Benchmarking is currenlty only supoorted on a single socket.
+    Note 1: Benchmarking is currently only supported on a single socket.
 
     :param model: Either a path to the model's onnx file, a SparseZoo model stub
         prefixed by 'zoo:', a SparseZoo Model object, or a SparseZoo ONNX File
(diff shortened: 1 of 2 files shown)


src/deepsparse/engine.py
```diff
@@ -537,6 +533,8 @@ def benchmark_model(
     Gives defaults of batch_size == 1 and num_cores == None
     (will use all physical cores available on a single socket).
 
+    Note 1: Benchmarking is currenlty only supoorted on a single socket.
```
mgoin, line 536:
```suggestion
    Note 1: Benchmarking is currently only supported on a single socket.
```

src/deepsparse/engine.py
```diff
@@ -592,6 +590,8 @@ def analyze_model(
     Gives defaults of batch_size == 1 and num_cores == None
     (will use all physical cores available on a single socket).
 
+    Note 1: Analysis is currenlty only supoorted on a single socket.
```
mgoin, line 593:
```suggestion
    Note 1: Analysis is currently only supported on a single socket.
```

mgoin (commented):
LGTM just nits

src/deepsparse/engine.py
bnellnm: I really need new glasses :)

mgoin (approved):

bfineran (approved):

PR #70, opened 2021-04-02
Abstract version to be used in setup and docs

markurtz:


bfineran (commented):
There is already a `deepsparse.version.version`, are we planning on keeping both?

markurtz:
> There is already a `deepsparse.version.version`, are we planning on keeping both?

Good catch, forgot we were pulling that up from the backend. Let me see if we can pipe that through

bfineran (approved):

mgoin (approved):

PR #85, opened 2021-04-14
Reformat version.py file

markurtz:
- follow python conventions for __all__ exports
- relative lookup of generated-version.py

### DIFF
--- a/src/deepsparse/version.py
+++ b/src/deepsparse/version.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import os
+import pathlib
 
 
 """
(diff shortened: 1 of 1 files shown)


bnellnm:
Thanks for fixing the path issues. Can you add an export of the splash variable as well? We need that to run dashboard benchmarks.

bnellnm (changes requested):
Thanks for fixing the path issues. Can you add an export of the splash variable as well? We need that to run dashboard benchmarks.

markurtz:
> Thanks for fixing the path issues. Can you add an export of the splash variable as well? We need that to run dashboard benchmarks.

@bnellnm added, let me know if that looks good

bnellnm (approved):
Looks good.  Thanks Mark

mgoin (approved):

PR #90, opened 2021-04-21
Add scheduler as Engine parameter

mgoin:
Exposes scheduler selection as a parameter, specifically to enable "multi-stream" support

### DIFF
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -18,6 +18,7 @@ Code related to interfacing with a Neural Network in the DeepSparse Engine using
 
 import os
 import time
+from enum import Enum
 from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 import numpy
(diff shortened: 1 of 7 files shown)


src/deepsparse/engine.py
```diff
@@ -144,15 +156,21 @@ def __init__(
         batch_size: int,
         num_cores: int,
         num_sockets: int = None,
+        scheduler: str = None,
```
markurtz, line 159: Could we make scheduler an enum so users have to pick one of those values? We can set the enum names to the values needed by the backend and have the call into deepsparse_engine pass the enum.value

mgoin: We have an enum already inside the engine that this immediately translates to. 
It looks like we could wrap it with pybind and expose it, I'll give it a try.
https://stackoverflow.com/questions/47893832/pybind11-global-level-enum

markurtz (dismissed):

markurtz (approved):

bnellnm (approved):
Seems reasonable to me.

PR #91, opened 2021-04-22
Unsupported systems install errors

markurtz:
Add in checks for the install process so that it will fail with a more legible error on unsupported systems. Once this lands, we can begin pushing a generic wheel up along with the many Linux so users will be shown an error message saying DeepSparse is currently not supported on any non-linux systems along with a pointer to our docs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -96,12 +96,12 @@ class OverrideInstall(install):
 
         # unknown system, raise error on install
         raise OSError(
-                f"Unknown OS given of {sys.platform}, "
-                "it is unsupported for DeepSparse. "
-                "Please run on a Linux system. "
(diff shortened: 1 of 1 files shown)


bfineran (approved):

dhuangnm:
How does pip install work when both general and manylinux wheels are there? Will the general wheel get picked instead of the manylinux one on Linux?

markurtz:
> How does pip install work when both general and manylinux wheels are there? Will the general wheel get picked instead of the manylinux one on Linux?

We should test with our internal pypi, but I can't imagine the implementors making it so less specific builds are pulled down for a more specific system.

bnellnm (approved):
LGTM

PR #92, opened 2021-04-22
Fixes for build version on release

markurtz:
Changing it so that docs/conf.py and setup.py conditionally load generated_version.py or version.py based on what's available.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,11 +29,15 @@ author = "Neural Magic"
 # The full version, including alpha/beta/rc tags
 version = "unknown"
 version_major_minor = version
-# load and overwrite version info from deepsparse package
-exec(open(os.path.join(os.pardir, "src", "deepsparse", "version.py")).read())
+
+# load and overwrite version and release info from deepsparse package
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -29,11 +29,15 @@
...
 version_major_minor = version
-# load and overwrite version info from deepsparse package
-exec(open(os.path.join(os.pardir, "src", "deepsparse", "version.py")).read())
+
+# load and overwrite version and release info from deepsparse package
+version_path = os.path.join(os.pardir, "src", "deepsparse", "generated_version.py")
+if not os.path.exists(version_path):
+    version_path = os.path.join(os.pardir, "src", "deepsparse", "version.py")
+exec(open(version_path).read())
 release = version
```
bnellnm, line 38: is this variable supposed to be is_release or release?

bnellnm (approved):
Seems reasonable to me.

mgoin (approved):

docs/conf.py
markurtz: this is supposed to be release, it's a special param for the sphinx build process

PR #96, opened 2021-04-23
update Flask examples

bfineran:
Introduce `engine_flask_server()` and `EngineFlaskClient` that can be used as simple client/server wrapper for the `Engine` class.

### DIFF
--- a/examples/flask/server.py
+++ b/examples/flask/server.py
@@ -48,70 +48,46 @@ import argparse
 import flask
 from flask_cors import CORS
 
-from deepsparse import compile_model
+from deepsparse import Scheduler, compile_model
 from deepsparse.utils import arrays_to_bytes, bytes_to_arrays
 
 
-def parse_args():
-    parser = argparse.ArgumentParser(
-        description=(
-            "Host an ONNX model as a server, using the DeepSparse Engine and Flask"
-        )
-    )
-
-    parser.add_argument(
-        "onnx_filepath",
-        type=str,
-        help="The full filepath of the ONNX model file",
-    )
-
-    parser.add_argument(
-        "-s",
-        "--batch_size",
-        type=int,
-        default=1,
-        help="The batch size to run the analysis for",
-    )
-    parser.add_argument(
-        "-j",
-        "--num_cores",
-        type=int,
-        default=0,
-        help=(
-            "The number of physical cores to run the analysis on, "
-            "defaults to all physical cores available on the system"
-        ),
-    )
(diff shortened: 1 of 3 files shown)


mgoin (dismissed):
LGTM, thanks!

examples/flask/server.py
```diff
@@ -48,10 +48,66 @@
...
+    :param num_sockets: The number of physical sockets to run the model on.
+        Pass None or 0 to run on the max number of sockets for the
+        current machine, default None
+    :param scheduler: The kind of scheduler to execute with. Pass None for the default.
+    :param address: IP address to run on. Default is 0.0.0.0
+    :param port: port to run on. Default is 5543
+    :return: launches a flask server on the given address and port can run the
+        given model on the DeepSparse engine via HTTP requests
+    """
+    print(f"Compiling model at {model_path}")
```
kevinaer, line 82: Could we potentially change the print statements to logs? Specially for the flask API endpoints.

markurtz: agree with Kevin, would be good to get these moved to logs now and potentially add logging into the classes

mgoin (dismissed):

examples/flask/server.py
```diff
@@ -44,12 +52,72 @@
...
+
+_LOGGER = log_init(os.path.basename(__file__))
+
+
+def engine_flask_server(
+    model_path: str,
+    batch_size: int = 1,
+    num_cores: int = None,
+    num_sockets: int = None,
+    scheduler: Scheduler = None,
```
bnellnm, line 72: Shouldn't this default to the multi-stream scheduler?

bfineran: I believe this should be handled by the `compile_model` function.
Was going off of the defaults [here](https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/engine.py#L548).

bfineran: are you saying that for a client/server scenario, we want to instead default to multi-stream?

bfineran: that makes sense to me, so going with that.

bnellnm: Yeah, that was the general idea for multi-stream.  Unless @mgoin or anyone else has objections.  It should be fine either way for an example.

bfineran: makes sense. thanks for the catch @bnellnm

markurtz (approved):

mgoin (approved):

PR #122, opened 2021-06-23
Update README.md

jeanniefinks:
clarification on ONNX and opset versions as well as IR.


bfineran (commented):
I'd like to specify that these are "ONNX" opset and IR versions, unless that's already clear

mgoin (approved):

markurtz (approved):

PR #125, opened 2021-07-07
transformer model pipelines - base and QA

bfineran:
base framework for huggingface-like `pipeline` objects for transformer model inference. pipelines will be able to pull down a default model from sparsezoo for the task based on the available CPU. Adapted from `huggingface/transformers` and `patil-suraj/onnx_transformers`. Currently supports DeepSparse and onnxruntime, planned support for pytorch models

Example usage:
```python
>>> from deepsparse.transformers import pipeline
>>> pipe = pipeline("question-answering")
>>> pipe(question="what is my name?", context="my name is bfineran")
{'score': 0.501025378704071, 'start': 11, 'end': 18, 'answer': 'bfineran'} 
```

this PR will be converted from draft after style and documentation are updated

### DIFF
--- a/src/deepsparse/transformers/pipelines.py
+++ b/src/deepsparse/transformers/pipelines.py
@@ -58,7 +58,6 @@ except Exception as transformers_import_err:
 
 __all__ = ["ArgumentHandler", "Pipeline", "QuestionAnsweringPipeline", "pipeline"]
 
-MAX_LENGTH = 128
 
 logger = logging.get_logger(__name__) if logging else None
 
@@ -161,14 +160,14 @@ class Pipeline(_ScikitCompat):
 
     Pipeline supports running with the DeepSparse engine or onnxruntime.
 
-    Some pipeline, output large tensor object as nested-lists.
-
     :param model: loaded inference engine to run the model with, can be a
         deepsparse Engine or onnxruntime InferenceSession
     :param tokenizer: tokenizer to be used for preprocessing
     :param config: transformers model config for this model
-    :param engine: name of inference engine that is used. Options are
+    :param engine_type: name of inference engine that is used. Options are
         deepsparse and onnxruntime
+    :param max_length: maximum sequence length to set for model inputs by default.
+        default value is 128
     :param input_names: list of input names to the neural network
     :param args_parser: Reference to the object in charge of parsing supplied
         pipeline parameters. A default is provided if None
@@ -183,8 +182,8 @@ class Pipeline(_ScikitCompat):
         model: Union[Engine, InferenceSession],
         tokenizer: PreTrainedTokenizer,
         config: PretrainedConfig,
-        engine: str,
-        task: str = "",
+        engine_type: str,
+        max_length: int = 128,
         input_names: Optional[List[str]] = None,
         args_parser: ArgumentHandler = None,
         binary_output: bool = False,
@@ -195,12 +194,13 @@ class Pipeline(_ScikitCompat):
         self.model = model
         self.tokenizer = tokenizer
         self.config = config
-        self.engine = engine
+        self.engine_type = engine_type
+        self.max_length = max_length
         self.input_names = input_names
         self.binary_output = binary_output
         self._args_parser = args_parser or DefaultArgumentHandler()
         self._framework = (
-            "np" if self.engine in [DEEPSPARSE_ENGINE, ORT_ENGINE] else "pt"
+            "np" if self.engine_type in [DEEPSPARSE_ENGINE, ORT_ENGINE] else "pt"
         )
 
     def transform(self, X):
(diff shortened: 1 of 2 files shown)


src/deepsparse/transformers/pipelines.py
```diff
@@ -0,0 +1,798 @@
...
+            return self.model.run(list(inputs.values()))
+        # TODO: torch
+        # with self.device_placement():
+        #         with torch.no_grad():
+        #             inputs = self.ensure_tensor_on_device(**inputs)
+        #             predictions = self.model(**inputs)[0].cpu()
+        # if return_tensors:
+        #     return predictions
+        # else:
+        #     return predictions.numpy()
```
mgoin, line 255: is this a todo for this diff?

src/deepsparse/transformers/pipelines.py
```diff
@@ -0,0 +1,798 @@
...
+        sess_options = SessionOptions()
+        if num_cores is not None:
+            sess_options.intra_op_num_threads = num_cores
+        sess_options.log_severity_level = 3
+        sess_options.graph_optimization_level = GraphOptimizationLevel.ORT_ENABLE_ALL
+
+        onnx_model = onnx.load(onnx_path)
+        model = InferenceSession(
+            onnx_model.SerializeToString(), sess_options=sess_options
+        )
```
mgoin, line 791: is there a reason why we aren't passing in the `onnx_path` directly?

mgoin (commented):
looks great, mostly minor nits here. are we going to add a requirements.txt here or add a new "target" to the setup.py?

src/deepsparse/transformers/pipelines.py
```diff
@@ -0,0 +1,798 @@
...
+
+    Base class implementing pipelined operations.
+    Pipeline workflow is defined as a sequence of the following operations:
+
+        Input -> Tokenization -> Model Inference ->
+        Post-Processing (task dependent) -> Output
+
+    Pipeline supports running with the DeepSparse engine or onnxruntime.
+
+    Some pipeline, output large tensor object as nested-lists.
```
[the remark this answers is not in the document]
bfineran: removing, leftover from HF.

src/deepsparse/transformers/pipelines.py
bfineran: leaving for future work if we add torch support

src/deepsparse/transformers/pipelines.py
```diff
@@ -0,0 +1,798 @@
...
+        model = InferenceSession(
+            onnx_model.SerializeToString(), sess_options=sess_options
+        )
+
+    return model, input_names
+
+
+def _validate_transformers_import():
+    if transformers_import_error is not None:
+        raise transformers_import_error
```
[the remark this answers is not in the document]
bfineran: good idea. https://github.com/neuralmagic/deepsparse/pull/127

markurtz (dismissed):
LGTM

src/deepsparse/transformers/pipelines.py
```diff
@@ -0,0 +1,804 @@
...
+    engine_type: str,
+    num_cores: Optional[int],
+    num_sockets: Optional[int],
+) -> Tuple[Union[Engine, InferenceSession], List[str]]:
+    onnx_path, input_names, _ = _overwrite_model_inputs(model_path)
+
+    if engine_type == DEEPSPARSE_ENGINE:
+        model = compile_model(
+            onnx_path, batch_size=1, num_cores=num_cores, num_sockets=num_sockets
+        )
```
mgoin, line 785: max_length isn't piped into _overwrite_model_inputs yet

bfineran: good catch. fixed and tested

src/deepsparse/transformers/pipelines.py
```diff
@@ -0,0 +1,836 @@
...
+
+    def predict(self, X):
+        """
+        Scikit / Keras interface to transformers' pipelines.
+        This method will forward to __call__().
+        """
+        return self(X=X)
+
+    def _parse_and_tokenize(
+        self, *args, padding=True, add_special_tokens=True, **kwargs
```
spacemanidol, line 227: Is there an engine reason for setting padding to true? IIRC if its set to false the batches will be dynamically padded to max size for that batch.

natuan (dismissed):
Left two small change requests.

On the compatibility with sklearn pipeline, I could see another way to refactor the QuestionAnsweringPipeline (specifically its current __call__) into (a) a transformer (in the data transformation sense) that converts input examples into features and (b) a predictor that produces the answers. The former would implements `fit_transform` and ends around the call to `squad_convert_examples_to_features`; and the latter implements `predict` and does the rest of the function. This way we could chain the same transformer to two different predictors (one for DeepSparse and the other for ORT)---right now there seems not much benefit of following sklearn pipeline standard.

src/deepsparse/transformers/pipelines.py
bfineran: this comes from huggingface's source code for the base pipeline. the QA pipeline doesn't follow this path, and goes through `squad_convert_examples_to_features` instead

markurtz (approved):

mgoin (approved):

bfineran:
@natuan I agree with your point, we could definitely add that in if needed. the current structure is there to match the huggingface API behavior

spacemanidol (commented):
looks good

PR #126, opened 2021-07-08
transformers benchmarking and deployment examples

rahul-tuli:
BERT benchmarking and client/server examples. Benchmarking supports deepsparse and onnxruntime benchmarking across different batch sizes, sequence lengths, and cpu core counts.  Deployment example shows usage of `deepsparse.transformers.pipelines` to create a simple E2E flask-based BERT QA server and provides a pipelines-like client to use it.

A full readme for transformers examples will be provided in a follow up PR


examples/huggingfacebert/server.py
```diff
@@ -0,0 +1,179 @@
...
+        help="The full filepath of the ONNX model file or SparseZoo stub to the model",
+    )
+
+    parser.add_argument(
+        "-b",
+        "--batch-size",
+        type=int,
+        default=1,
+        help="The batch size to run the analysis for",
+    )
```
bfineran, line 128: delete since `batch_size` will always be 1 currently

examples/huggingfacebert/server.py
```diff
@@ -0,0 +1,179 @@
...
+    @app.route("/predict", methods=["POST"])
+    def predict():
+        """
+        Expects data in json format with question and context
+
+        '{"question":"what's my name?","context":"my name is snorlax"}'
+
+        :returns: A json with results from predictor and the inputs
+        """
+        data = json.loads(flask.request.get_data())
```
bfineran, line 77: let's add quick input validation that `"question"` and `"context"` are present and both are either a string or list of strings

examples/huggingfacebert/server.py
```diff
@@ -0,0 +1,179 @@
...
+
+Host a BERT ONNX model as a server, using the DeepSparse Engine and Flask
+
+positional arguments:
+  onnx_filepath         The full filepath of the ONNX model file or SparseZoo
+                        stub to the model
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -b BATCH_SIZE, --batch-size BATCH_SIZE
```
bfineran, line 33: update help string once batch size is removed

examples/huggingfacebert/server.py
```diff
@@ -0,0 +1,179 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: nit: let's change this directory name to `examples/huggingface-transformers`

bfineran (dismissed):
great work @rahul-tuli, left a few comments and we should also add an example client for this server

bfineran (commented):
@rahul-tuli looks like "benchmark" may have been replaced by "benchmark_bert" in more places than intended

README.md
```diff
@@ -177,7 +177,7 @@ engine_base = compile_model(
     model="zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/base-none", 
     batch_size=batch_size,
 )
-benchmarks_base = engine_base.benchmark(sample_inputs)
+benchmarks_base = engine_base.benchmark_bert(sample_inputs)
```
natuan, line 180: This is the subsection for ResNet50. Did you mean to have a different one for BERT?

bfineran: will revert these

rahul-tuli (commented):
Naice!

examples/huggingface-transformers/client.py
```diff
@@ -0,0 +1,47 @@
...
+
+class PipelineClient:
+    """
+    Client object for making requests to the example DeepSparse BERT inference server
+
+    :param address: IP address of the server, default is 0.0.0.0
+    :param port: Port the server is hosted on, default is 5543
+    """
+
+    def __init__(self, address: str = "0.0.0.0", port: str = 5543):
```
markurtz, line 35: port is marked as str but stored here as an int

examples/huggingface-transformers/server.py
```diff
@@ -0,0 +1,189 @@
...
+from flask_cors import CORS
+
+from deepsparse.transformers import pipeline
+
+
+app = Flask(__name__)
+CORS(app)
+
+
+def is_str_list_or_str(val):
```
markurtz, line 59: need a doc string for public methods or mark as private

examples/huggingface-transformers/server.py
```diff
@@ -0,0 +1,189 @@
...
+            }
+        )
+
+    @app.route("/<page_name>")
+    def unimplemented(page_name: str):
+        """
+        Route for unimplemented pages
+        """
+        return make_response(
+            f"The page named {page_name} has not been implemented.",
```
markurtz, line 119: NIT, would be nice if we're overriding this to provide more info than what would normally be available such as to list out the routes that are callable

examples/huggingface-transformers/benchmark.py
```diff
@@ -0,0 +1,394 @@
...
+            if array.shape[0] < args.max_sequence_length:
+                processed_array[: array.shape[0], ...] = array
+            else:
+                processed_array[:, ...] = array[: args.max_sequence_length, ...]
+            sample[idx] = processed_array
+        processed_samples.append(sample)
+    return processed_samples
+
+
+def _iter_batches(
```
markurtz, line 344: For the future would be good to see how we might merge this logic into the sparsezoo dataset class so we can reuse it and make the integration with the sparsezoo more seemless

bfineran: definitely, looks like adding an `iter_batches` function to `Dataset` should work out fine. will add to the backlog

markurtz (approved):

mgoin (approved):

examples/huggingface-transformers/benchmark.py
```diff
@@ -0,0 +1,394 @@
...
+        "--num-iterations",
+        help="The number of iterations the benchmark will be run for",
+        type=int,
+        default=80,
+    )
+    parser.add_argument(
+        "-w",
+        "--num-warmup-iterations",
+        help=(
+            "The number of warmup iterations that will be executed before the actual"
```
spacemanidol, line 186: rookie here. Why does warm-up matter?

examples/huggingface-transformers/benchmark.py
spacemanidol: agreed. It makes sense that there should be some common processing method for datasets. Otherwise any gains may be attributed to model and methods vs differences in testing

examples/huggingface-transformers/server.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        Expects data in json format with question and context
+
+        '{"question":"what's my name?","context":"my name is snorlax"}'
+
+        :returns: A json with results from predictor and the inputs
+        """
+        data = json.loads(flask.request.get_data())
+
+        assert "question" in data and "context" in data
```
spacemanidol, line 120: we probably also want to run an assert that no values are < -1 or greater than vocab size

spacemanidol: scratch that. Tokenization is not there. given we ware tying to have a fair benchmark it probably makes sense to have a fully processed and tokenized benchmark happening before we do any type of inference. Logically tokenization can happen on client because that is non GPU bound and only focusing on inference on input tokens will give a clearer representation of gains. I don't have data top back this up but by pushing the tokenization to the dataset this should make the gains found by deepsparse clearer.

spacemanidol (commented):
Looks good overall. In the future we probably want to move tokenization to the client or to before this pipeline to ensure that the % improvement/loss we see is reflected NN compute as much as possible.

spacemanidol (commented):
otherwise looks good

PR #127, opened 2021-07-09
option to pip install transformers with deepsparse

bfineran:
support for `pip install deepsparse[transformers]`


mgoin (approved):

rahul-tuli (approved):
Naice!

PR #144, opened 2021-07-20
Jfinks link fixes

jeanniefinks:
Markdown for URL links ending with parentheses doesn't convert to html; the recommendation is to enclose URL  with "<" and ">"

Removing unnecessary image link that wasn't rendering downstream


rahul-tuli (approved):
Looks good to me!

bfineran (approved):

PR #145, opened 2021-07-20
conditional import for onnxruntime in transformers example

bfineran:


rahul-tuli (dismissed):

examples/huggingface-transformers/pipelines.py
```diff
@@ -970,21 +978,32 @@ def _create_model(
...
 
-        model = InferenceSession(onnx_path, sess_options=sess_options)
+        model = onnxruntime.InferenceSession(onnx_path, sess_options=sess_options)
 
     return model, input_names
 
 
+def _validate_ort_import():
+    if ort_import_error is not None:
+        raise ImportError(
```
markurtz, line 997: Can we printout the error that occurred here in the exception as well for better debugging

kevinaer (dismissed):

examples/huggingface-transformers/pipelines.py
```diff
@@ -970,21 +978,32 @@ def _create_model(
...
+            "An exception occurred when importing onxxruntime. Please verify that "
+            "onnxruntime is installed in order to use the onnxruntime inference engine"
+        )
+
+
 def _validate_transformers_import():
     if transformers_import_error is not None:
         raise ImportError(
-            "An exception occured when importing from the transformers library. "
+            "An exception occurred when importing from the transformers library. "
```
markurtz, line 1006: Same thing, can we print out the error for better debugging along with the message

examples/huggingface-transformers/requirements.txt
```diff
@@ -1,4 +1,5 @@
 deepsparse
+onnxruntime>=1.4.0,<=1.6.0
```
markurtz, line 2: What's our current restriction on greater than 1.6?

markurtz (changes requested):

examples/huggingface-transformers/requirements.txt
bfineran: this is to match our `setup.py` this won't affect the example though so I'm okay taking it out

markurtz (approved):

rahul-tuli (approved):

PR #150, opened 2021-07-26
[YOLO examples] detect postprocessing in ONNX graph

bfineran:
the DeepSparse engine has planned support for YOLO post-processing. This change adds a simple rule for deciding if postprocessing is included in the YOLO ONNX graph and skips additional postprocessing if it is.

This soft rule is that the first output of the model will have fewer dimensions than the other (original) outputs since the grid based dimensions will be flattened in the post-processed outputs. (e.g. the base yolov3 and yolov5 models have 5 dims in their original outputs and 3 in the postprocessed output)

### DIFF
--- a/examples/ultralytics-yolo/README.md
+++ b/examples/ultralytics-yolo/README.md
@@ -54,7 +54,7 @@ run the following examples.
 to run inferences on images, videos, or webcam streams. For a full list of options
 `python annotate.py -h`.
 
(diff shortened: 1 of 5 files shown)


mgoin (dismissed):

markurtz (dismissed):

markurtz (approved):

rahul-tuli (approved):
Looks good to me. A small nit: maybe needs_postprocessing is a better name than has_postprocessing.

Will leave that upto you. Approving the changes! Good Work.

PR #152, opened 2021-07-28
BugFix:Download model if stub

rahul-tuli:
This PR will download a Yolo model from if a SparseZoo stub is given as model path

### DIFF
--- a/examples/ultralytics-yolo/deepsparse_utils.py
+++ b/examples/ultralytics-yolo/deepsparse_utils.py
@@ -514,6 +514,15 @@ def yolo_onnx_has_postprocessing(model_path: str) -> bool:
     return all(num_dims > outputs_num_dims[0] for num_dims in outputs_num_dims[1:])
 
 
+def _download_model_if_stub(path: str) -> str:
+    if path.startswith("zoo"):
+        model = Zoo.load_model_from_stub(path)
+        downloded_path = model.onnx_file.downloaded_path()
+        print(f"model with stub {path} downloaded to {downloded_path}")
+        return downloded_path
+    return path
+
+
 _YOLO_CLASSES = [
     "person",
     "bicycle",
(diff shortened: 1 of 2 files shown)


examples/ultralytics-yolo/deepsparse_utils.py
```diff
@@ -514,6 +514,15 @@ def yolo_onnx_has_postprocessing(model_path: str) -> bool:
     return all(num_dims > outputs_num_dims[0] for num_dims in outputs_num_dims[1:])
 
 
+def _download_model_if_stub(path: str) -> str:
```
bfineran, line 517: this function is used outside the scope of this file, so let's remove the starting `_` from the name and also add this to the `__all__` list.  Additionally, let's use this at the start of `modify_yolo_onnx_input_shape` since it's the same code

bfineran (changes requested):
good fix @rahul-tuli! small refactor request and then LGTM

examples/ultralytics-yolo/deepsparse_utils.py
```diff
@@ -514,6 +511,22 @@ def yolo_onnx_has_postprocessing(model_path: str) -> bool:
     return all(num_dims > outputs_num_dims[0] for num_dims in outputs_num_dims[1:])
 
 
+def download_model_if_stub(path: str) -> str:
+    """
+    Utility method to download model if path is a SparseZoo stub
+    :param path: file path to YOLO ONNX model or SparseZoo stub of the model
```
bfineran, line 517: nit: add blank line between summary and params

rahul-tuli: Fixed! Thanks for catching.

examples/ultralytics-yolo/deepsparse_utils.py
rahul-tuli: Resolved! ⭐

bfineran (approved):
LGTM!

markurtz (approved):

PR #162, opened 2021-07-30
[Cherry Pick] Update release 0.6 from main

markurtz:


bfineran (commented):
the first chunk of these commits were included in 0b9a6df, does this rebase revert that commit? If we're not reverting then checking out `release/0.6` and cherry picking only the new commits should fix the merge issue

PR #163, opened 2021-07-30
[BERT examples] torch benchmarking support

bfineran:
tested against:
  * CPU, GPU, FP32, FP16
  * Zoo and local models
  * base and layer-dropped models


rahul-tuli (approved):
Looks good @bfineran

markurtz (approved):

mgoin:
Definitely dead for now - that examples folder has passed away 😢

PR #165, opened 2021-08-03
Add VNNI support warning to resnet50_benchmark

mgoin:


examples/benchmark/resnet50_benchmark.py
```diff
@@ -146,6 +146,12 @@ def main():
     )
     print(f"ResNet-50 v1 Pruned Moderate FP32 {results}")
 
+    if not VNNI:
+        print(
+            "WARNING: VNNI instructions not detected, "
+            "quantization speedup not well supported"
```
bfineran, line 152:
```suggestion
            "quantization (INT8) speedup not well supported"
```

bfineran (dismissed):
added a suggestion to specify INT8 quantization since the script logs specify INT8 ResNet-50 instead of "quantized". am fine landing as-is though - up to you

bfineran (approved):

rahul-tuli (approved):
LGTM!

markurtz:
Is there anyway we can make this detection more generic and move it inside of DeepSparse by detecting quantized portions of the graph? People are going to run into this in pretty much every example we have along with their own models

mgoin:
> Is there anyway we can make this detection more generic and move it inside of DeepSparse by detecting quantized portions of the graph? People are going to run into this in pretty much every example we have along with their own models

yes we'll work on getting this into the engine as a one-time warning, like the splash message

PR #168, opened 2021-08-04
[YOLO examples] torch quant CPU benchmark

bfineran:
supports loading a recipe and model from sparsezoo, applying that recipe to the model, and then possibly converting it to a quantized torch model to run on CPU.

`torch.quantization.convert` has no affect on normal FP32 models

### DIFF
--- a/examples/ultralytics-yolo/benchmark.py
+++ b/examples/ultralytics-yolo/benchmark.py
@@ -24,7 +24,7 @@ usage: benchmark.py [-h] [-e {deepsparse,onnxruntime,torch}]
                     [-b BATCH_SIZE] [-c NUM_CORES] [-s NUM_SOCKETS]
                     [-i NUM_ITERATIONS] [-w NUM_WARMUP_ITERATIONS] [-q]
                     [--fp16] [--device DEVICE]
-                    [--model-config MODEL_CONFIG]
+                    [--model-config MODEL_CONFIG] [--recipe RECIPE]
                     model_filepath
 
 Benchmark sparsified YOLO models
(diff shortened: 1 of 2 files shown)


examples/ultralytics-yolo/benchmark.py
```diff
@@ -385,18 +399,38 @@ def _load_model(args) -> (Any, bool):
...
         else:
-            print("Using full precision")
             model.float()
+
+        if args.device == "cpu":
+            # convert any QAT models to fully quantized for CPU execution
+            try:
+                model = torch.quantization.convert(model)
+            except Exception:
+                pass
```
rahul-tuli, line 430: Thoughts on using the following idiom?

```suggestion
            with suppress(Exception):
                model = torch.quantization.convert(model)
```

`suppress` decorator can be imported from contextlib
`from contextlib import suppress`
Available from python3.4 onwards

rahul-tuli (changes requested):
Looks good overall, left one small suggestion!

rahul-tuli (approved):
Good Work Ben! This PR has my

examples/ultralytics-yolo/benchmark.py
```diff
@@ -271,6 +277,15 @@ def parse_args(arguments=None):
             "post-processing. Defaults to use standard YOLOv3/YOLOv5 anchors"
         ),
     )
+    parser.add_argument(
```
markurtz, line 280: the recipe, if any was used, is already present in the state dict exported from ultralytics. Check the loading logic that we have in the models/export.py class for yolov5. We shouldn't need to have the user supply a recipe here

markurtz (changes requested):

mgoin:
This also died 😢 a valiant death though

PR #199, opened 2021-11-16
revert import * for version.py in base init file

bfineran:


Willtor (changes requested):
On line 39, I think it wants to be `package_version=__version__`.

Willtor (approved):
Lgtm!

bfineran:
good catch!

markurtz (approved):

dhuangnm:
It looks we need this fix across all the sparse* repos?

kazukageorge:
> It looks we need this fix across all the sparse* repos?

`__version__` gives 0.9.0 and `version` yields 0.9.0.2021XXXX. Both works fine in the api

PR #203, opened 2021-11-18
removed is_release flag from init

horheynm:


src/deepsparse/__init__.py
```diff
@@ -20,6 +20,8 @@
 # flake8: noqa
 # isort: skip_file
 
+from sparsezoo.package import check_package_version as _check_package_version
```
dhuangnm, line 23: Do we need it here? It looks already in the try loop.

markurtz (approved):

PR #214, opened 2021-12-01
text and token classification pipeline for transformer examples

bfineran:
* added pipelines to support text classification/sentiment analysis and token classification/NER (gives support for MNLI, QQP, CONNL2003, and SST examples)
* added functionality to read in config from default zoo model since config needs to be task specific for classification classes
* TODO (in future): update transformer example docs, add in sparsezoo stubs to `SUPPORTED_TASKS`, until we get configs pushed to sparsezoo, classes will have default names and only support two classes

example:
```python
from pipelines import pipeline

p = pipeline("sentiment-analysis", "models/bert-sst-test.onnx")
print(p("I am happy"))

...

[{'label': 'LABEL_1', 'score': 0.9998352527618408}]


markurtz (approved):
Looks good, putting in two notes to make sure we track:

1. need to get this merged into the deepsparse source code for next release
2. need to go through benchmarks and see how much time is spent in these pipeline sections and if there are any performance improvements we need/can do

PR #222, opened 2021-12-10
Create `deepsparse.benchmark` CLI script

mgoin:
`deepsparse.benchmark` is meant to simplify benchmarking and be the central codepath for benchmarking the engine on ONNX or SparseZoo models, without worrying about end-to-end flows. This is to aid competitive benchmarking against other runtimes.

Example usage:
```
> deepsparse.benchmark zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/pruned-moderate                                                      
[INFO benchmark_model.py:134 ] Thread pinning to cores enabled
DeepSparse Engine, Copyright 2021-present / Neuralmagic, Inc. version: 0.10.0 (4af1c5e3) (optimized) (system=avx512, binary=avx512)
[INFO benchmark_model.py:177 ] deepsparse.engine.Engine:
        onnx_file_path: /home/mgoin/.cache/sparsezoo/c0673523-c5fe-40e4-bfdc-f999876371ea/model.onnx
        batch_size: 1
        num_cores: 18
        num_sockets: 0
        scheduler: Scheduler.multi_stream
        cpu_avx_type: avx512
        cpu_vnni: False
[INFO            onnx.py:176 ] Generating input 'input', type = float32, shape = [1, 3, 224, 224]
[INFO benchmark_model.py:192 ] num_streams default value chosen of 9. This requires tuning and may be sub-optimal
[INFO benchmark_model.py:202 ] Starting 'async' performance measurements for 10 seconds
| Scenario      | Throughput    | Mean (ms)     | Median (ms)   | Std (ms)      | Iterations    
| multistream   | 1992.5518     | 4.4633        | 4.4053        | 0.4183        | 19929.0000 
```
```
> deepsparse.benchmark zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/pruned_3layers-aggressive_86 -s sync -shapes "[1,128],[1,128],[1,128]" -b 64
[INFO benchmark_model.py:134 ] Thread pinning to cores enabled
DeepSparse Engine, Copyright 2021-present / Neuralmagic, Inc. version: 0.10.0 (4af1c5e3) (optimized) (system=avx512, binary=avx512)
[INFO benchmark_model.py:177 ] deepsparse.engine.Engine:
        onnx_file_path: /home/mgoin/.cache/sparsezoo/124df9a3-eda9-4b62-bba9-d98c0f1dcb57/model.onnx
        batch_size: 64
        num_cores: 18
        num_sockets: 0
        scheduler: Scheduler.single_stream
        cpu_avx_type: avx512
        cpu_vnni: False
[INFO            onnx.py:176 ] Generating input 'input_ids', type = int64, shape = [64, 128]
[INFO            onnx.py:176 ] Generating input 'attention_mask', type = int64, shape = [64, 128]
[INFO            onnx.py:176 ] Generating input 'token_type_ids', type = int64, shape = [64, 128]
[INFO benchmark_model.py:202 ] Starting 'sync' performance measurements for 10 seconds
| Scenario      | Throughput    | Mean (ms)     | Median (ms)   | Std (ms)      | Iterations    
| singlestream  | 759.8180      | 84.2119       | 76.8535       | 15.9610       | 119.0000 
 ```

### DIFF
--- a/src/deepsparse/benchmark_model/benchmark.py
+++ b/src/deepsparse/benchmark_model/benchmark.py
@@ -1,0 +1,172 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import json
+import queue
+import threading
+import time
+from typing import Dict, List, Optional
+
+import numpy as np
+
+from deepsparse import Engine, Scheduler, compile_model
+from deepsparse.utils import generate_random_inputs
+
+
+__all__ = ["model_stream_benchmark"]
+
+
+def verify_output(outputs: List[np.ndarray], correct_outputs: List[np.ndarray]):
+    assert len(correct_outputs) == len(outputs)
+    for i in range(len(correct_outputs)):
+        gt_output = correct_outputs[i]
+        nm_output = outputs[i]
+
+        assert gt_output.shape == nm_output.shape
+        assert type(gt_output) == type(nm_output)
+
+        max_diff = np.max(np.abs(nm_output - gt_output))
+        if not np.allclose(nm_output, gt_output, 0.0, 8.0e-4):
+            raise Exception(
+                "ERROR: output {}: {} {} MAX DIFF: {}".format(
+                    i, gt_output.shape, nm_output.shape, max_diff
+                )
+            )
+
+
+def iteration(model: Engine, input: List[np.ndarray]):
+    start = time.time()
+    output = model(input)
+    end = time.time()
+    return output, start, end
+
+
+def singlestream_benchmark(
+    model: Engine,
+    input_list: List[np.ndarray],
+    output_list: Optional[List[np.ndarray]],
+    seconds_to_run: float,
+) -> List[float]:
+    batch_times = []
+
+    stream_end_time = time.time() + seconds_to_run
+    while time.time() < stream_end_time:
+        output, start, end = iteration(model, input_list)
+        batch_times.append([start, end])
+        if output_list:
+            verify_output(output, output_list)
+
+    return batch_times
+
+
+class EngineExecutorThread(threading.Thread):
+    def __init__(
+        self,
+        model: Engine,
+        input_list: List[np.ndarray],
+        output_list: Optional[List[np.ndarray]],
+        time_queue: queue.Queue,
+        max_time: float,
+    ):
+        super(EngineExecutorThread, self).__init__()
+        self._model = model
+        self._input_list = input_list
+        self._output_list = output_list
+        self._time_queue = time_queue
+        self._max_time = max_time
+
+    def run(self):
+        while time.time() < self._max_time:
+            output, start, end = iteration(self._model, self._input_list)
+            self._time_queue.put([start, end])
+            if self._output_list:
+                verify_output(output, self._output_list)
+
+
+def multistream_benchmark(
+    model: Engine,
+    input_list: List[np.ndarray],
+    output_list: Optional[List[np.ndarray]],
+    seconds_to_run: float,
+    num_streams: int,
+) -> List[float]:
+    time_queue = queue.Queue()
+    max_time = time.time() + seconds_to_run
+    threads = []
+
+    for thread in range(num_streams):
+        threads.append(
+            EngineExecutorThread(model, input_list, output_list, time_queue, max_time)
+        )
+
+    for thread in threads:
+        thread.start()
+
+    for thread in threads:
+        thread.join()
+
+    return list(time_queue.queue)
+
+
+def model_stream_benchmark(
+    model: Engine,
+    input_list: List[np.ndarray],
+    output_list: Optional[List[np.ndarray]],
+    scenario: str,
+    seconds_to_run: float,
+    num_streams: int,
+) -> Dict:
+    batch_times = []
+
+    # Run the benchmark scenario and collect batch times
+    if scenario == "singlestream":
+        batch_times = singlestream_benchmark(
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/deepsparse/benchmark_model/benchmark.py [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+
+    # Convert times to milliseconds
+    batch_times_ms = [
+        (batch_time[1] - batch_time[0]) * 1000 for batch_time in batch_times
+    ]
+
+    # Calculate statistics
+    first_start_time = min([b[0] for b in batch_times])
+    last_end_time = max([b[1] for b in batch_times])
+    total_time_executing = last_end_time - first_start_time
```
bnellnm, line 119: I don't get this bit here.  It seems like this calculation will include all the test overhead in total_time_executing?

Why not use the sum of all the (end-start) times for each batch?

src/deepsparse/benchmark_model/benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+    return output, start, end
+
+
+def singlestream_benchmark(
+    model: Engine,
+    input_list: List[np.ndarray],
+    seconds_to_run: float,
+) -> List[float]:
+    batch_times = []
+
```
bnellnm, line 41: should some warmup runs be included here?

src/deepsparse/benchmark_model/benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+def multistream_benchmark(
+    model: Engine,
+    input_list: List[np.ndarray],
+    seconds_to_run: float,
+    num_streams: int,
+) -> List[float]:
+    time_queue = queue.Queue()
+    max_time = time.time() + seconds_to_run
+    threads = []
+
```
bnellnm, line 79: ditto

src/deepsparse/benchmark_model/benchmark_model.py
```diff
@@ -0,0 +1,255 @@
...
+            "async scenario. Default is automatically determined for given hardware "
+            "and may be sub-optimal."
+        ),
+    )
+    parser.add_argument(
+        "-pin",
+        "--thread_pinning",
+        type=str,
+        default="true",
+        choices=["true", "false", "numa"],
```
bnellnm, line 105: would "none", "cores", "numa" be more descriptive names?

src/deepsparse/benchmark_model/benchmark_model.py
```diff
@@ -0,0 +1,255 @@
...
+
+    if pinning_mode == "true":
+        os.environ["NM_BIND_THREADS_TO_CORES"] = "1"
+        log.info("Thread pinning to cores enabled")
+    elif pinning_mode == "false":
+        os.environ["NM_BIND_THREADS_TO_CORES"] = "0"
+        log.info("Thread pinning to cores disabled, performance may be sub-optimal")
+    elif pinning_mode == "numa":
+        os.environ["NM_BIND_THREADS_TO_CORES"] = "1"
+        os.environ["NM_BIND_THREADS_TO_SOCKETS"] = "1"
```
bnellnm, line 140: I don't think you want both of these at the same time.  THREADS_TO_CORES defeats the purpose of THREADS_TO_SOCKETS

src/deepsparse/benchmark_model/benchmark_model.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # If num_streams isn't defined, find a default
+    if not args.num_streams and scenario not in "singlestream":
+        args.num_streams = int(model.num_cores / 2)
+        log.info(
+            "num_streams default value chosen of {}. "
+            "This requires tuning and may be sub-optimal".format(args.num_streams)
+        )
+
+    # Run the model once to warm up
```
bnellnm, line 195: i missed this above.  nevermind the previous comments.  is running once enough tho?

src/deepsparse/benchmark_model/benchmark.py
mgoin: We want to know all of the executions that could be performed within a given amount of wallclock time. For multistream if we summed all the times for each batch this would be incorrect. 
This is meant to be a realistic but best case benchmark so I tried to remove overhead from the actual runs, those sections are definitely up to be audited.

src/deepsparse/benchmark_model/benchmark_model.py
mgoin: i like that, will change

src/deepsparse/benchmark_model/benchmark_model.py
mgoin: Maybe not, I will add a warmup running for 1 second

src/deepsparse/benchmark_model/benchmark.py
bnellnm: Ok I guess that's reasonable.  A comment would be good tho.  If this only applicable for multi-stream then maybe we should not penalize ourselves in the singlestream case.  It also would probably be simpler to just record the start and end times around multistream_benchmark/singlestream_benchmark instead?

src/deepsparse/benchmark_model/benchmark_model.py
mgoin: okay

src/deepsparse/benchmark_model/benchmark_model.py
```diff
@@ -0,0 +1,251 @@
...
+    pinning_mode = pinning_mode.lower()
+
+    if pinning_mode in "core":
+        os.environ["NM_BIND_THREADS_TO_CORES"] = "1"
+        log.info("Thread pinning to cores enabled")
+    elif pinning_mode in "none":
+        os.environ["NM_BIND_THREADS_TO_CORES"] = "0"
+        log.info("Thread pinning to cores disabled, performance may be sub-optimal")
+    elif pinning_mode in "numa":
+        os.environ["NM_BIND_THREADS_TO_SOCKETS"] = "1"
```
bnellnm, line 139: maybe `os.environ["NM_BIND_THREADS_TO_CORES"] = "0"` just to be extra paranoid here?

bnellnm (dismissed):
LGTM.  I didn't take a super close look at the shape override stuff though.

src/deepsparse/benchmark_model/benchmark.py
mgoin: This gives less overhead compared to timing the whole benchmark function call so I'll keep it as is. I left a hefty comment and think we should do measurements of overhead later

bnellnm (approved):

rahul-tuli (approved):

PR #223, opened 2021-12-13
Add support for the elastic scheduler.

Willtor:
The elastic scheduler distributes requests across NUMA nodes rather than opportunistically multi-streaming them within a single node. This eliminates cache contention between requests which is beneficial for many workloads.


bfineran (approved):
LGTM
just want to note that the `from_str` method functionality mirrors just doing `Scheduler("elastic")`, not sure if this method was added for a particular reason

Willtor:
> let's wait until the engine can make use of it avocado

:avocado:

PR #239, opened 2022-01-12
Add support for the elastic scheduler.

Willtor:
The elastic scheduler distributes requests across NUMA nodes rather than opportunistically multi-streaming them within a single node. This eliminates cache contention between requests which is beneficial for many workloads.


mgoin (approved):
looks good to me! might not be the final scheduler though ;)

Willtor:
> looks good to me! might not be the final scheduler though ;)

Lol! Fair enough. "Finally" for now. ;)

markurtz (approved):

PR #254, opened 2022-01-26
Allow input shape override to broadcast across all inputs

mgoin:
For example, many BERT models have 3 inputs with exactly the same shape.
So instead of writing 
`deepsparse.benchmark bert.onnx --input_shapes [1,128],[1,128],[1,128]`
you can now write
`deepsparse.benchmark bert.onnx --input_shapes [1,128]`
and it will try to apply the shape to all inputs.

### DIFF
--- a/src/deepsparse/utils/onnx.py
+++ b/src/deepsparse/utils/onnx.py
@@ -232,7 +232,8 @@ def override_onnx_input_shapes(
     # Input shapes should be a list of lists, even if there is only one input
     assert all(isinstance(inp, list) for inp in input_shapes)
 
-    # If there is a single input shape given and multiple inputs, duplicate for all inputs
+    # If there is a single input shape given and multiple inputs,
+    # duplicate for all inputs to apply the same shape
     if len(input_shapes) == 1 and len(external_inputs) > 1:
         for i in range(len(external_inputs) - 1):
             input_shapes.append(input_shapes[0])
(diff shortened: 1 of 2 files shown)


src/deepsparse/utils/onnx.py
```diff
@@ -229,14 +229,31 @@ def override_onnx_input_shapes(
...
-    assert len(external_inputs) == len(input_shapes)
-
     # Input shapes should be a list of lists, even if there is only one input
     assert all(isinstance(inp, list) for inp in input_shapes)
 
-    # Overwrite the shapes
+    # If there is a single input shape given and multiple inputs,
+    # duplicate for all inputs to apply the same shape
+    if len(input_shapes) == 1 and len(external_inputs) > 1:
+        for i in range(len(external_inputs) - 1):
```
rahul-tuli, line 238: Why do we leave out the last external input here?

rahul-tuli, line 238: nit
```suggestion
        for _ in range(len(external_inputs) - 1):
```

bfineran (dismissed):
nice

rahul-tuli (dismissed):

src/deepsparse/utils/onnx.py
mgoin: This is trying to duplicate input_shapes to the same size as external_inputs, so i had it do this external_inputs-1 times. I'm not a fan of this loop so cleaner suggestions would be nice

ElizaWszola: Maybe something like
`input_shapes.extend([input_shapes[0] for i in range(1, len(external_inputs))])` ?

ElizaWszola:
Looks good

ElizaWszola (dismissed):
Looks good

src/deepsparse/utils/onnx.py
mgoin: Cool, that works and I like it better

andy-neuma (approved):
cool

bfineran (approved):

PR #258, opened 2022-02-01
Add model config and tokenizer specification for eval_downstream

mgoin:


src/deepsparse/transformers/eval_downstream.py
```diff
@@ -264,6 +268,13 @@ def parse_args():
         type=int,
         default=None,
     )
+    parser.add_argument(
+        "--model-dir",
+        help="The name of the model config to use, or path to directory with "
+        "`config.json` and `tokenizer.json` files. Default is None",
```
rahul-tuli, line 274: Could we add this arg description to the top level docstring?

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -116,6 +117,7 @@ def mnli_eval(args):
     # load pipeline
     text_classify = pipeline(
         task="text-classification",
+        model_name=args.model_dir,
```
rahul-tuli, line 120: This does not seem correct, `model_name` should be canonical name of the huggingface model, if model is to be initialized using a directory, I think the correct usage is to pass that directory to `model_path`, which should contain `config,json`, `tokenizer.json`, and `model.onnx`

Like the following:
```
my_model_dir
├── config.json
├── model.onnx
└── tokenizer.json
```

mgoin: Yup, I was using this wrong thanks Rahul!

PR #261, opened 2022-02-07
Restrict mistune version to fix docs build

mgoin:
After starting from a fresh virtualenv, dev dependencies installation resolves to mistune version is 2.0.2. It seems that the newer version doesn't work with our docs build. We can probably use something newer than 0.8.4 but that works for now


dhuangnm (approved):
Thanks Michael.

bfineran (approved):

PR #263, opened 2022-02-09
Accept configurable pytest args from env var

dbarbuzzi:
Add an optional variable to the Makefile to allow passing custom flags to the pytest commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ DOCDIR := docs
 MDCHECKGLOBS := 'docs/**/*.md' 'docs/**/*.rst' 'examples/**/*.md' 'scripts/**/*.md'
 MDCHECKFILES := CODE_OF_CONDUCT.md CONTRIBUTING.md DEVELOPING.md README.md
 SPARSEZOO_TEST_MODE := "true"
+PYTEST_ARGS ?= ""
 
 PYTHON := python3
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ DOCDIR := docs
 MDCHECKGLOBS := 'docs/**/*.md' 'docs/**/*.rst' 'examples/**/*.md' 'scripts/**/*.md'
 MDCHECKFILES := CODE_OF_CONDUCT.md CONTRIBUTING.md DEVELOPING.md README.md
 SPARSEZOO_TEST_MODE := "true"
+PYTEST_ARGS ?= ""
```
dhuangnm, line 11: maybe we should name it JUNIT_ARGS instead since it's to determine the report format?

dbarbuzzi: I could change the name that way, but then I’d also want to change the way it’s used in the command invocation. Currently, it is acting as an actual generic placeholder to allow any additional pytest args. If it were to be specifically for the purpose of the JUnit report file, I’d probably give it a name like JUNIT_FILE and change the invocation to pass that value to pytest’s `--junitxml` flag — which do you think is the better approach?

dbarbuzzi: Certainly a possibility, but to keep it as trivial as your example, that would also mean hardcoding the filepath to the XML report file as the `--junitxml` flag requires a filename/filepath to the report file. With the flexibility of the current approach, you would be able to define the path as the consumer requires rather than having it hardcoding (e.g. `make test PYTEST_ARGS='--junitxml path/to/report.xml'`).

Let me know if the flexibility is not desired in which case I will make the suggested changes.

dhuangnm: I think the solution looks fine to me. @neuralmagic/machine-learning can you please review and provide comments?

bfineran (approved):
LGTM

PR #272, opened 2022-02-24
Refactor of server to generalize it, correct performance issues, and cleanup implementation

markurtz:


src/deepsparse/server/main.py
```diff
@@ -0,0 +1,212 @@
...
+    "--workers",
+    type=int,
+    default=1,
+    help="Use multiple worker processes. Defaults to 1.",
+)
+@click.option(
+    "--log_level",
+    type=str,
+    default="info",
+    help="Bind to a socket with this port. Defaults to 5543.",
```
mgoin, line 152: help comment is incorrect and this should be a choice arg

src/deepsparse/server/main.py
```diff
@@ -0,0 +1,212 @@
...
+    from starlette.responses import RedirectResponse
+except Exception as err:
+    raise ImportError(
+        "Exception while importing install dependencies, "
+        "run `pip install deepsparse[server]` to install the dependencies. "
+        f"Recorded exception: {err}"
+    )
+
+
+ENV_DEEPSPARSE_SERVER_CONFIG = "DEEPSPARSE_SERVER_CONFIG"
```
bfineran, line 103: should be imported from config.py

src/deepsparse/server/main.py
```diff
@@ -0,0 +1,212 @@
...
+            response_model=pipeline_def.response_model,
+            tags=["prediction"],
+        )(_predict_func)
+        defined_tasks.add(pipeline_def.config.task)
+    else:
+        raise ValueError(
+            f"Multiple tasks defined for {pipeline_def.config.task} and no alias "
+            f"given for {pipeline_def.config}. "
+            "Either define an alias or supply a single model for the task"
+        )
```
bfineran, line 157: since we are now only adding one route per `pipeline_def` with the only difference between them being the route name, it would probably be cleaner to choose the route first (raise error if applicable), then define+decorate the predict fn

src/deepsparse/server/main.py
```diff
@@ -0,0 +1,283 @@
...
+    task: str,
+    model_path: str,
+    batch_size: int,
+):
+    """
+    Start a DeepSparse inference server for serving the models and pipelines given
+    within the config_file or a single model defined by task, model_path, and batch_size
+
+    Example config.yaml for serving: \n
+        models: \n
```
bfineran, line 258: why are these `\n`s necessary here?

src/deepsparse/server/main.py
```diff
@@ -0,0 +1,283 @@
...
+    "--workers",
+    type=int,
+    default=1,
+    help="Use multiple worker processes. Defaults to 1.",
+)
+@click.option(
+    "--log_level",
+    type=str,
+    default="info",
+    help="Bind to a socket with this port. Defaults to info.",
```
bfineran, line 212: do we want to add choices here, or just let it fail?

markurtz:
screencapture-192-168-201-236-5543-docs-2022-02-24-17_19_13

bfineran (changes requested):
LGTM overall, would be good in a follow up to add some performance info/timings and report back in the health check

src/deepsparse/server/main.py
markurtz: yep, good catch

src/deepsparse/server/main.py
markurtz: click was automatically folding all of the text into single lines and ignoring the new lines. Found a work around in their docs now that uses \b before a paragraph to prevent this

src/deepsparse/server/main.py
markurtz: yep, added

bfineran (approved):

KSGulin (approved):

src/deepsparse/server/config.py
```diff
@@ -0,0 +1,206 @@
...
+    Put a server configuration in an environment variable given by env_key.
+    If config_file is given, ignores task, model_path, and batch_size.
+    Otherwise, creates a configuration file from task, model_path, and batch_size
+    for serving a single model.
+
+    :param config_file: the path to the config file to store in the environment
+    :param task: the task the model_path is serving such as question_answering.
+        If config_file is supplied, this is ignored.
+    :param model_path: the path to a model.onnx file, a model folder containing
+        the model.onnx and supporting files, or a SparseZoo model stub.
```
mgoin, line 185: have you tested the model folder method? this isn't supported in the default deepsparse flow i.e. compile_model

src/deepsparse/server/main.py
```diff
@@ -0,0 +1,280 @@
...
+    model_path: str,
+    batch_size: int,
+):
+    """
+    Start a DeepSparse inference server for serving the models and pipelines given
+    within the config_file or a single model defined by task, model_path, and batch_size
+
+    Example config.yaml for serving:
+
+    \b
```
mgoin, line 254: \b?

mgoin (approved):

src/deepsparse/server/config.py
markurtz: yes, this is tested for transformer pipeline flows

src/deepsparse/server/main.py
markurtz: this is necessary for click to print out the paragraph in --help without folding the white space

PR #277, opened 2022-03-02
Remove server import for transformers pipelines to not require server dependencies

markurtz:


src/deepsparse/transformers/server.py
```diff
@@ -20,9 +20,8 @@
 
 from pydantic import BaseModel, Field
```
mgoin, line 21: can be obscure pydantic or does this need to be part of the autoinstall?

mgoin (commented):
thanks, one comment about pydantic still

src/deepsparse/transformers/server.py
markurtz: oh yep, good catch. That makes this a bit harder, but we can get something optional working here

mgoin (approved):

spacemanidol (approved):

PR #287, opened 2022-03-16
Update README.md

jeanniefinks:
Updating the tone of instruction set support.


bfineran (commented):
would suggest breaking up the first sentence or rephrasing the "with support" portion of the sentence as it is unclear what with is modifying

NMsasa (approved):

bfineran (approved):

PR #307, opened 2022-04-05
Allow input overrides in check_correctness.py

tlrmchlsmth:
Hack up check_correctness.py so that we can override input shapes

### DIFF
--- a/examples/benchmark/check_correctness.py
+++ b/examples/benchmark/check_correctness.py
@@ -51,7 +51,10 @@ from deepsparse.utils import (
     generate_random_inputs,
     get_input_names,
     get_output_names,
+    model_to_path,
     override_onnx_batch_size,
+    override_onnx_input_shapes,
+    parse_input_shapes,
     verify_outputs,
 )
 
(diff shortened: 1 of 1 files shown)


examples/benchmark/check_correctness.py
```diff
@@ -51,7 +51,10 @@
     generate_random_inputs,
     get_input_names,
     get_output_names,
+    model_to_path,
     override_onnx_batch_size,
+    override_onnx_input_shapes,
+    parse_input_shapes,
     verify_outputs,
 )
```
mgoin, line 60:
```suggestion
from deepsparse.benchmark_model.ort_engine import ORTEngine
```

examples/benchmark/check_correctness.py
```diff
@@ -81,30 +84,51 @@ def parse_args():
...
-
-        ort_outputs = ort_network.run(output_names, inputs_dict)
+    if input_shapes:
+        with override_onnx_input_shapes(onnx_filepath, input_shapes) as override_onnx_filepath:
+            ort_network = onnxruntime.InferenceSession(override_onnx_filepath, sess_options)
+            ort_outputs = ort_network.run(output_names, inputs_dict)
+    else:
+        with override_onnx_batch_size(onnx_filepath, batch_size) as override_onnx_filepath:
+            ort_network = onnxruntime.InferenceSession(override_onnx_filepath, sess_options)
+            ort_outputs = ort_network.run(output_names, inputs_dict)
```
mgoin, line 127: you could delete all the names/dict variables too
```
        ort_network = ORTEngine(
            model=onnx_filepath,
            batch_size=batch_size,
            input_shapes=input_shapes,
        )
        ort_outputs = ort_network.run(inputs)
```

mgoin (commented):
we should probably move `ORTEngine` out of benchmark_model but using it here would simplify a bit

bnellnm (approved):

PR #315, opened 2022-04-12
[feature/Pipeline] Pipeline base class implementation

bfineran:
(main feature: #317)

Pipeline bass class to abstract the general flow of wrapping `Engine` objects with pre/post processing and standardizing inputs/outputs with Pydantic models.  Includes structure for a registry for easy loading by task name

full documentation on intended usage provided in `Pipeline` docstring.

General plan is for the pipeline to be easily dropped into generic scripts for server, batch inference, benchmarking, etc so these scripts may be shared across domain

### DIFF
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -50,9 +50,9 @@ class Pipeline(ABC):
     def __init__(
         self,
         model_path: str,
-        engine_type: str,
-        batch_size: int,
-        num_cores: int,
+        engine_type: str = DEEPSPARSE_ENGINE,
+        batch_size: int = 1,
+        num_cores: int = None,
         scheduler: Scheduler = None,
         input_shapes: List[List[int]] = None,
     ):
@@ -81,9 +81,9 @@ class Pipeline(ABC):
     def create(
         task: str,
         model_path: str,
-        engine_type: str,
-        batch_size: int,
-        num_cores: int,
+        engine_type: str = DEEPSPARSE_ENGINE,
+        batch_size: int = 1,
+        num_cores: int = None,
         scheduler: Scheduler = None,
         input_shapes: List[List[int]] = None,
         **kwargs,
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -102,6 +102,8 @@ class Pipeline(ABC):
         Pass None for the default
     :param input_shapes: list of shapes to set ONNX the inputs to. Pass None
         to use model as-is. Default is None
+    :param alias: optional name to give this pipeline instance, useful when
+        inferencing with multiple models. Default is None
     """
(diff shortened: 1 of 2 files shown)


src/deepsparse/pipeline.py
```diff
@@ -0,0 +1,375 @@
...
+                pipeline_class is not _REGISTERED_PIPELINES[task_name]
+            ):
+                raise RuntimeError(
+                    f"task {task_name} already registered by Pipeline.register. "
+                    f"attempting to register pipeline: {pipeline_class}, but"
+                    f"pipeline: {_REGISTERED_PIPELINES[task_name]}, already registered"
+                )
+            _REGISTERED_PIPELINES[task_name] = pipeline_class
+
+        def decorator(pipeline_class: Pipeline):
```
rahul-tuli, line 250: Could we wrap this in `functools.wraps(cls, updated=())`? to ensure proper propagation of dunder attributes

rahul-tuli, line 250: Could we also change the name `decorator` to something more specific?

src/deepsparse/pipeline.py
```diff
@@ -0,0 +1,375 @@
...
+        """
+        :return: optional name to give this pipeline instance, useful when
+            inferencing with multiple models
+        """
+        return self._alias
+
+    @property
+    def model_path_orig(self) -> str:
+        """
+        :return: value originally passed to the model_path argument to initialize this
```
rahul-tuli, line 323: nit:
```suggestion
        :return: value originally passed to the `model_path` argument to initialize this
```

src/deepsparse/pipeline.py
```diff
@@ -0,0 +1,375 @@
...
+            # set task and task_aliases as class level property
+            pipeline_class.task = task
+            pipeline_class.task_aliases = task_aliases
+
+        return decorator
+
+    @abstractmethod
+    def setup_onnx_file_path(self) -> str:
+        """
+        Performs any setup to unwrap and process the given model_path and other
```
rahul-tuli, line 268: nit:
```suggestion
        Performs any setup to unwrap and process the given `model_path` and other
```

rahul-tuli (changes requested):

src/deepsparse/pipeline.py
bfineran: `cls` would refer to `Pipeline` here, not the implementing `pipeline_class`,  I think as long as I fix the decorator to `return pipeline_class` the wrapping should go through fine

rahul-tuli: Changes look good, resolving this thread!

rahul-tuli (approved):

KSGulin (approved):
Great code, intuitive documentation, LGTM

PR #317, opened 2022-04-13
deepsparse.Pipeline - generic pipeline, deepsparse.server support, NLP,IC,OD pipelines

bfineran:
co-authored by @rahul-tuli

The goal of this feature is to create a generic Pipeline class to abstract the general flow of wrapping `Engine` objects with pre/post processing and standardizing inputs/outputs with Pydantic models. These Pipelines should be easily dropped into deepsparse scripts for serving, inference, benchmarking, etc

instantiate a Pipeline using `Pipeline.create`
```python
from deepsparse import Pipeline

qa_pipeline = Pipeline.create("qa")
qa_pipeline(question="where do i live", context="i am an engineer and i live in america")
```

TODO:
- [x] Pipeline base class (#315)
- [x] PIpelineConfig (#318)
- [x] deepsparse.server refactor for Pipeline compatibility (#319)
- [x] deepsparse.transformers refactor to use Pipeline base class (#320)
- [x] deepsparse.transformers pipeline cleanup (#335)
- [x] support for sparsezoo model defaults included with `Pipeline.register` (#339)
- [x] rename _model to _schema (#340)
- [x] Image Classification Pipeline (#322)
- [x] Image Classification Pipeline scripts (#328)
- [x] YOLO Pipeline (#327)
- [x] YOLO annotation script refactor (#332)

PR #320, opened 2022-04-14
[feature/Pipeline] deepsparse.Pipeline implementations for transformers

bfineran:
(main feature: #317)

This PR refactors `src/deepsparse/transformers/{pipelines.py,server.py}` to use `deepsparse.Pipeline` as the main pipeline framework. Goal is to keep as much of the existing flexible functionality while sacrificing some for consistency provided by the generic pipeline class.

TODO:
- [x] add parsing layer to `deepsparse.Pipeline` to allow flexible inputs similar to existing transformers pipelines
- [x] `deepsparse.transformers.TransformersPipeline` base class
- [x] TextClassificationPipeline
- [x] TokenClassificationPipeline
- [x] QuestionAnsweringPipeline


markurtz (approved):
Need to fix conflict, but looks good.

I'm not a fan of some of the code for the implemented pipelines, but understand why we're pulling it in that way for future updates.

Additionally, we should get the license included in these files for the code that influenced

KSGulin (approved):
Great stuff

PR #323, opened 2022-04-18
Update label mapping for deepsparse.transformers.eval_downstream

mgoin:
Fixes QQP, SST2, and MNLI dataset pipelines

### DIFF
--- a/src/deepsparse/transformers/eval_downstream.py
+++ b/src/deepsparse/transformers/eval_downstream.py
@@ -161,11 +161,13 @@ def qqp_eval(args):
     )
     print(f"Engine info: {text_classify.model}")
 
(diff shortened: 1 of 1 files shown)


markurtz (approved):

bfineran (dismissed):
lgtm, might want to consider a fallback if the mappings from the data config aren't there

mgoin:
> lgtm, might want to consider a fallback if the mappings from the data config aren't there

i wish there was an easy way to check this. i assume the best is we see if the output string starts with "LABEL"?

bfineran (approved):

markurtz (approved):

andy-neuma (approved):
cool

PR #325, opened 2022-04-19
DeepSparse Dockerfile and build processes

dbogunowicz:
Related to Asana task:
https://app.asana.com/0/1201735099598270/1201964439431277/f
See related PR in SparseML:
https://github.com/neuralmagic/sparseml/pull/720


rahul-tuli:
Would it make sense to maybe start from a python3.8 image instead of ubuntu?

dbogunowicz:
> Would it make sense to maybe start from a python3.8 image instead of ubuntu?

No particular reason. But you are right, bare python image may be more lightweight, let me test it.

dbogunowicz:
@rahul-tuli good point. The original `deepsparse_docker` was 1.11GB. Using `bullseye` allowed me to forgo installing dependencies, but bare bullseye is pretty big, 1.28GB. Swapping for `bullseye-slim` gives me 490mb and the benefits of bullseye (less clutter in the Dockerfile).

And it works all right. Tested manually and successfully ran:
```
from deepsparse.transformers import pipeline

# SparseZoo model stub or path to ONNX file
model_path = "zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/12layer_pruned80_quant-none-vnni"

qa_pipeline = pipeline(
    task="question-answering",
    model_path=model_path,
)

my_name = qa_pipeline(question="What's my name?", context="My name is Snorlax")
```

docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Setup the base image
+FROM python:3.8-slim-bullseye
+
+# Activate venv
+RUN python3.8 -m venv /venv
+ENV PATH="venv/bin:$PATH"
+
+# Setup DeepSparse
+RUN pip3 install deepsparse[server]
+
```
KSGulin, line 10: We should make bash the default entrypoint. Partially to encourage CLI commands as the first point of use and partially because when python is the entrypoint, exiting to bash is non-intuitive, if possible.

```suggestion

# Bash as default entrypoint
CMD ["bin/bash"]

```

KSGulin (commented):
Looks great @dbogunowicz! Approved pending the entrypoint update

dbogunowicz:
@KSGulin Thank you. Included your suggestion into the PR.

docker/README.md
```diff
@@ -0,0 +1,6 @@
+## DeepSparse docker image
+This directory contains the Dockerfile to create minimal DeepSparse docker image.
+This image is based off the latest official Ubuntu image.
+
+In order to build and launch this image, run from the root directory:
+`docker build -t deepsparse_docker . && docker run -it deepsparse_docker`
```
bfineran, line 6: let's also give an example of invoking a script directly
`docker run deepsparse-image deepsparse.server ...`

docker/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+# Bash as default entrypoint
+CMD ["bin/bash"]
```
bfineran, line 12: should likely switch `CMD` to `ENTRYPOINT` to allow for override

bfineran (changes requested):

docker/Dockerfile
dbogunowicz: this is not necessary since the base image by default has python set as an entry point.

bfineran (dismissed):

rahul-tuli (approved):

KSGulin (approved):

PR #326, opened 2022-04-20
added dockerised example for sparseserver ui

Chibukach:
Added dockerised example for sparseserverui

- added docker-compose file for single host setup
- added Dockerfiles for both client and server
- split requirements.txt for client and server
- updated settings.py with new variant due to docker address accessed internally
- updated README.MD


mgoin:
Hi @Chibukach , thanks for the docker example. As an initial cleanup do you think the client and server prepended files into the client and server folders?

examples/sparseserver-ui/README.md
```diff
@@ -86,3 +86,19 @@ Visit `http://localhost:8501` in your browser to view the demo.
 - Ubuntu 20.04.4 LTS
 - Python 3.8.10
 </samp>
```
InquestGeronimo, line 88: would add </samp> at the end of the file to maintain style

examples/sparseserver-ui/clientrequirements.txt
```diff
@@ -0,0 +1 @@
+streamlit==1.8.1
```
mgoin, line 1: this could be put in the client dockerfile

examples/sparseserver-ui/serverrequirements.txt
```diff
@@ -0,0 +1 @@
+deepsparse[server]>=0.11
```
mgoin, line 1: ditto for server

examples/sparseserver-ui/client/settings.py
```diff
@@ -29,17 +31,29 @@ class FeatureHandler:
...
+            model="question_answering/12l_pruned80_quant", address = "deepsparseserver"
         ),
         "6-Layer BERT, Quantized, 96% of Base Accuracy": MultiPipelineClient(
-            model="question_answering/quant6lagg96"
+            model="question_answering/quant6lagg96", address = "deepsparseserver"
         ),
         "3-Layer BERT, Quantized, 89% of Base Accuracy": MultiPipelineClient(
-            model="question_answering/quant3lagg89"
-        ),
+            model="question_answering/quant3lagg89", address = "deepsparseserver"
```
mgoin, line 43: `address = "deepsparseserver"` doesn't seem to be needed based on the README update referring to localhost. The address could be settable as a command-line argument for the app if you want to keep that feature. Seems not extensible by hardcoding it to each model

examples/sparseserver-ui/clientrequirements.txt
Chibukach: I normally use requirements.txt incase we need more than one python library in the future, but i'll remove it and add the  installation directly to the Dockerfile

examples/sparseserver-ui/README.md
Chibukach: Will do, cheers

examples/sparseserver-ui/client/settings.py
Chibukach: Thanks for the review. The deepsparseserver was used due to the internal connectivity via the docker bridge network. 
I've changed the docker network to host in the next PR, so there's no need to keep the feature and it can be referenced directly via localhost.

Chibukach (commented):
Updated PR in response to PR #326
- context added to dockercompose to cleanup Dockerfiles for client and server
- removed requirements.txt files for client and server. Moved to pip install within the Docker files
- updated README.md for consistence
- updated docker network from bridge to host

examples/sparseserver-ui/README.md
```diff
@@ -85,4 +85,27 @@ Visit `http://localhost:8501` in your browser to view the demo.
...
 - Python 3.8.10
-</samp>
+
+
+## <div>`DOCKER COMPOSE`</div>
+
+### Pre-requisite
+
+- Docker
+- Docker Compose
```
mgoin, line 95: would be nice to have links out how to install these

mgoin (approved):
thanks for responding to comments and lgtm!

KSGulin (approved):
LGTM!

PR #331, opened 2022-04-26
server integration check bug fix

bfineran:
need to verify integration is set before calling `integration.lower()`


mgoin:
could we make integration a "choice" argument between default and sagemaker? then it could always be a string

rahul-tuli:
Could we update the docstring with the `--integration` arg,
Noting I also think this option should be a choice b/w "default" and "sagemaker", it will allow an early exit.

bfineran:
thanks guys, updated

horheynm (approved):

rahul-tuli (approved):

mgoin (approved):

PR #334, opened 2022-04-27
Topology file for HB120rs_v3

govindr-nm:
Specifies core-per-CCX grouping for HB120rs_v3 VM's, used by multi-process script.


tlrmchlsmth:
Could you add a blurb to the README.md telling users that they can use this .json file for this AWS instance type?

govindr-nm:
> Could you add a blurb to the README.md telling users that they can use this .json file for this AWS instance type?

Done

PR #354, opened 2022-05-06
update twitter-nlp example for compatibility with new and old pipelines

bfineran:


mgoin (approved):
nice, good to have this example thanks

rahul-tuli (approved):

PR #355, opened 2022-05-09
Added the engine context class to engine.py

SageMoore:
Currently, this class can be used to use the same scheduler for multiple Engines.


src/deepsparse/engine.py
```diff
@@ -131,6 +132,21 @@ def _validate_scheduler(scheduler: Union[None, str, Scheduler]) -> Scheduler:
...
 
+class Context(object):
+    def __init__(
+        self,
+        num_cores: int,
+        num_streams: int = 1,
+        scheduler: Scheduler = None,
+    ):
+        self._deepsparse_context = LIB.deepsparse_context(
+            num_cores, num_streams, _validate_scheduler(scheduler).value
```
mgoin, line 143:
```suggestion
            _validate_num_cores(num_cores), num_streams, _validate_scheduler(scheduler).value
```

mgoin (commented):
lgtm when you're ready

src/deepsparse/engine.py
```diff
@@ -131,6 +132,21 @@ def _validate_scheduler(scheduler: Union[None, str, Scheduler]) -> Scheduler:
     return scheduler
 
 
+class Context(object):
```
mgoin, line 135: can you put a block comment describing the function of Context? i'm thinking similar to the Scheduler

SageMoore:
I made some non-trivial changes so this probably deserves another look

src/deepsparse/engine.py
```diff
@@ -170,22 +200,39 @@ def __init__(
...
+                "The Context was initialized with {} cores and the Engine was initialized"
+                " with {}. These two values should be equal.".format(
+                    str(context.num_cores()), str(num_cores)
+                )
+            )
+
+        model_path = self._model_path
         if self._input_shapes:
-            with override_onnx_input_shapes(
+            model_path = override_onnx_input_shapes(
```
mgoin, line 217: we need to keep the `with` statement here so the temporary file generated by `override_onnx_input_shapes` will be cleaned up after the engine is created

src/deepsparse/engine.py
```diff
@@ -170,22 +200,39 @@ def __init__(
         self._cpu_avx_type = AVX_TYPE
         self._cpu_vnni = VNNI
 
+        if context is not None and (scheduler is not None and scheduler != "elastic"):
```
mgoin, line 203:
```suggestion
        if context and scheduler != Scheduler.elastic:
```

src/deepsparse/engine.py
```diff
@@ -170,22 +200,39 @@ def __init__(
...
+            raise ValueError(
+                "Contexts are currently only supported when running the elastic scheduler"
+            )
+        if context is not None and context.num_cores() != num_cores:
```
mgoin, line 207:
```suggestion
        if context and context.num_cores() != num_cores:
```

src/deepsparse/engine.py
```diff
@@ -521,6 +525,92 @@ def _properties_dict(self) -> Dict:
         }
 
 
+class Context(object):
+    """
+    Contexts can be used to run multiple instances of the MultiModelEngine with the same
+    scheduler. This is useful for running multiple models concurrently in a server-like
+    environment.
```
tlrmchlsmth, line 532: Can you elaborate here on why it's useful? e.g. "This allows one scheduler to manage the resources of the system effectively, keeping engines that are running different models from fighting over system resources."

tlrmchlsmth (commented):
looks good, just one nit

tlrmchlsmth (approved):

bfineran:
excited for this addition @SageMoore.  What was the motivation for going for `MultiModelEngine` over having `context` as an `Engine` parameter?

SageMoore:
@bfineran There was a lot of interface pollution because we had to support both passing in the number of cores and the scheduler AND the Context object, meaning we needed to do some not-nice consistency checks. In a different language, I may advocate for overloading the constructor of the Engine vs creating the MultiModelEngine, but we can't really do that here.

For miscellaneous reasons that I'm happy to get into behind the firewall, we only support running the elastic scheduler with the Context object. This along with a desire to preserve the existing Engine interface means that we can't just use the context for everything. It is specifically designed for multi model support.

Honestly, I think separating out these two use cases now is a good way forward. We are essentially changing who owns the scheduler and all data associated with it. IMO, that's a fundamental enough change to justify a subclass.

bfineran:
thanks @SageMoore. In terms of spinning up pipelines/servers is there any reason to prefer `Engine` over `MultiModelEngine` even if running only one model?

SageMoore:
There's no reason to favor it if you don't need to run concurrent requests with different models on the same system. Performance should be equivalent.

src/deepsparse/engine.py
```diff
@@ -521,6 +525,93 @@ def _properties_dict(self) -> Dict:
...
+        if self._input_shapes:
+            with override_onnx_input_shapes(
+                self._model_path, self._input_shapes
+            ) as model_path:
+                self._eng_net = LIB.deepsparse_engine(
+                    model_path,
+                    self._batch_size,
+                    self._num_cores,
+                    self._scheduler.value,
+                    context.value(),
```
bnellnm, line 603: is the reason these are all passed in together because there's no overload of LIB.deepsparse_engine?  i.e. one that just takes the context and one that takes num_cores/scheduler?  it would be nice to hide the num_cores/scheduler on Context but nbd.

bnellnm (approved):

src/deepsparse/engine.py
SageMoore: Yeah, this is the reason.

PR #357, opened 2022-05-09
Yolo pipeline propagate args

rahul-tuli:
This PR allows propagation of postprocesing kwargs in the YOLO pipeline like `iou_thres` and `conf_thres`

Before this PR the callee had no control over iou/conf thresholds for `non_maximal_suppression` when using pipelines,

This PR enables passing in such kwargs during call:
```python
object_detection_pipeline = Pipeline.create(...)
object_detection_pipeline(images=[inputs], iou_thres=0.5, conf_thres=0.001)
```

### DIFF
--- a/.github/workflows/local-install-check.yaml
+++ b/.github/workflows/local-install-check.yaml
@@ -1,0 +1,31 @@
+name: Local Install Checks
+on: 
+  push:
+    branches:
+      - main
+      - 'release/*'
+  pull_request:
+    branches:
+      - main
+      - 'release/*'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  local-install-test:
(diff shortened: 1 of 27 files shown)


dbogunowicz (dismissed):

bfineran (dismissed):
tentatively approving, but I think it would be better to eventually move this into the input schema OR the model constructor so it will be compatible with `deepsparse.server`

KSGulin (dismissed):

KSGulin (approved):

bfineran (approved):

bfineran:
Please see comment from before:
```
tentatively approving, but I think it would be better to eventually move this into the input schema OR the model constructor so it will be compatible with deepsparse.server
```

dbogunowicz (approved):

PR #358, opened 2022-05-09
Add support for a num_streams parameter to the engine that allows use…

Willtor:
…rs to specify the max number of concurrent requests.


mgoin (commented):
it seems like #355 is enabling the same feature through the Context class so i would like to wait for that, is that right?

Willtor:
Looks like it's going to go with Sage's changes entirely. If I need to make any adjustments (don't think I will) I'll create another diff for them. I'm closing this pull request.

PR #359, opened 2022-05-09
Update DeepSparse Benchmark link on README

mgoin:


rahul-tuli (approved):
Note: Should we change these links to relative links?

bnellnm (approved):

mgoin:
> Note: Should we change these links to relative links?

this would be nice but that should change everything so a larger discussion

PR #362, opened 2022-05-13
Pass num_streams as a parameter to the engine.

Willtor:
This gives the engine the information it needs to make intelligent decisions about work division across threads.


Willtor:
Want to get D4226 approved for landing before this goes in, since our nightlies automatically incorporate changes from this repo. This will break without that diff because it's an API change.

src/deepsparse/engine.py
```diff
@@ -117,6 +118,18 @@ def _validate_num_cores(num_cores: Union[None, int]) -> int:
...
 
+def _validate_num_streams(num_streams: Union[None, int], num_cores: int) -> int:
+    if not num_streams:
+        num_streams = NUM_STREAMS
+
+    if num_streams < 0:
+        raise ValueError("num_streams must be greater than 0")
+
+    if num_streams > num_cores:
+        raise ValueError("num_cores cannot exceed num_cores.")
```
mgoin, line 129:
```suggestion
        raise ValueError("num_streams cannot exceed num_cores.")
```

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -298,12 +298,23 @@ def main():
     orig_model_path = args.model_path
     args.model_path = model_to_path(args.model_path)
 
+    if args.num_streams:
+        _LOGGER.info("num_streams set to {}".format(args.num_streams))
+    elif not args.num_streams and scenario not in "singlestream":
+        # If num_streams isn't defined, find a default
+        args.num_streams = max(1, int(args.num_cores / 2))
```
bnellnm, line 305: is the /2 because of hyper threads?

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -298,12 +298,23 @@ def main():
...
+        _LOGGER.info(
+            "num_streams default value chosen of {}. "
+            "This requires tuning and may be sub-optimal".format(args.num_streams)
+        )
```
tlrmchlsmth, line 309: This looks good to land as-is but for the elastic scheduler, when num_streams isn't defined, I think it makes sense to pass in None, and then query the engine for its preferred num streams parameter and use that in the python code for the number of requests to put in for benchmarking. We'll do this in a follow-up.

tlrmchlsmth (commented):
lgtm once Michael's fixup is incorporated

src/deepsparse/engine.py
Willtor: Derp. I also reformatted. Trying to remember how to push after rebasing...

src/deepsparse/benchmark/benchmark_model.py
Willtor: That's the original code. I just moved it thinking I was going to minimize the change size. But that's probably not right. I think the right answer is None, here. I'm going to change it.

Willtor: It's changed. BAM!

src/deepsparse/benchmark/benchmark_model.py
Willtor: Makes sense.

Willtor:
I'm perplexed by the failing linter. When I run it, I get:
```
[2018]$ black -l 80 src/deepsparse/engine.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
```

Willtor:
Derp. I get so used to 80 character wide terminals... This should fix it. :P

src/deepsparse/benchmark/benchmark_model.py
mgoin: @bnellnm @Willtor this num_streams is used to generate the number of python threads submitting inferences to the engine, so it cannot be None it needs to be a real number

mgoin (changes requested):
Need to maintain num_streams as a positive integer value for the multistream benchmarking

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -282,6 +282,27 @@ def parse_scenario(scenario):
...
+def parse_num_streams(model, scenario):
+
+    # If model.num_streams is set, and the scenario is either "multi_stream" or
+    # "elastic", use the value of num_streams given to us by the model, otherwise
+    # use a semi-sane default value.
+    if scenario == "single_stream":
+        return 1
+    else:
+        if model.num_streams:
+            return model.num_streams
```
mgoin, line 294: okay nice!

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -315,6 +337,7 @@ def main():
             input_shapes=input_shapes,
         )
     _LOGGER.info(model)
+    num_streams = parse_num_streams(model, scenario)
```
mgoin, line 340: this will fail if you ask the ORT engine for num_streams, it doesn't have that property. either need to add it to `ort_engine.py` or do some thing else

tlrmchlsmth (approved):
lgtm!

src/deepsparse/engine.py
```diff
@@ -553,6 +586,9 @@ def value(self):
     def num_cores(self):
         return self._num_cores
 
+    def num_streams(self):
+        return self._num_streams
+
     def scheduler(self):
         return self._scheduler
```
mgoin, line 593: can you turn these member functions to properties like the Engine class?

mgoin (changes requested):

docs/source/scheduler.md
```diff
@@ -47,19 +47,19 @@ _Elastic scheduling; requests execute in parallel, but not multiplexed on indivi
 Depending on your engine execution strategy, enable one of these options by running:
 
 ```python
-engine = compile_model(model_path, batch_size, num_cores, "single_stream")
+engine = compile_model(model_path, batch_size, num_cores, 1, "single_stream")
```
mgoin, line 50:
```suggestion
engine = compile_model(model_path, scheduler="single_stream")
```

docs/source/scheduler.md
```diff
@@ -47,19 +47,19 @@ _Elastic scheduling; requests execute in parallel, but not multiplexed on indivi
...
 ```
 
 or
 
 ```python
-engine = compile_model(model_path, batch_size, num_cores, "multi_stream")
+engine = compile_model(model_path, batch_size, num_cores, num_streams, "multi_stream")
```
mgoin, line 56:
```suggestion
engine = compile_model(model_path, scheduler="multi_stream")
```

docs/source/scheduler.md
```diff
@@ -47,19 +47,19 @@ _Elastic scheduling; requests execute in parallel, but not multiplexed on indivi
...
 ```
 
 or
 
 ```python
-engine = compile_model(model_path, batch_size, num_cores, "elastic")
+engine = compile_model(model_path, batch_size, num_cores, num_streams, "elastic")
```
mgoin, line 62:
```suggestion
engine = compile_model(model_path, scheduler="elastic")
```

src/deepsparse/engine.py
```diff
@@ -535,16 +564,20 @@ class Context(object):
     :param num_cores: The number of physical cores to run the model on. If more
         cores are requested than are available on a single socket, the engine
         will try to distribute them evenly across as few sockets as possible.
+    :param num_streams: The max number of requests the model can handle
+        concurrently.
     """
 
     def __init__(
         self,
         num_cores: int,
```
mgoin, line 573:
```suggestion
        num_cores: int = None,
```

src/deepsparse/engine.py
```diff
@@ -161,12 +177,14 @@ def __init__(
         model: Union[str, Model, File],
         batch_size: int,
         num_cores: int,
```
mgoin, line 179:
```suggestion
        num_cores: int = None,
```

docs/source/scheduler.md
SageMoore: Why remove batch size and num cores?

tlrmchlsmth (approved):

PR #371, opened 2022-05-17
Fix squad pipeline

anmarques:
Enable the use of squad examples with multiple features so it matches the score computed during training.


src/deepsparse/transformers/pipelines/question_answering.py
```diff
@@ -147,6 +147,32 @@ def __init__(
 
         super().__init__(**kwargs)
 
+    def __call__(self, *args, **kwargs) -> BaseModel:
```
bfineran, line 150: does this call override add something besides unpacking `engine_inputs, postprocess_kwargs` this should already be supported by the base class right?

bfineran (changes requested):
looks like quality needs to be updated. LGTM overall, my main question here is - is this backwards compatible with the QA models trained before the latest transformers updates?

src/deepsparse/transformers/pipelines/question_answering.py
anmarques: The main difference this override adds is on line 162 where it calls the engine multiple times if you have more than one set of inputs. The unpacking is just something I simplified for this specific case.

bfineran: ah. nice, missed that. is that to support more spans? would prefer to instead add an overridable method in the base Pipeline `def engine_forward(self, inputs, **kwargs)` that just does the engine forward, and then override in the QA pipeline. that way if there's changes to `__call__` in the future they will be propagated to QA

anmarques:
I don't think it'll affect backwards compatibility. This is only about how to process the inputs/outputs.

bfineran:
ok can you test this branch against one of the older squad models in the zoo? LGTM after that, pending the suggested __call__ refactor

mgoin (approved):
LGTM, I like this solution using engine_forward

src/deepsparse/transformers/pipelines/question_answering.py
bfineran: implementation looks great!

PR #375, opened 2022-05-18
Sagemaker Feature Addition

InquestGeronimo:
Adding a class to automate the build of a SageMaker endpoint.

TODO
Need to add docstrings, and a small edit to readme to show how users can call this class if they're interested.


examples/aws-sagemaker/create_endpoint.py
```diff
@@ -0,0 +1,171 @@
...
+        self.sm_boto3.create_endpoint_config(**self.endpoint_config)
+
+    def create_endpoint(self):
+
+        self.sm_boto3.create_endpoint(
+            EndpointName=self.endpoint_name,
+            EndpointConfigName=self.endpoint_config_name,
+        )
+
+        pprint(self.sm_boto3.describe_endpoint(EndpointName=self.endpoint_name))
```
mgoin, line 136: can you share the output of running this script in the PR description? want to make sure user has the information to know what to do next

examples/aws-sagemaker/create_endpoint.py
```diff
@@ -0,0 +1,171 @@
...
+        role_arn="<PLACEHOLDER>",
+    )
+
+    Sage.create_image()
+    Sage.create_ecr_repo()
+    Sage.push_image()
+    Sage.create_model()
+    Sage.create_endpoint_config()
+    Sage.create_endpoint()
+    # Sage.nuke_endpoint()
```
mgoin, line 171: if this is being run as a script we should have shutting down the endpoint as part of it. maybe once the endpoint is setup prompt to user to press a key to shutdown endpoint and stop the process. even better if this happened on exit/deconstruction

examples/aws-sagemaker/README.md
```diff
@@ -19,13 +19,40 @@ limitations under the License.
...
+
+```python
+from qa_client import Endpoint
+
+
+qa = Endpoint("us-east-1", "question-answering-example-endpoint")
+
+qa.predict(question="who is batman?", context="Mark is batman.")
+```
+
```
mgoin, line 51: Add the destroy command after to tell them to clean up

examples/aws-sagemaker/README.md
```diff
@@ -19,13 +19,40 @@ limitations under the License.
...
+git clone https://github.com/neuralmagic/deepsparse.git
+cd deepsparse/examples/aws-sagemaker
+pip install -r requirements.txt
+```
+
+Run the following command to build your SageMaker endpoint.
+
+```bash
+python endpoint.py create
+```
```
mgoin, line 38: tell them to replace the PLACEHOLDER ARN with their's before running

mgoin (approved):
thanks for the edits!

mgoin (dismissed):

examples/aws-sagemaker/endpoint.py
```diff
@@ -0,0 +1,219 @@
...
+        dest="action",
+        type=str,
+        choices=["create", "destroy"],
+        help="Args to 'create' or 'destroy' a SageMaker endpoint.",
+    )
+
+    return parser.parse_args()
+
+
+def main():
```
markurtz, line 184: Can we remove argparse and change over to click, this will also alleviate my concerns below about the if else statement

examples/aws-sagemaker/endpoint.py
```diff
@@ -0,0 +1,219 @@
...
+
+    if args.action == "create":
+        SM.create_image()
+        SM.create_ecr_repo()
+        SM.push_image()
+        SM.create_model()
+        SM.create_endpoint_config()
+        SM.create_endpoint()
+
+    else:
```
markurtz, line 214: generally, it's not great to default to destroy without checking that that's what the user passed in. This can lead to bugs in the future if we were to add more options and a later implementor forgets to update here. It'd be better if we could migrate back to the original code that was marked as resolved here. As noted above, though, moving to click will take care of issues like this for us so we don't have to maintain the code

markurtz (changes requested):
Looks good, a few small requests

examples/aws-sagemaker/endpoint.py
mgoin: @markurtz it isn't defaulting to destroy, the argparse `action` param is required and errors if you don't provide one of the choices. you have to make a choice on the CLI and click gives the same functionality in this case. as said before, we should wait on click as a refactor to all of the examples here

examples/aws-sagemaker/qa_client.py
```diff
@@ -0,0 +1,51 @@
...
+            )
+        )
+        res = self.client.invoke_endpoint(
+            EndpointName=self.endpoint_name,
+            Body=body,
+            ContentType=self.content_type,
+            Accept=self.accept,
+        )
+
+        print(res["Body"].readlines())
```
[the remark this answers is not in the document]
mgoin: +1, good call

examples/aws-sagemaker/endpoint.py
markurtz: Correct, but if a new choice is added in the future this opens up a potential pathway that a later implementor may forget about. Generally, it's never good to default to an else statement and the code across SparseML, SparseZoo, and DeepSparse has generally adhered to this principle to make future maintenance and testing easier. Towards the inclusion of click, it's a simple change that removes this issue altogether

mgoin: if someone adds a new choice they have to add to `choices=["create", "destroy"],`

markurtz: Correct, but again the if else condition becomes a weak point especially if there are multiple of them. This is a general design pattern and code maintence we've strived to keep across the codebase -- no default else statements in the event that there are pathways that people forget to update. If the next implementor updates the choices but forgets to update one of the if else statements that is called only in certain conditions, we're now deleting in that case rather than raising an exception.

markurtz: Same for click, new pathways added into the code base were agreed upon to go through click rather than argparse because it limits the amount of code and testing we have to own.

mgoin (approved):

PR #382, opened 2022-05-23
Expose `benchmark_model` and `benchmark_sweep` functions

mgoin:
Gives a functional code interface into the benchmarks that `deepsparse.benchmark` runs

### DIFF
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -258,7 +258,7 @@ def decide_thread_pinning(pinning_mode: str) -> None:
         )
 
 
-def parse_scheduler(scenario) -> Scheduler:
+def parse_scheduler(scenario: str) -> Scheduler:
     scenario = scenario.lower()
     if scenario == "multistream":
         return Scheduler.multi_stream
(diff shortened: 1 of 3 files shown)


src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -303,94 +307,131 @@ def parse_num_streams(num_streams, num_cores, scenario):
...
+            model=model_path,
+            batch_size=batch_size,
+            num_cores=num_cores,
+            num_streams=num_streams,
             scheduler=scheduler,
             input_shapes=input_shapes,
         )
         num_streams = parse_num_streams(model.num_streams, model.num_cores, scenario)
-    elif args.engine == ORT_ENGINE:
+    elif engine == ORT_ENGINE:
```
KSGulin, line 347: Could use an else statement with a `raise ValuError` to prevent an undefined `model` variable further down.

KSGulin (commented):
Looks good! Approved pending a couple comments

src/deepsparse/benchmark/benchmark_model.py
mgoin: good call thanks

anmarques (approved):
LGTM

InquestGeronimo (approved):

PR #391, opened 2022-05-27
Update cpu.py for arm/neon

bnellnm:
This updates cpu.py to accept neon as a valid architecture and changes some of the messaging/checks for fp32/quantized performance.


andy-neuma (approved):
cool

tlrmchlsmth:
Should we go ahead update this for SVE as well? Otherwise it looks like we're only planning on supporting NEON

mgoin (approved):

tlrmchlsmth (approved):

rahul-tuli (approved):

PR #402, opened 2022-06-01
Added pipeline creation functions

kylesayrs:
This PR adds creation functions that allow easier pipeline creation. The new functions positional arguments and keyword arguments to be passed.


src/deepsparse/pipeline.py
```diff
@@ -49,6 +54,51 @@
 _REGISTERED_PIPELINES = {}
 
 
+def question_answering_pipeline(*args, **kwargs) -> "Pipeline":
+    """
+    :param args: the positional arguments to pass into Pipeline.create
+    :param kwargs: the keyword arguments to pass into Pipeline.create
+    :return: question_answering pipeline created with args and kwargs
+    """
+    return Pipeline.create(*(("question_answering",) + args), **kwargs)
```
rahul-tuli, line 63: This is so cool

rahul-tuli (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -49,6 +54,51 @@
...
+
+
+def text_classification_pipeline(*args, **kwargs) -> "Pipeline":
```
bfineran, line 66: let's make another one called `sentiment_analysis_pipeline` that does the same thing

src/deepsparse/pipeline.py
bfineran, line 63: any reason we can't do `Pipeline.create("question-answering", *args, **kwargs)` a little easier to read

bfineran (changes requested):
looks good overall @kylesayrs.

it's nice that using generic args, keyword args for these will limit the impact downstream updates causing future issues if these functions aren't updated, let's at least get the existing docstrings for each of these pipelines copied in though, so when docs are generated, each of these functions will have some info on what should be included

bfineran (approved):

rahul-tuli (approved):

PR #409, opened 2022-06-01
Propagate Context in Pipelines

rahul-tuli:
Add support for using `MultiModelEngine` with `Pipeline` when `Context` is passed

Needed by #410

### DIFF
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -108,7 +108,11 @@ class Pipeline(ABC):
         to use model as-is. Default is None
     :param alias: optional name to give this pipeline instance, useful when
         inferencing with multiple models. Default is None
-    :param context: Optional Context to use while creating the engine
+    :param context: Optional Context object to use for creating instances of
+        MultiModelEngine. The Context contains a shared scheduler along with
+        other runtime information that will be used across instances of the
+        MultiModelEngine to provide optimal performance when running multiple
+        models concurrently.
     """
 
     def __init__(
@@ -120,7 +124,7 @@ class Pipeline(ABC):
         scheduler: Scheduler = None,
         input_shapes: List[List[int]] = None,
         alias: Optional[str] = None,
-        context: Optional["Context"] = None,
+        context: Optional[Context] = None,
     ):
         self._model_path_orig = model_path
         self._model_path = model_path
@@ -138,7 +142,8 @@ class Pipeline(ABC):
         self.context = context
         if self.context and self.context.num_cores != num_cores:
             raise ValueError(
-                f"num_cores mismatch. Expected {self.context.num_cores}, but got {num_cores}."
+                f"num_cores mismatch. Expected {self.context.num_cores}, "
+                f"but got {num_cores}."
             )
         self.onnx_file_path = self.setup_onnx_file_path()
         self.engine = self._initialize_engine()
@@ -190,6 +195,7 @@ class Pipeline(ABC):
         scheduler: Scheduler = None,
         input_shapes: List[List[int]] = None,
         alias: Optional[str] = None,
+        context: Optional[Context] = None,
         **kwargs,
     ) -> "Pipeline":
         """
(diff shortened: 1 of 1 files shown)


src/deepsparse/pipeline.py
```diff
@@ -119,6 +120,7 @@ def __init__(
         scheduler: Scheduler = None,
         input_shapes: List[List[int]] = None,
         alias: Optional[str] = None,
+        context: Optional["Context"] = None,
```
mgoin, line 123: why are the quotes needed here? looks like Context is already imported

mgoin (commented):
looks good to me

src/deepsparse/pipeline.py
```diff
@@ -301,10 +304,15 @@ def _register_pipeline_tasks_decorator(pipeline_class: Pipeline):
...
-    def from_config(cls, config: Union["PipelineConfig", str, Path]) -> "Pipeline":
+    def from_config(
+        cls,
+        config: Union["PipelineConfig", str, Path],
+        context: Optional[Context] = None,
+    ) -> "Pipeline":
         """
         :param config: PipelineConfig object, filepath to a json serialized
             PipelineConfig, or raw string of a json serialized PipelineConfig
+        :param context: Optional Context object for MultiModelEngine
```
bfineran, line 315: same update here

bfineran (commented):
looks good. few quick comments

src/deepsparse/pipeline.py
rahul-tuli: Good catch, thanks!

src/deepsparse/pipeline.py
```diff
@@ -108,6 +108,11 @@ class PipelineImplementation(Pipeline):
         to use model as-is. Default is None
     :param alias: optional name to give this pipeline instance, useful when
         inferencing with multiple models. Default is None
+    :param context: Optional Context object to use for creating instances of
+        MultiModelEngine. The Context contains a shared scheduler along with
+        other runtime information that will be used across instances of the
+        MultiModelEngine to provide optimal performance when running multiple
+        models concurrently.
```
bfineran, line 115: nit: remove `.` for consistency

src/deepsparse/pipeline.py
```diff
@@ -133,6 +139,12 @@ def __init__(
         if engine_type.lower() == DEEPSPARSE_ENGINE:
             self._engine_args["scheduler"] = scheduler
 
+        self.context = context
+        if self.context and self.context.num_cores != num_cores:
```
bfineran, line 143: if num_cores is the default (aka `None`) let's not raise an error and just override num_cores to that of the context

bfineran: this does have the potential to miss unintended edge cases though where the user could intentionally pass in None for `num_cores` and something else for `Context.num_cores`.  Don't want to overengineer, can discuss solutions if you think this is an issue

src/deepsparse/pipeline.py
```diff
@@ -301,10 +320,19 @@ def _register_pipeline_tasks_decorator(pipeline_class: Pipeline):
...
+        context: Optional[Context] = None,
+    ) -> "Pipeline":
         """
         :param config: PipelineConfig object, filepath to a json serialized
             PipelineConfig, or raw string of a json serialized PipelineConfig
+        :param context: Optional Context object to use for creating instances of
+            MultiModelEngine. The Context contains a shared scheduler along with
+            other runtime information that will be used across instances of the
+            MultiModelEngine to provide optimal performance when running
+            multiple models concurrently.
```
bfineran, line 335: nit: remove . for consistency

bfineran (commented):
few quick comments, updated docstring reads great

src/deepsparse/pipeline.py
```diff
@@ -119,12 +124,22 @@ def __init__(
...
         self._model_path = model_path
         self._engine_type = engine_type
         self._alias = alias
 
+        self.context = context
+        if self.context is not None:
+            num_cores = num_cores or self.context.num_cores
+            if self.context.num_cores != num_cores:
+                raise ValueError(
+                    f"num_cores mismatch. Expected {self.context.num_cores}, "
```
bfineran, line 139: should specify mismatch is between context and Pipeline. would be hard for a user to figure out where the expected val is coming from

bfineran (approved):

dbogunowicz (approved):

PR #410, opened 2022-06-01
Scheduler Multi Model support in DeepSparse server

rahul-tuli:
Create and pass a `Context` object for all pipelines in `deepsparse.server`

TODO:
- [x] Change Branch to main before landing

Also depends on PR #409


src/deepsparse/server/main.py
```diff
@@ -213,8 +213,11 @@ def server_app_factory():
     initialize_aysnc(config.workers)
     _LOGGER.debug("loaded server config %s", config)
     _add_general_routes(app, config)
-
-    pipelines = [Pipeline.from_config(model_config) for model_config in config.models]
+    context = Context()
```
mgoin, line 216: is there any num_cores to pass in here?

mgoin (dismissed):

src/deepsparse/server/main.py
bfineran: good catch, num_cores needs to be set here, @rahul-tuli let's make sure this gets supported

bfineran (changes requested):

src/deepsparse/server/main.py
rahul-tuli: Addressed

src/deepsparse/server/config.py
```diff
@@ -119,6 +120,7 @@ def server_config_to_env(
     model_path: str,
     batch_size: int,
     integration: str,
+    num_cores: Optional[int] = None,
```
bfineran, line 123: as discussed, we don't want to add in `num_cores` at the server config level because this is also a property of the pipeline config. let's instead read in from the pipeline config where `num_cores` already exists.

bfineran (changes requested):

bfineran (dismissed):
LGTM

src/deepsparse/server/main.py
```diff
@@ -214,7 +214,15 @@ def server_app_factory():
     _LOGGER.debug("loaded server config %s", config)
     _add_general_routes(app, config)
 
-    pipelines = [Pipeline.from_config(model_config) for model_config in config.models]
+    num_cores = None
+    for model_config in config.models:
+        num_cores = num_cores or model_config.num_cores
```
mgoin, line 219: it'd be nice to expand the Context to run on the largest num_cores specified
```suggestion
    num_cores = None
    for model_config in config.models:
        num_cores = max(num_cores, model_config.num_cores)
```

mgoin (dismissed):

bfineran (dismissed):

bfineran (approved):

mgoin (approved):

PR #413, opened 2022-06-03
Quick-Fix SageMaker readme

InquestGeronimo:
Brought up the `requirements` section to the top of page so users now see reqs more clearly.


dhuangnm:
I think we also need to mention the ECR permission (maybe sagemaker too) granted to the user who does aws configure since it's also needed.

mgoin (dismissed):

mgoin (approved):

bfineran (dismissed):
LGTM, might be worth it to call out the required permissions in its own bullet / small section to reduce friction

dhuangnm (approved):

mgoin (approved):

PR #416, opened 2022-06-06
[Bugfix] add torchvision in image_classification dependencies

rahul-tuli:
This PR, adds torchvision to image_classification dependencies,

And wraps external library imports into a try catch block, showing apt install command to the user

### DIFF
--- a/src/deepsparse/image_classification/__init__.py
+++ b/src/deepsparse/image_classification/__init__.py
@@ -31,6 +31,17 @@ def _check_torchvision_install(raise_on_fail=False):
         return torchvision_import_exception
 
 
+def _check_click_install(raise_on_fail=False):
+    try:
+        import click as _click
+
+        return None
+    except Exception as click_import_exception:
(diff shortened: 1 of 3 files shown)


tdg5 (approved):

bfineran (changes requested):
(discussed offline) blocking for the time being since `torchvision` is already automatically installed in the `image_classification` directory's init file. click was a straggling dep that was causing issues, so we are looking to move that to the auto install as well

src/deepsparse/image_classification/__init__.py
```diff
@@ -64,6 +75,30 @@ def _check_install_deps():
             f"{torchvision_exception}"
         )
 
+    try:
+        _subprocess.check_call(
```
bfineran, line 79: let's extract some of the shared code to a helper

rahul-tuli:
Closing this out to rework the logic, will re-open once ready

PR #421, opened 2022-06-07
yolo pipeline image shape override support

bfineran:
adds support for adding x,y shape override for yolo pipeline inputs. both dimensions must be divisible by all model strides

example for overriding to (416,320)
```python
>>> from deepsparse import Pipeline
>>> yolo_pipeline = Pipeline.create("yolo", image_size=(416, 320))
```

### DIFF
--- a/src/deepsparse/yolo/pipelines.py
+++ b/src/deepsparse/yolo/pipelines.py
@@ -21,7 +21,14 @@ import onnx
 from deepsparse.pipeline import Pipeline
 from deepsparse.utils import model_to_path
 from deepsparse.yolo.schemas import YOLOInput, YOLOOutput
-from deepsparse.yolo.utils import COCO_CLASSES, YoloPostprocessor, postprocess_nms
+from deepsparse.yolo.utils import (
+    COCO_CLASSES,
+    YoloPostprocessor,
+    get_onnx_expected_image_shape,
+    modify_yolo_onnx_input_shape,
+    postprocess_nms,
+    yolo_onnx_has_postprocessing,
+)
 
 
 try:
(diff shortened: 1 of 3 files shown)


src/deepsparse/yolo/pipelines.py
```diff
@@ -158,7 +191,7 @@ def process_inputs(self, inputs: YOLOInput) -> List[numpy.ndarray]:
 
             if isinstance(image, str):
                 image = cv2.imread(image)
-                image = cv2.resize(image, dsize=self.input_shape)
+                image = cv2.resize(image, dsize=tuple(reversed(self.image_size)))
```
rahul-tuli, line 194: is this cause channel dim is last in cv2?

src/deepsparse/yolo/utils/utils.py
```diff
@@ -326,6 +330,100 @@ def yolo_onnx_has_postprocessing(model_path: str) -> bool:
...
+    :param dim: dimension index of the tensor to get the shape of
+    :return: shape of the tensor at the given dimension
+    """
+    return tensor.type.tensor_type.shape.dim[dim].dim_value
+
+
+def set_tensor_dim_shape(tensor: onnx.TensorProto, dim: int, value: int):
+    """
+    Sets the shape of the tensor at the given dimension to the given value
+    :param tensor: ONNX tensor to modify the shape of
```
rahul-tuli, line 420: nit: blank line b/w description, params

rahul-tuli (dismissed):

markurtz (dismissed):
Looks good, just noting for the future it would be good to start getting all of our ONNX utilities and code migrated into SparseZoo so we can share the functionality across DeepSparse and SparseML code pathways

src/deepsparse/yolo/pipelines.py
bfineran: I don't think that's it, but this was required for shapes to line up

markurtz (approved):

rahul-tuli (approved):

PR #433, opened 2022-06-09
Added exception for when given task is not in list of supported tasks

kylesayrs:
Added an exception when trying to get a task name that is not in the list of supported tasks and an accompanying helper class method to get the names of the currently supported tasks.

The hope is that this error message will be more useful than the current one.
`ValueError: Unknown Pipeline task my_fake_task. Pipeline tasks should be must be declared with the Pipeline.register decorator. Currently registered pipelines: []`

->

`ValueError: Unknown Pipeline task my_fake_task. Currently supported tasks are ['question_answering', 'qa', 'text_classification', 'glue', 'sentiment_analysis', 'token_classification', 'ner', 'image_classification', 'yolo']`


src/deepsparse/tasks.py
```diff
@@ -134,3 +142,11 @@ def is_yolo(cls, task: str) -> bool:
         :return: True if it is an segmentation task using YOLO, False otherwise
         """
         return cls.yolo.yolo.matches(task)
+
+    @classmethod
+    def task_names(cls):
+        task_names = []
+        for task_category in cls.all_task_categories:
+            for task in task_category:
+                task_names += [task._name] + task._aliases
```
rahul-tuli, line 151: can we add a check here to add aliases to task_names only if they're different from task._name?

rahul-tuli (approved):

mgoin (approved):

bfineran (approved):

PR #434, opened 2022-06-09
[BugFix] Remove paranthesis when asking for context properties

rahul-tuli:
The current Implementation of `MultiModelEngine` has a bug where it refers to context properties as functions, this PR fixes that by removing paranthesis [python calls respective property functions automatically under the hood]

TLDR; if props are referred to with paranthesis an exception of the following form is thrown: `An object of ..... type is not callable`. This PR fixes that


kylesayrs (approved):
Confirmed these members have the property decorator

mgoin (approved):

SageMoore (approved):

PR #438, opened 2022-06-10
Added unit tests for MultiModelEngine

SageMoore:
These tests take about 90 seconds to  run on my 18 core computer.

### DIFF
--- a/tests/test_multi_engine.py
+++ b/tests/test_multi_engine.py
@@ -1,0 +1,73 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+import threading
+from deepsparse import MultiModelEngine, Context
+from deepsparse.utils import verify_outputs
+from sparsezoo.models import classification
+from sparsezoo.objects import Model
+
+
+model_test_registry = {
+    "mobilenet_v1": classification.mobilenet_v1,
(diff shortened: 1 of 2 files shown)


dbarbuzzi:
Worth noting that the GitHub action base test check with all of the added tests took just over 13m to run (previously this check was probably 5-7m on average). If this is too long, some or all could be removed by removing the smoke tag from them. Partial coverage would make it a bit messy as the loops would have to be changed/unrolled to only tag certain parametrizations as smoke.

SageMoore:
> 

This is definitely too much. I'll break some of the tests out.

SageMoore:
@dbarbuzzi I made a minor fix to the model selection logic and it looks like that dropped the time to 7 minutes. Is that OK?

dbarbuzzi:
Yeah, 7 minutes should be fine; my given target when setting up the GitHub actions was less than 10 minutes.

tests/test_multi_engine.py
```diff
@@ -0,0 +1,74 @@
...
+        pred_outputs = engine(inputs)
+        verify_outputs(pred_outputs, outputs)
+
+    def test_multi_engine(
+        self,
+        num_streams: int,
+        num_requests: int,
+    ):
+        models = list(model_test_registry.values())
+        batch_size = 64
```
mgoin, line 59: do we need a large batch size here?

SageMoore: The batch size doesn't make any difference in terms of coverage. I just picked 64. 1 would also be fine. Most of the runtime of the test is dominated by compilation so I don't have a strong opinion here. On my computer b1 is about 5 seconds faster (out of 75). The runtime difference could be more significant on the github action vm so we can use 1 instead and see if it helps.

SageMoore:
Ok the first time I used b1 the total runtime of the base-tests job was 10 minutes. The second time it was 6. I assume this kind of variance is somewhat expected? Both times I ran the b64 job it took 7 minutes.

dbarbuzzi:
In the short time these tests have been running, I have seen a range of ~4-7 minutes so 6-10 is not too surprising.

We can consider adding a flag to the runs which would include per-test runtime in the output if we wanted to track that. It wouldn’t otherwise affect the runs’ outcome and would only be visible if you drilled deep into the actions to see the specific step’s output.

SageMoore:
It's worth mentioning that on the 10 minute run, the tests that normally take 4 minutes were taking 7 so it could have just been a slow run overall.

mgoin (approved):

PR #442, opened 2022-06-14
Zero shot multistream

kylesayrs:
This PR adds support for multistream inference of labels by default with the zero shot pipeline. It also fixes some typing errors in `engine.py`.

Here are some sample improved inference times.
Batch = 3 sequences, 1003 labels

```
1 stream : 55 seconds
2 streams: 45 seconds
4 streams: 40 seconds
6 streams: 40 seconds
```

### DIFF
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -561,7 +561,7 @@ class Engine(object):
 
 class Context(object):
     """
-    Contexts can be used to run multiple instances of the MultiModelEngine with the same
+    Contexts can be used to run multiple instances of the (MultiModelEngine) with the same
     scheduler. This allows one scheduler to manage the resources of the system
     effectively, keeping engines that are running different models from fighting over system
     resources.
@@ -575,12 +575,13 @@ class Context(object):
 
     def __init__(
         self,
-        num_cores: int = None,
-        num_streams: int = None,
+        num_cores: Optional[int] = None,
+        num_streams: Optional[int] = None,
     ):
         self._num_cores = _validate_num_cores(num_cores)
         self._num_streams = _validate_num_streams(num_streams, self._num_cores)
         self._scheduler = Scheduler.from_str("elastic")
+        print([self._num_cores, self._num_streams, self._scheduler.value])
         self._deepsparse_context = LIB.deepsparse_context(
             self._num_cores, self._num_streams, self._scheduler.value
         )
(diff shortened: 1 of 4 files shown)


src/deepsparse/transformers/pipelines/nli_text_classification.py
```diff
@@ -218,16 +219,28 @@ def nli_engine_forward(
...
         ]
         labelwise_inputs = [labelwise_input for labelwise_input in labelwise_inputs]
         engine_output = pipeline.engine(labelwise_inputs)  # TODO: Parallelize
         engine_outputs += engine_output
 
+    engine_inputs = numpy.array(engine_inputs)
+
+    engine_outputs = []
+    threads = []
+    for batch_origin_index in range(0, engine_inputs.shape[1], pipeline._batch_size):
```
mgoin, line 235: it isn't clear to me what `engine_inputs.shape[1]` is, could you add a comment for this loop?

src/deepsparse/transformers/pipelines/nli_text_classification.py
```diff
@@ -218,16 +219,28 @@ def nli_engine_forward(
...
+        thread = Thread(
+            target=_engine_forward, args=(batch_origin_index, engine_outputs)
+        )
```
mgoin, line 238: these Threads aren't guaranteed to execute sequentially obviously - is it important for the `engine_outputs += engine_output` appending to be ordered?

src/deepsparse/engine.py
```diff
@@ -561,7 +561,7 @@ def _properties_dict(self) -> Dict:
 
 class Context(object):
     """
-    Contexts can be used to run multiple instances of the MultiModelEngine with the same
+    Contexts can be used to run multiple instances of the (MultiModelEngine) with the same
```
mgoin, line 564: nit: why the parenthesis?

src/deepsparse/transformers/pipelines/zero_shot_text_classification.py
```diff
@@ -159,6 +163,10 @@ def __init__(
                 f"schemes are {ModelSchemes.to_list()}"
             )
 
+        if context is None and model_scheme == ModelSchemes.nli.value:
+            context = Context(num_cores=None, num_streams=2)
```
mgoin, line 167: this is a bit unfortunate, let's talk with @SageMoore

src/deepsparse/transformers/pipelines/nli_text_classification.py
```diff
@@ -218,16 +219,28 @@ def nli_engine_forward(
...
+        threads.append(thread)
+        thread.start()
+
+    for thread in threads:
+        thread.join()
```
mgoin, line 243: it would be nice if we could use a threadpool here since constructing/destructing threads is expensive. especially if the engine only is compiled for X streams, it makes sense to also have X threads calling in at once

mgoin (commented):
@kylesayrs can you add a time measurement comparing this branch to the base on an example so we can see if there is an improvement?

mgoin (changes requested):
.

src/deepsparse/transformers/pipelines/nli_text_classification.py
kylesayrs: Thanks for the catch!

mgoin (approved):
nice implementation, LGTM

PR #449, opened 2022-06-15
transformers pipelines raise exception if tokenizer and config not supplied

bfineran:
raising this exception will avoid hidden issues if incorrect default tokenizers/configs are used


KSGulin (approved):

mgoin (approved):
good call, let's do it!

PR #453, opened 2022-06-16
Feature/damian/ic annotation refactor

dbogunowicz:


src/deepsparse/image_classification/pipelines.py
```diff
@@ -19,6 +19,7 @@
 from typing import Dict, List, Optional, Tuple, Type, Union
 
 import numpy
+import numpy as np
```
dbogunowicz, line 22:
```suggestion
```

src/deepsparse/image_classification/constants.py
```diff
@@ -16,7 +16,1011 @@
 __all__ = [
     "IMAGENET_RGB_MEANS",
     "IMAGENET_RGB_STDS",
+    "IMAGENET_LABELS",
 ]
 
 IMAGENET_RGB_MEANS = [0.485, 0.456, 0.406]
 IMAGENET_RGB_STDS = [0.229, 0.224, 0.225]
+
+IMAGENET_LABELS = [
```
bfineran, line 25: was wondering why this PR was so long :)

bfineran (approved):
LGTM

markurtz:
Noting on here that @bfineran we need to get the class labels and pre processing pipelines into configuration files for the deployment folder. We can hardcode in the code for default, but should always enable the user to override so it's usable for their datasets

src/deepsparse/image_classification/constants.py
dbogunowicz: I was really unhappy about putting those 1000 lines of code effectively, but I guess this is not something super awful.

PR #491, opened 2022-06-30
Pipeline Async run

horheynm:
This PR adds asynchronous execution support to Pipelines.

The idea: If a `ThreadPoolExecutor()` object is passed either during Pipeline creation or when `__call__` is invoked, the pipeline should use the passed threadpool and execute requests in a non-blocking way.

Tests have been added as a part of this PR
local testing can be done like the following:

**Scenario 1**: `threadpool` passed during creation
```python
from concurrent.futures import ThreadPoolExecutor

qa_input = {
        "question": "Who am I?",
        "context": "I am Snorlax",
    }
pipeline = Pipeline.create(task="question-answering", threadpool=ThreadPoolExecutor())
result = pipeline(**qa_input).result()
```

**Scenario 2:** `threadpool` passed when invoking `__call__`
```python
from concurrent.futures import ThreadPoolExecutor

qa_input = {
        "question": "Who am I?",
        "context": "I am Snorlax",
    }
pipeline = Pipeline.create(task="question-answering",)
result = pipeline(**qa_input, threadpool=ThreadPoolExecutor()).result()
```
Noting Dynamic Batch Support was also added as a part of this parent PR, Dynamic Batch PR and it's description : #512 

---
  - https://app.asana.com/0/0/1203643985275976


mgoin:
That's pretty cool, do you know if this works on python3.6?

kazukageorge:
> That's pretty cool, do you know if this works on python3.6?
Nope, got 

```bash
  File "/home/ubuntu/neuralmagic/deepsparse/src/deepsparse/lib.py", line 40, in init_deepsparse_lib
    engine = importlib.util.module_from_spec(spec)
  File "<frozen importlib._bootstrap>", line 571, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 922, in create_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
ImportError: /home/ubuntu/neuralmagic/deepsparse/src/deepsparse/avx2/deepsparse_engine.so: undefined symbol: PyThread_tss_alloc

During handling of the above exception, another exception occurred:

  File "/home/ubuntu/neuralmagic/deepsparse/script.py", line 3, in <module>
  File "/home/ubuntu/neuralmagic/deepsparse/src/deepsparse/__init__.py", line 33, in <module>
    from .engine import *
  File "/home/ubuntu/neuralmagic/deepsparse/src/deepsparse/engine.py", line 72, in <module>
    LIB = init_deepsparse_lib()
  File "/home/ubuntu/neuralmagic/deepsparse/src/deepsparse/lib.py", line 46, in init_deepsparse_lib
    "Unable to import deepsparse engine binaries. "
ImportError: Unable to import deepsparse engine binaries. Please contact support@neuralmagic.com
```

Should I be concerned about this?

src/deepsparse/pipeline.py
```diff
@@ -129,13 +130,20 @@ def __init__(
...
     ):
         self._model_path_orig = model_path
         self._model_path = model_path
         self._engine_type = engine_type
         self._alias = alias
-
         self.context = context
+        self._threadpool = threadpool
+
+        if self._threadpool and not isinstance(self._threadpool, ThreadPool):
```
tdg5, line 142: This seems overly strict to me. I think I'd either cope with loose duck typing or add an abstract class to use as an interface and require that whatever match the interface instead of being a specific class implementation.

src/deepsparse/pipeline.py
```diff
@@ -156,6 +164,8 @@ def __init__(
         self.onnx_file_path = self.setup_onnx_file_path()
         self.engine = self._initialize_engine()
 
+
```
tdg5, line 167:
```suggestion
```

src/deepsparse/pipeline.py
```diff
@@ -193,6 +203,17 @@ def __call__(self, *args, **kwargs) -> BaseModel:
 
         return pipeline_outputs
 
+    def submit(self, *args, **kwargs) -> concurrent.futures:
+        """
+        Asynchronous inference, returns a future
+        """
+        if not self._threadpool:
+            raise ValueError(
+                "Cannot submit jobs to an uninitialized threadpool"
```
tdg5, line 212: This message suggests that there is a thread pool and that it needs to be initialized. But if I understand correctly, there is no thread pool at all, so perhaps the error should reflect that.

tests/deepsparse/test_threads.py
```diff
@@ -0,0 +1,64 @@
...
+
+@pytest.mark.parametrize(
+    "compute_times",
+    [
+        [0.25, 0.5, 0.75, 1],
+        [0.02, 0.05, 0.075, 0.1],
+        [0.1, 0.05, 0.3],
+    ],
+)
+def test_submit_is_asynchronous(threadpool, compute_times):
```
tdg5, line 42: I mentioned this to George, but I'll mention it again. I've had mixed results with timing-based tests like these. I've too often run into scenarios where some CI machine that is running the tests is overworked and it throws off all my timing expectations and leads to spurious test failures at inconvenient times.

Some of this also tests that `ThreadPoolExecutor` does what it claims to do. Perhaps there's an opportunity to introduce a test seam and create an instance of the Pipeline that has a mock version of a `ThreadPoolExecutor` that can be used to check that `ThreadPoolExecutor.submit` is being called with the expected arguments in the expected scenarios and leave testing `ThreadPoolExecutor` to python. Loosening the requirements of what `Pipeline._threadpool` should be would be helpful in this regard.

tdg5: Also, some of this is totally going to depend on what machine the tests are being run on. If there are 10 threads then you may do fine on a quad-core processor, but on a single core processor you could find that the multi-threaded stuff runs more slowly than the synchronous stuff because of thread contention.

src/deepsparse/pipeline.py
```diff
@@ -193,6 +203,17 @@ def __call__(self, *args, **kwargs) -> BaseModel:
 
         return pipeline_outputs
 
+    def submit(self, *args, **kwargs) -> concurrent.futures:
```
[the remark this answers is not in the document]
rahul-tuli: Class deleted!

src/deepsparse/pipeline.py
rahul-tuli: Now we check if a `ThreadPoolExecutor()` is passed

src/deepsparse/pipeline.py
rahul-tuli: Instead of raising the error, we piggyback on sequential execution now

src/deepsparse/threads.py
```diff
@@ -0,0 +1,70 @@
...
+    def submit(self, job) -> Future:
+        """
+        Utility method to submit jobs to the ThreadPool
+
+        :pre: The job must be a callable
+        :param job: Task to be executed in a Thread
+        :return: A Future object, to get the result use Future.result(), note:
+            Future.result() is a blocking call
+        """
+        assert callable(job)
```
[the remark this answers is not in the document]
rahul-tuli: Agreed and removed! according to what python does

tests/deepsparse/test_threads.py
rahul-tuli: Thanks!, changed the test just to look at submit times in async mode vs end to end inference in sync mode

src/deepsparse/pipeline.py
```diff
@@ -156,42 +167,11 @@ def __init__(
...
-
-        # validate outputs format
-        if not isinstance(pipeline_outputs, self.output_schema):
-            raise ValueError(
-                f"Outputs of {self.__class__} must be instances of "
-                f"{self.output_schema} found output of type {type(pipeline_outputs)}"
-            )
-
-        return pipeline_outputs
+    def __call__(self, threadpool=None, *args, **kwargs) -> Union[BaseModel, Future]:
```
kylesayrs, line 170: It might be nice to define run as a locally scoped variable within __call__ to make the code clearer
```
def __call__:
    def run:
```

tests/deepsparse/transformers/test_async_pipeline_qa.py
```diff
@@ -0,0 +1,124 @@
...
+    for _ in range(tries):
+        pipeline(**qa_input)
+    total_sequential_execution_time = time.time() - sequential_start_time
+
+    async_start_time = time.time()
+    for _ in range(tries):
+        threaded_pipeline(**qa_input)
+    async_submit_time = time.time() - async_start_time
+
+    assert (
```
kylesayrs, line 80: May want to add a margin of error here (pytest approx + fixture)

src/deepsparse/pipeline.py
```diff
@@ -129,13 +133,20 @@ def __init__(
...
         self._engine_type = engine_type
         self._alias = alias
-
         self.context = context
+        self._threadpool = threadpool
+
+        if self._threadpool and not isinstance(self._threadpool, ThreadPoolExecutor):
+            raise ValueError(
+                f"Expected a ThreadPoolExecutor object to submit jobs"
+                f"asynchronously but got {self._threadpool} instead"
```
kylesayrs, line 148: {type(self._threadpool)}

bfineran (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -730,6 +805,34 @@ def route_input_to_bucket(
         pass
 
 
+def _initialize_executor_and_workers(
+    batch_size: Optional[int],
+    workers_or_executor: Optional[Union[int, ThreadPoolExecutor]],
+) -> Tuple[Optional[ThreadPoolExecutor], int]:
```
corey-nm, line 811: It might be good to either raise an error or print a warning if the user passes in executor but the schemas aren't joinable/splittable. Right now it would just silently not actually used the initialized threadpool right?

bfineran: executor can be used for a general async run outside of dynamic batch

src/deepsparse/pipeline.py
```diff
@@ -117,25 +120,38 @@ class PipelineImplementation(Pipeline):
...
     ):
         self._model_path_orig = model_path
         self._model_path = model_path
         self._engine_type = engine_type
         self._alias = alias
-
         self.context = context
+        self._batch_size = batch_size
+
+        self.executor, self._num_async_workers = _initialize_executor_and_workers(
```
corey-nm, line 150: It might be good to either raise an error or print a warning if the user passes in executor but the schemas aren't joinable/splittable. Right now it would just silently not actually used the initialized threadpool right?

src/deepsparse/pipeline.py
```diff
@@ -146,7 +162,7 @@ def __init__(
                 )
 
         self._engine_args = dict(
-            batch_size=batch_size,
+            batch_size=self._batch_size or 1,  # bs=1 for dynamic batch
```
corey-nm, line 165: Does the comment mean that when using thread pool executor batch size has to be 1? What happens if user specifies batch size > 1 and a thread pool executor?

bfineran: dynamic batch wouldn't be supported, but async runs would be

bfineran: this brings up a good point though, we could probably add some generic support for batch splitting as long as the given batch size is divisible by the engine batch size

corey-nm: Ah gotcha, I think the interaction between the two modes should be captured, it sounds like:

1. batch_size=None
    a. executor is specified => async runs
    b. executor is not specified => sync static batch
2. batch_size=1
    a. executor is specified => sync dynamic batch, _executor is not used_
    b. executor is not specified => sync dynamic batch
3. batch_size > 1 (does it have to be engine_batch_size?)
    a. executor is specified => sync dynamic batch, _executor is not used_
    b. executor is not specified => sync static batch

Is that correct?

bfineran (approved):

src/deepsparse/yolo/schemas.py
```diff
@@ -14,14 +14,17 @@
 
 
 """
-Input/Output Schemas for Image Segmentation with YOLO
+Input/Output Schemas for Object Detection with YOLO
```
kylesayrs, line 17: nice

src/deepsparse/pipeline.py
```diff
@@ -730,6 +808,39 @@ def route_input_to_bucket(
...
+    if isinstance(workers_or_executor, ThreadPoolExecutor):
+        num_async_workers = workers_or_executor._max_workers  # noqa
+        executor = workers_or_executor
+    elif isinstance(workers_or_executor, int):
+        num_async_workers = max(1, workers_or_executor)
+        executor = ThreadPoolExecutor(max_workers=num_async_workers)
+    elif batch_size is None and workers_or_executor is None:
+        # default num workers to num available cores / 2
+        num_cpu_cores_avaailable = cpu_details()[0]
+        num_async_workers = max(1, num_cpu_cores_avaailable // 2)
```
kylesayrs, line 824: Nice, good catch with the max

kylesayrs (approved):
Beautiful

src/deepsparse/pipeline.py
bfineran: close, currently dynamic batch is only executed when bs=None

PR #495, opened 2022-07-05
Add `utils/artifacts.py` to MANIFEST

mgoin:
The functions needed for the artifact store functionality to work in setup.py aren't included by default in the source distribution

https://packaging.python.org/en/latest/guides/using-manifest-in/

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 include LICENSE
+include utils/artifacts.py
 recursive-include src/deepsparse/avx2 *
 recursive-include src/deepsparse/avx512 *


markurtz (approved):

bfineran (approved):
lgtm pending confirmation of fix

mgoin:
Tested on internal pypi and got the proper error for my macbook
```
OSError: Native Mac is currently unsupported for the DeepSparse Engine. Please run on a Linux system or within a Linux container on Mac. More info can be found in our docs here: https://docs.neuralmagic.com/deepsparse/source/hardware.html
```

PR #514, opened 2022-07-14
Added support to autoinstall yolact dependencies

rahul-tuli:
This PR adds support to autoinstall all yolact dependecies over a separate install pathway
 ~tests to check auto install were added~ removed cause autoinstall changes user environment and this pathway will tested in integration tests

Both of the following ways(CLI + python import) should be sufficient to trigger auto-install of dependencies, both pathways have been tested locally

```bash
deepsparse.instance_segmentation.annotate --help
```

```python
import deepsparse.yolact
```


bfineran (approved):

src/deepsparse/yolact/__init__.py
```diff
@@ -13,8 +13,95 @@
...
+        return None
+    except Exception as dependency_import_error:
+        if raise_on_fail:
+            raise dependency_import_error
+        return dependency_import_error
+
+
+_auto_install_dependencies()
 
-from .annotate import *
```
dbogunowicz, line 17: why we remove `.annotate` from imports? I feel like this may be ok given that we are calling the function directly from the CLI. I will give the green light, but Rahul, please confirm that this is an intended change.

dbogunowicz (approved):

src/deepsparse/yolact/__init__.py
rahul-tuli: Hi yes! It's an intended change.

PR #516, opened 2022-07-21
Adding CustomTaskPipeline

corey-nm:
Notes:
1. I thought the name FunctionPipeline seemed more clear than CustomPipeline, def flexible on name
2. I see that Pipelines are supposed to be created via the functions that call `Pipeline.create` (e.g. `yolo_pipeline(...)`), and then subclasses are supposed to be registered via `Pipeline.register`.  Since the user would be creating this on the fly instead of us creating it for them, it didn't seem to fit into the create/register pattern.

If we want to fit this into the register/create scheme, we'd have to:
1. Create a function to return the pipleline (similar to other pipelines)
2. Define the actual concrete class inside the function (since that is where we first have access to the functions to use)
3. Call Pipeline.register in that function
4. Finally call Pipeline.create also in that function

I thought just having a single class seemed better!


tests/test_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        assert isinstance(input, ImageClassificationInput)
+        assert isinstance(input.images, list)
+        assert all(isinstance(i, np.ndarray) for i in input.images)
+        return [
+            standard_imagenet_transforms(Image.fromarray(img)).unsqueeze(0).numpy()
+            for img in input.images
+        ]
+
+    def postprocess(outputs):
+        # assert len(outputs) == 1
```
corey-nm, line 67: Oh for some reason there are 2 outputs even though there's only 1 input image. Did I do something wrong?

src/deepsparse/pipeline.py
```diff
@@ -565,6 +566,71 @@ def _initialize_engine(self) -> Union[Engine, ORTEngine]:
...
+        **kwargs,
+    ):
+        self._input_schema = input_schema
+        self._output_schema = output_schema
+        self._process_inputs_fn = process_inputs_fn
+        self._process_outputs_fn = process_outputs_fn
+        super().__init__(model_path, *args, **kwargs)
+
+    def setup_onnx_file_path(self) -> str:
+        return model_to_path(self.model_path)
```
corey-nm, line 613: I saw different pipelines do different things. Is this an okay assumption for this class to use `model_to_path`?

bfineran: for now, yes. we can assume model_path is an onnx file/zoo stub. other pipelines may require a directory so need extra processing to load in supplemental configs + locate the model file

src/deepsparse/pipeline.py
```diff
@@ -565,6 +566,71 @@ def _initialize_engine(self) -> Union[Engine, ORTEngine]:
             )
 
 
+class FunctionPipeline(Pipeline):
```
bfineran, line 569: make sure to register the pipeline with the `Pipeline.register` decorator

tests/test_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import OrderedDict
+
+import numpy as np
```
bfineran, line 17: nit: avoid using aliases where possible. prefer `import numpy`

bfineran (changes requested):

tests/test_pipeline.py
corey-nm: the 2 outputs are both numpy arrays, but you're saying it could also have a dictionary output as well?

bfineran: ah shoot, yeah. our IC models do have two outputs. the logits before/after softmax

bfineran: preferring the second for postprocess is safe

src/deepsparse/pipeline.py
```diff
@@ -565,6 +566,71 @@ def _initialize_engine(self) -> Union[Engine, ORTEngine]:
...
+        return self._output_schema
+
+    def process_inputs(
+        self, inputs: BaseModel
+    ) -> Union[List[numpy.ndarray], Tuple[List[numpy.ndarray], Dict[str, Any]]]:
+        return self._process_inputs_fn(inputs)
+
+    def process_engine_outputs(
+        self, engine_outputs: List[numpy.ndarray], **kwargs
+    ) -> BaseModel:
```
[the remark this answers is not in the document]
corey-nm: Do you prefer docstrings specific to this class or were oyu thinking just copy/paste Pipeline docstrings? Right now IDES will grab Pipeline docstrings for these methods since they don't have a new one specified

tests/test_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+    ImageClassificationInput,
+    ImageClassificationOutput,
+)
+from deepsparse.pipeline import FunctionPipeline
+from sparsezoo import Zoo
+from sparsezoo.utils import load_numpy_list
+
+
+from PIL import Image  # isort:skip
+from torchvision import transforms  # isort:skip
```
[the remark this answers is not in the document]
corey-nm: Copied from [tests/deepsparse/image_classification/test_pipelines.py](https://github.com/neuralmagic/deepsparse/blob/main/tests/deepsparse/image_classification/test_pipelines.py#L28)

Want me to fix there as well?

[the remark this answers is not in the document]
bfineran: ah yeah, that's in there due to some dependency chain issues, not important here

src/deepsparse/pipeline.py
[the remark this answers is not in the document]
bfineran: if there's any wording that would change between the parent and this docstring i'd add a new one

tests/test_pipeline.py
```diff
@@ -0,0 +1,96 @@
...
+        (
+            "zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/"
+            "pruned85_quant-none-vnni",
+            224,
+        )
+    ],
+)
+def test_function_pipeline_as_image_classifier(zoo_stub, image_size):
+    non_rand_resize_scale = 256.0 / 224.0  # standard used
+    standard_imagenet_transforms = transforms.Compose(
```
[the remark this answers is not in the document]
corey-nm: Hmm thinking about this more - this test is supposed to be testing the FunctionPipeline. probably don't even need to parameterize the test tbh. Does the main thing we are testing belong in a fixture?

src/deepsparse/pipelines/function_pipeline.py
```diff
@@ -0,0 +1,118 @@
...
+
+import numpy
+from pydantic import BaseModel
+
+from deepsparse.pipeline import Pipeline
+from deepsparse.utils.onnx import model_to_path
+
+
+@Pipeline.register(task="custom")
+class FunctionPipeline(Pipeline):
```
markurtz, line 25: I'd prefer to rename this back to CustomPipeline as originally required -- this is how we'll be referring to it in docs and custom is a much more natural thing to search for than function

src/deepsparse/tasks.py
```diff
@@ -93,7 +93,11 @@ class SupportedTasks:
 
     @classmethod
     def check_register_task(cls, task: str):
-        if cls.is_nlp(task):
+        if task == "custom":
```
markurtz, line 96: We should enable multiple custom pipelines for the user here rather than a hardcoded single one

src/deepsparse/pipelines/function_pipeline.py
```diff
@@ -0,0 +1,118 @@
...
+        into the forward pass of the pipeline engine.
+    :param process_outputs_fn: A callable (function, method, lambda, etc) that maps
+        the list of numpy arrays that are the output of the engine forward pass
+        into an `OutputSchema` object.
+    """
+
+    def __init__(
+        self,
+        model_path: str,
+        input_schema: Type[BaseModel],
```
markurtz, line 65: What happens if the user passes none here in the event that they don't want to specify an input and output schema?

markurtz (changes requested):

src/deepsparse/tasks.py
corey-nm: Would that be via multiple task names/aliases? Right now they would use `Pipeline.create("custom", input_schema=..., process_input_fn=..., ...)` which could support any processing functions/schemas, but only via the "custom" task tag

src/deepsparse/pipelines/function_pipeline.py
corey-nm: schemas are required for other parts of the pipeline right? not sure what happens now but I can add some type checking in the constructor

src/deepsparse/pipelines/function_pipeline.py
corey-nm: Thoughts on adding some more wording to CustomPipeline, maybe `CustomProcessingPipeline` or `CustomFunctionPipeline` or something else? It doesn't feel specific enough about what was custom about it or when a user should use this. Maybe `UnsupportedTaskPipeline` to highlight that it should be used when the task a user wants to run isn't implemented? what use case is this covering for a user?

corey-nm: Maybe `CustomTaskPipeline`?

corey-nm: I renamed to CustomPipeline, I am still in favor of CustomTaskPipeline though 😄

markurtz: I like CustomTaskPipeline :)

src/deepsparse/pipelines/function_pipeline.py
markurtz: Yeah, main point for me here was whether or not we could get by with not requiring an input and output schema for the most basic pipelines. I'd like to have it so the base case is someone could create a custom pipeline by simpling passing in the model and it would work. Additions on top of that would then be pre and post processing functions. And then further additions for additional functionality would be the input and output schemas

corey-nm:
@bfineran getting circular dependency issue now from `pipeline.py` importing `pipelines/schema_helpers.py` via `pipelines/__init__.py` which now also imports `pipelines/custom_pipeline.py` which imports from `pipeline.py` again.

We could either:
1. Move schema_helpers into pipeline.py
2. Move CustomTakePipeline back into pipeline.py
3. Move pipeline.py under pipelines/ (a more breaking change for users)

src/deepsparse/tasks.py
corey-nm: I changed it to `task.startswith("custom")`. That seem sufficient for now? Do we need to capture whatever is after the `custom` for something?

corey-nm:
This currently has the same `transformer.models` import error in the tests. The pipeline tests that I added pass locally for me

src/deepsparse/pipelines/function_pipeline.py
corey-nm: I see the vision yeah that makes sense. I think it'd be straightforward to not require passing the processing functions and just pass input/output to the schemas. It'd probably involve a bunch more changes to not require the schemas - for example the dynamic batch stuff checks various things on the schemas right now, and I know there's at least a little bit of validation done (looks easy to remove though).

I can try out the no functions tomorrow, but the no schemas will probably have to wait until after the release

src/deepsparse/pipeline.py
```diff
@@ -194,6 +194,40 @@ def __call__(self, *args, **kwargs) -> Union[BaseModel, Future]:
...
+
+        :param task: The task name to get the constructor for
+        :return: The class registered to `task`
+        :raises ValueError: if `task` was not registered via `Pipeline.register`.
+        """
+        task = task.lower().replace("-", "_")
+
+        # support any task that has "custom" at the beginning via the "custom" task
+        if task.startswith("custom"):
+            task = "custom"
```
corey-nm, line 213: This new function is pulled from Pipeline.create() (which now uses this function). except for these two lines, which allow support for any task starting with "custom".

Pulling this function out made the unit tests  a bit faster since they don't have to go through compiling the model via Pipeline.create

src/deepsparse/pipelines/function_pipeline.py
markurtz: Can we get a basic pipeline working without dynamic batch and anything else enabled?

tests/test_pipeline.py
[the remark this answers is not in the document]
bfineran: only if we specifically re-use and do not want to re-generate

src/deepsparse/pipeline.py
bfineran: let's avoid breaking this funcitonality out into a new API, we still want users going through `.create`. additionally, for mapping `custom` tasks.

additionally, what's the motivation for making 'custom' a catch-all. if needed, can we add logic for that in SupportedTask / task.py instead? want to avoid too much casework in the base pipeline file (breaking out the bloat in this file is towards the top of our tech debt)

bfineran (changes requested):
looks good overall pending existing comments and adding logic to make schemas for custom pipelines optional

src/deepsparse/pipeline.py
corey-nm: I can make it a private function - this one is mainly intended to make unit testing easier.

The issue with supported task was that Pipeline.create does the check for `task not in _REGISTERED_PIPELINES`. SupportedTask can pass anything that starts with custom, but it'd have to register that task somehow into that dictionary.

src/deepsparse/pipelines/function_pipeline.py
corey-nm: Was pretty easy to do this after all - users now don't need to pass anything other than model path, and they will be interacting directly with engine in this case. see unit tests for example

bfineran (approved):

markurtz (approved):

bfineran (approved):

PR #519, opened 2022-07-22
Remove num sequences

kylesayrs:
No num_sequences

batch size is engine batch size

if labels given, then batch size % labels == 0
if no labels, require batch size == 1

### DIFF
--- a/src/deepsparse/transformers/pipelines/mnli_text_classification.py
+++ b/src/deepsparse/transformers/pipelines/mnli_text_classification.py
@@ -121,6 +121,7 @@ class MnliTextClassificationPipeline(ZeroShotTextClassificationPipelineBase):
                 f"be divisible by the number of labels {len(self._labels)}"
             )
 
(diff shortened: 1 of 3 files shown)


src/deepsparse/transformers/pipelines/mnli_text_classification.py
```diff
@@ -114,6 +115,18 @@ def __init__(
...
 
+        if self._labels and batch_size % len(self._labels) != 0:
+            raise ValueError(
+                f"if static labels are provided then batch_size {batch_size} must "
+                f"be divisible by the number of labels {len(self._labels)}"
+            )
+
+        if not self._labels and batch_size != 1:
+            raise ValueError(
+                "if no static labels are provided then batch_size must be set to 1"
```
bfineran, line 126: noting we'll update to add support for None as well

bfineran (approved):

PR #521, opened 2022-07-26
[BugFix] Server Computer Vision `from_file` fixes

rahul-tuli:
This PR is now a parent PR for 3 different bugs across deepsparse server
The `from_file` bugfix was tested locally for all 3 CV Pipelines.

There is a bug in the current implementation of `deepsparse.server`, when files are sent by client (in CV tasks) the server wasn't actually reading in the files sent over the network, but was looking for `local files(on the server machine) with the same name`, if such a file existed it would use that for inference and return the result thus giving an illusion that everything worked as intended when it actually didn't, on the other hand if a local file with the same name wasn't found a `FileNotFoundError` was thrown on the server side (Which should not happen cause the file is not intended to be on the server machine) as follows:

```bash
FileNotFoundError: [Errno 2] No such file or directory: 'buddy.jpeg'
```

This current PR fixes this, The changes are two fold:
1) Changes were made on the server side to rely on actual filepointers rather than filenames
2) The `from_files` factory method for all CV Schemas was updated to accept an `Iterable` of `FilePointers` rather than a `List[str]`,  `List --> Iterable` change was made to make the function depend on behavior rather than the actual type;
`str --> TextIO` change was made to accept File Pointers, `TextIO` is a generic typing module for File Pointers 
3) Now we rely on `PIL.Image.open(...)` function to actually read in the contents from the file pointer; this library in included with `torchvision`(necessary requirements for all CV tasks) thus does not require any additional installation steps, or additional auto-installation code.

This bug was found while fixing another bug in the documentation as reported by @dbarbuzzi, the docs did not pass in an input with correct batch size, that bug was also fixed as a part of this Pull request

Testing:

Note: Please Test with an image that is not in the same directory from where the server is run, to actually check if the bug was fixed

Step 1: checkout this branch `git checkout server-file-fixes`
Step 2: Start the server
```bash
deepsparse.server \
    --task image_classification \
    --model_path "zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned95-none" \
    --port 5543
```
Step 3: Use the following code to make requests, the returned status must be [200], An HTTP status code 200 means success, change the image path accordingly in the following code:

```python
import requests

url = 'http://0.0.0.0:5543/predict/from_files'
image_path = "/home/dummy-data/buddy.jpeg"
path = [
    image_path,
]
files = [('request', open(img, 'rb')) for img in path]
resp = requests.post(url=url, files=files)
print(resp)
```

Also fixes the following [issue](https://github.com/neuralmagic/deepsparse/pull/521#issuecomment-1211784059) thanks to @dbogunowicz


src/deepsparse/server/main.py
```diff
@@ -150,8 +150,10 @@ def _create_endpoint(endpoint_path: str, from_files: bool = False):
         )
         async def _predict_func(request: input_schema):
             if from_files:
-                files = [file.filename for file in request]
-                request = pipeline.input_schema.from_files(files)
+                request = pipeline.input_schema.from_files(
+                    file.file for file in request
+                )
```
bfineran, line 155: why not just change `file.filename` -> `file.file`?

src/deepsparse/yolo/schemas.py
```diff
@@ -51,9 +51,9 @@ class YOLOInput(BaseModel):
     )
 
     @classmethod
-    def from_files(cls, files: List[str], **kwargs) -> "YOLOInput":
+    def from_files(cls, files: Iterable[TextIO], **kwargs) -> "YOLOInput":
         """
-        :param files: list of file paths to create YOLOInput from
+        :param files: Iterable of file pointers to create YOLOInput from
```
bfineran, line 56: why don't we need any extra processing here?

rahul-tuli: Good catch Ben, forgot to push the commit, it's pushed now

src/deepsparse/server/main.py
bfineran: resolved offline - use of generator

rahul-tuli:
PR changed to Draft mode, to change implementation for `yolo` pipelines as they do not rely on PIL, and we do not want to add an extra dependency

dbogunowicz:
image

image

Tested it locally, also for a batch size of two. Looking great! Thank's for spotting it.

src/deepsparse/image_classification/README.md
```diff
@@ -151,7 +151,7 @@ Making a request:
 import requests
 
 url = 'http://0.0.0.0:5543/predict/from_files'
-path = ['golfish.jpeg', 'golfish.jpeg'] # just put the name of images in here
+path = ['goldfish.jpeg'] # just put the name of images in here
```
dbogunowicz, line 154: I'd still keep two image strings there, so that the user can intuitively understand that they can pass multiple images at once.

dbogunowicz (approved):

rahul-tuli:
Noting these changes revealed another bug in yolact pipelines, where they do not work well `deepsparse.server`. the fix will be added in a separate PR

dbogunowicz:
@rahul-tuli I took a look at the problem. The problem seems to be solved, however, the solution is a tad hacky.

1. What FastAPI does not like, is the type `np.ndarray`. That's because it cannot be directly serialized and sent as a json request. Therefore, I remove explicitly setting this type in any field of the schema. `Optional` fields are ok, no worries about that.
2. As a result, we are now sending over a batch of masks in a new format:
- instead of sending directly `masks : List[Optional[np.ndarray]]` for every `image_batch: List[Any]` as it was before
- we are now sending `masks: List[Optional[List[float]]]`. So we flatten every mask to a list of float values.
3. One problem is, that we need then to retrieve the original dimensions of the mask when we annotate the image. This is not impossible but awkward; we need to basically solve the "equation":
```
masks_flattened[flattened_shape] = mask_retrieved[num_detections, orig_dim, orig_dim]
```
where `flattened_shape` is given, `num_detections` can be inferred from the number of bounding boxes and we solve for `orig_dim`.
My commit [a5bb20c](https://github.com/neuralmagic/deepsparse/pull/521/commits/a5bb20c14ed647b95c73f46bda6fef1c19cabd91) displays how I do the conversion.
3. The code that I've committed has been tested on the server inference as well as using the annotation endpoint.

But this is not the end of the problem. Instance segmentation over an image requires us to send a flattened batch of masks. Given that we have let's say 50 detections and the size of the original mask is 138 x 138, for every (single-batch) image we are sending 952200 floats. This is not scalable and actually prohibitive - during inference, the server returns a green request, but it immediately hangs, because of the massive payload it needs to send over. Let's sync on the solution today, please. (@bfineran)

corey-nm:
Can you add some unit tests to make sure this doesn't regress? Also is this a problem for other pipelines?

src/deepsparse/image_classification/README.md
dbarbuzzi: @dbogunowicz They actually cannot pass two images at once with the example server command just above as it does not pass the `--batch_size` flag with an appropriate value.

This specific doc change is required to make the examples work together as-is and is what prompted @rahul-tuli ’s investigation which led to him finding the issue with the server reading local files.

rahul-tuli: Thank you @dbarbuzzi

dbogunowicz:
@corey-nm I agree with you. We should be testing for regression (but IMO also correctness) for all the integration pipelines.

src/deepsparse/image_classification/README.md
dbogunowicz: @dbarbuzzi you are right.

dbogunowicz:
Commentary to the commit [5159dba](https://github.com/neuralmagic/deepsparse/pull/521/commits/5159dbaf63a847d60d56a1dce581193d7764c553)

## I managed to get Swagger UI to display the payload from inference (batch size = 1)

### Inference Using Python API (input image located on the server)

```python
import requests
import json

url = 'http://0.0.0.0:5543/predict/from_files'
path = ['gettyimages-1401873190-594x594.jpg'] # list of images for inference
files = [('request', open(img, 'rb')) for img in path]
resp = requests.post(url=url, files=files)
annotations = json.loads(resp.text) # dictionary of annotation results
print(annotations)
```

The command above runs fine. The image is being sent, we get a JSON with correct output in a timely fashion, not much to say. I tried testing also with a batch size of 16, and can confirm, no problems there.

### Inference Using Swagger UI (input image located on the server)

Before: process stalling after

```bash
INFO:     127.0.0.1:35748 - "POST /predict/from_files HTTP/1.1" 200 OK.
``` 

So the post request went fine. This means that the problem lies on the swagger UI side.  It is a pretty well-known problem among Swagger UI users: https://github.com/swagger-api/swagger-ui/issues/3832 for instance. The go-to solution is disabling syntax highlighting in Swagger UI. Took me some time, but I found out that there is an argument in the FastAPI python class object that takes swagger UI-related arguments. So the solution to our bug is a one-liner. I disabled syntax highlighting and voila: it takes a couple of seconds for Swagger to process the request, my browser occasionally freezes for a second or so, but then I can freely scroll and inspect the massive output from the pipeline (literally every single pixel of every mask :)), without any problem.

**Note**: Swagger works okay for a single image, but for a batch size of 2 and more (which means feeding more than two images at once through the interface), my web browser crashes.

### Todo:
I am still experiencing problems with running inference when the image is sent from my local machine to the remote server (so the expected pathway). But @rahul-tuli probably has the solution.

rahul-tuli:
> Commentary to the commit [5159dba](https://github.com/neuralmagic/deepsparse/pull/521/commits/5159dbaf63a847d60d56a1dce581193d7764c553)
> 
> ## I managed to get Swagger UI to display the payload from inference (batch size = 1)
> ### Inference Using Python API (input image located on the server)
> ```python
> import requests
> import json
> 
> url = 'http://0.0.0.0:5543/predict/from_files'
> path = ['gettyimages-1401873190-594x594.jpg'] # list of images for inference
> files = [('request', open(img, 'rb')) for img in path]
> resp = requests.post(url=url, files=files)
> annotations = json.loads(resp.text) # dictionary of annotation results
> print(annotations)
> ```
> 
> The command above runs fine. The image is being sent, we get a JSON with correct output in a timely fashion, not much to say. I tried testing also with a batch size of 16, and can confirm, no problems there.
> 
> ### Inference Using Swagger UI (input image located on the server)
> Before: process stalling after
> 
> ```shell
> INFO:     127.0.0.1:35748 - "POST /predict/from_files HTTP/1.1" 200 OK.
> ```
> 
> So the post request went fine. This means that the problem lies on the swagger UI side. It is a pretty well-known problem among Swagger UI users: [swagger-api/swagger-ui#3832](https://github.com/swagger-api/swagger-ui/issues/3832) for instance. The go-to solution is disabling syntax highlighting in Swagger UI. Took me some time, but I found out that there is an argument in the FastAPI python class object that takes swagger UI-related arguments. So the solution to our bug is a one-liner. I disabled syntax highlighting and voila: it takes a couple of seconds for Swagger to process the request, my browser occasionally freezes for a second or so, but then I can freely scroll and inspect the massive output from the pipeline (literally every single pixel of every mask :)), without any problem: image
> 
> **Note**: Swagger works okay for a single image, but for a batch size of 2 and more (which means feeding more than two images at once through the interface), my web browser crashes...
> 
> ### Todo:
> I am still experiencing problems with running inference when the image is sent from my local machine to the remote server (so the expected pathway). But @rahul-tuli probably has the solution.

Noting the code now works with files over the network; tested for all computer vision based pipelines

src/deepsparse/yolact/utils/annotate.py
```diff
@@ -69,7 +69,15 @@ def annotate_image(
         return image
 
     image_res = copy.copy(image)
-
+    num_detections = len(boxes)
+    # len(masks) = num_detections * orig_mask_dim * orig_mask_dim
```
KSGulin, line 73: nit: remove commented out line

KSGulin (approved):
One nit, but otherwise looks good. Good test coverage as well

setup.py
```diff
@@ -84,6 +84,7 @@ def _parse_requirements_file(file_path):
     "onnxruntime>=1.7.0",
     "flask>=1.0.0",
     "flask-cors>=3.0.0",
+    "Pillow>=8.3.2",
```
rahul-tuli, line 87: Test dependency

src/deepsparse/yolact/utils/annotate.py
rahul-tuli: Thank you! Removed.

src/deepsparse/yolact/schemas.py
```diff
@@ -40,7 +41,7 @@ class YOLACTInputSchema(BaseModel, Splittable):
     Input Model for YOLACT
     """
 
-    images: Union[str, numpy.ndarray, List[Union[str, numpy.ndarray]]] = Field(
+    images: Union[str, List[str], List[Any]] = Field(
```
corey-nm, line 44: Can you capture what `Any` is intended to cover either via a comment or adding to description?

src/deepsparse/image_classification/schemas.py
```diff
@@ -44,15 +45,16 @@ class Config:
     @classmethod
     def from_files(cls, files: List[str], **kwargs) -> "ImageClassificationInput":
```
corey-nm, line 46: Should the type of files be changed to `Iterable[TextIO]`?

setup.py
bfineran: will this be auto installed for the Yolo integration now? it is a requirement for from files right?

bfineran (commented):
please confirm in comment or PR description that the readme examples for querying the server from a local file have been tested locally and pass. lgtm after other comments resolved

rahul-tuli:
They have been tested, I mentioned in a comment will update description

setup.py
rahul-tuli: Yes! `pip install deepsparse[yolo]` will install `torchvision` that gives us `Pillow` for free

src/deepsparse/image_classification/schemas.py
rahul-tuli: Good catch! thanks

src/deepsparse/yolact/schemas.py
rahul-tuli: Done!

corey-nm (approved):
🎉

corey-nm (approved):

PR #527, opened 2022-07-29
updated the num_streams property on the engine class to get the acutal num_streams value being used

SageMoore:


SageMoore:
We need to wait for the latest version of nyann to get pulled in, but this diff is ready for review.

src/deepsparse/engine.py
```diff
@@ -284,7 +284,7 @@ def num_streams(self) -> int:
         :return: The max count of streams the current instance can handle
            concurrently.
         """
-        return self._num_streams
+        return self._eng_net.num_streams()
```
mgoin, line 287: could we `self._num_streams = self._eng_net.num_streams()` after the `_eng_net` is constructed so there aren't two possible sources of truth?

bfineran: if not already, can we add documentation on how the engine chooses the default `num_streams`?

SageMoore: I'd prefer to just delete the instance variable all together. It's no longer used after construction.

tlrmchlsmth (approved):

mgoin (approved):
nice

PR #532, opened 2022-08-02
Deleted dynamic installation, install from requirements file in setup.py

kylesayrs:
A quick comment about installation:

Some of the haystack dependencies require wheel building. I've tried adding `wheel` to the beginning of my list of decencies, but I still get the same error `error: invalid command 'bdist_wheel'`

I've tested installation and it works. HF/Transformers are installed during Haystack installation as a dependency, but are replaced by NM/Transformers with dynamic installation.


dbarbuzzi (changes requested):
Would you please add the `haystack` optional to the `pip3 install` lines (total of three) in the [Test Checks GitHub Action](https://github.com/neuralmagic/deepsparse/blob/main/.github/workflows/test-check.yaml)? It looks like it is required for each of the test runs given they all failed during collection with a `ModuleNotFoundError` about it.

corey-nm:
We added the tests importing from transformers & haystack https://github.com/neuralmagic/deepsparse/blob/main/tests/deepsparse/transformers/__init__.py#L18 to force auto install in unit tests so that would be covered. Should that be removed (& maybe that is causing the thing @dbarbuzzi  noticed)?

corey-nm:
Also it looks like the tests are failing because they are using the old version of Model from sparsezoo (i.e. not the 1.1 release)

dbarbuzzi:
> Also it looks like the tests are failing because they are using the old version of Model from sparsezoo (i.e. not the 1.1 release)

From what I can see within the GitHub Actions runs specifically, the tests are not currently getting to that point. During `pytest`'s collection, it hits the error `E   ModuleNotFoundError: No module named 'haystack'` and never actually runs the tests. You can check this by scrolling up to the early output of the test run in the actions log:

https://github.com/neuralmagic/deepsparse/runs/7625871884?check_suite_focus=true#step:4:26

From above that line you can follow the import chain leading to that error. If the auto-install during tests is supposed to occur, it doesn’t seem to be working (at least in that environment).

corey-nm:
> > Also it looks like the tests are failing because they are using the old version of Model from sparsezoo (i.e. not the 1.1 release)
> 
> From what I can see within the GitHub Actions runs specifically, the tests are not currently getting to that point. During `pytest`'s collection, it hits the error `E ModuleNotFoundError: No module named 'haystack'` and never actually runs the tests. You can check this by scrolling up to the early output of the test run in the actions log:
> 
> https://github.com/neuralmagic/deepsparse/runs/7625871884?check_suite_focus=true#step:4:26
> 
> From above that line you can follow the import chain leading to that error. If the auto-install during tests is supposed to occur, it doesn’t seem to be working (at least in that environment).

That import is coming from the tests `__init__.py` file (which we added to force auto install during tests)!

```
tests/deepsparse/transformers/__init__.py:18: in <module>
    from deepsparse.transformers import haystack as _haystack
...
E   ModuleNotFoundError: No module named 'haystack'

**Comment by kylesayrs** [comment]
I've added the haystack option to tests and removed the haystack auto install trigger in tests/deepsparse/transformers/__init__.py

File: setup.py [hunk 1]
```diff
@@ -46,6 +46,14 @@
 # File regexes for binaries to include in package_data
 binary_regexes = ["*/*.so", "*/*.so.*", "*.bin", "*/*.bin"]
 
+
+def _parse_requirements_file(file_path):
+    with open(file_path, "r") as requirements_file:
+        lines = requirements_file.read().splitlines()
```
bfineran, line 52:
```suggestion
        lines = requirements_file.readlines()
```

bfineran (approved):

dbarbuzzi (approved):

PR #537, opened 2022-08-04
update pipeline register error check to pass pipelines registered at runtime

bfineran:
currently supported tasks check-register will fail for tasks that are not included in `SupportedTasks`. That class is only meant to deal with predefined NM pipelines - this PR adds a layer to allow for tasks registered at runtime

**test_plan:**
manual - verified unblock of custom pipeline tutorial


src/deepsparse/tasks.py
```diff
@@ -137,10 +144,11 @@ def check_register_task(cls, task: str):
             # register with Pipeline.register
             import deepsparse.transformers.haystack  # noqa: F401
 
-        else:
+        all_tasks = set(cls.task_names() + (list(extra_tasks or [])))
```
rahul-tuli, line 147:
```suggestion
        all_tasks = set(cls.task_names() + list(extra_tasks or []))
```

src/deepsparse/tasks.py
```diff
@@ -110,7 +110,14 @@ class SupportedTasks:
     all_task_categories = [nlp, image_classification, yolo, yolact, haystack]
 
     @classmethod
-    def check_register_task(cls, task: str):
+    def check_register_task(
+        cls, task: str, extra_tasks: Optional[Iterable[str]] = None
```
rahul-tuli, line 114: Nice to not force extra tasks to be a list.

rahul-tuli (approved):

rahul-tuli (commented):
Reviewed this by mistake

KSGulin (approved):
Some tests are failing. Otherwise LGTM

PR #538, opened 2022-08-04
Fix server CLI tests

dbarbuzzi:
The server CLI tests currently have a presumably inadvertently left-in use of `pdb` and also fail due to some unfortunate logger names in recent uvicorn. This name, `uvicorn.error`, triggers an assert that looks at the server’s output. This assert has been disabled but left in (to ideally re-enable it at a later time); let me know if it should just be deleted instead.

Also, this includes a small change to the Makefile to fix the `nobase` TARGET to properly ignore the base tests with their new organization under `tests/deepsparse`.

### DIFF
--- a/.github/workflows/test-check.yaml
+++ b/.github/workflows/test-check.yaml
@@ -33,11 +33,3 @@ jobs:
         run: pip3 install .[dev,server,haystack]
       - name: Run CLI smoke tests
         run: PYTEST_ARGS="-m smoke" make test TARGETS=cli,nobase
-  examples-smoke-tests:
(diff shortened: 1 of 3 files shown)


rahul-tuli (approved):

bfineran (approved):

dbarbuzzi:
Just pushed one more small change – the examples-smoke-test test-check action has been removed from the GitHub Actions config as the tests that this was designed for have since been removed.

bfineran (approved):

rahul-tuli (approved):

bfineran (approved):

PR #540, opened 2022-08-04
remove outdated examples and update docs to reference latest pipelines

bfineran:
Multiple legacy examples and docs are now out of date for either not using the latest pipelines or patches, or using now defunct sparsezoo apis.  This PR removes them and points all removed examples to their updated doc locations.

Removed examples: classification, detection, flask, transformers, yolo, yolact


rahul-tuli (approved):
Long due!

mgoin (approved):
wow 😤

PR #544, opened 2022-08-05
Add embed_title arg to DPR

kylesayrs:
It was noticed that Tevatron embeds the passage title into the raw text input. This PR adds that functionality to the `DeepSparseDensePassageRetriever` as an argument `embed_title`, which allows a user to generate passage embeddings with embedded titles. This argument is also implemented in the Haystack version of this class.


src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -298,7 +295,15 @@ def embed_documents(self, docs: List[Document]) -> List[numpy.ndarray]:
         :param docs: list of document strings to embed
         :return: list of embeddings for each document
         """
-        passage_inputs = [doc.content for doc in docs]
+        passage_inputs = [
+            f"{doc.meta['title']}{self.title_separator}{doc.content}"
```
corey-nm, line 299: Is prepending title before content with a character between pretty standard for including title?

src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -192,22 +193,22 @@ def __init__(
         emb_extraction_layer: Union[int, str, None] = -1,
         pooling_strategy: str = "cls_token",
         top_k: int = 10,
-        embed_title: bool = False,
         progress_bar: bool = False,
         scale_score: bool = True,
         context: Optional[Context] = None,
+        title_separator: Optional[str] = " ",
```
corey-nm, line 199: Do you envision people actually changing the separator or is this mainly just a yes/no for including?

src/deepsparse/transformers/haystack/nodes.py
corey-nm, line 299: i'd recommend moving this logic to a method, maybe like

```python
def doc_to_passage(self, doc): # you can probably think of a better name for this function haha
    ...
```

As this allows you to unit test the functionality and probably will be formatted slightly better

corey-nm (changes requested):

src/deepsparse/transformers/haystack/nodes.py
kylesayrs: Yes I've seen it done this way by Haystack as well

src/deepsparse/transformers/haystack/nodes.py
kylesayrs: Main use cases would be a space " " or a "<SEP>" token (I've confirmed it gets tokenized properly)

src/deepsparse/transformers/haystack/nodes.py
kylesayrs: Makes sense, I'll make that change

src/deepsparse/transformers/haystack/nodes.py
corey-nm: My 2 cents are the less configuration options ppl have to understand the better. So i'd lean towards just picking one of those two and going back to the embed_title parameter.

Ofc if the title separator option is commonly done within haystack then go with that!

src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -320,6 +307,25 @@ def load(*args, **kwargs):
...
+        # Optionally appends title
+        #
+        # :param docs: documents to turn into inputs
+        # :return: a list of raw text inputs
+        return [
+            f"{doc.meta['title']}{self.title_separator}{doc.content}"
+            if hasattr(doc, "meta")
+            and doc.meta.get("title", None) is not None
+            and self.title_separator is not None
+            else doc.content
```
corey-nm, line 320: Ahh I specifically meant this inside of the loop here, and then in the other function do `passage_inputs = [self._document_to_passage(doc) for doc in docs]`

Then this becomes easy to unit test because you can call it with a single document and make sure the behavior matches expected
```python
doc = Document(meta=None, content="asdf") # or however you create documents
assert obj._document_to_passage(doc) == "asdf"

doc = Document(meta={"title": "qwer", content="asdf") # or however you create documents
assert obj._document_to_passage(doc) == "qwer asdf"
```

And it also makes it less dependendant on receiving a lists of documents as input. for example if you get a dictionary of documents or some other data structure, only the iteration of that structure has to change while the single document processing function (_document_to_passage) doesn't have to change at all

kylesayrs: Ah yeah, that's better. Making those changes...

src/deepsparse/transformers/haystack/nodes.py
kylesayrs: Ha that's fair enough. I'll go back to embed_title (IIRC Haystack implements it the same way with a space)

src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -320,6 +306,25 @@ def load(*args, **kwargs):
             "DeepSparse Engine does not support loading from files"
         )
 
+    def _document_to_passage_input(self, document: Document) -> str:
+        # Optionally appends title
+        #
+        # :param document: document to turn into raw text input
+        # :return: raw text input of document title and content
```
corey-nm, line 313:
```suggestion
        """
        Optionally appends title

        :param document: document to turn into raw text input
        :return: raw text input of document title and content
        """
```

kylesayrs: Because this is a private method, we use # to avoid the docstring being picked up by sphinx (at least that's what I've been taught)

src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -320,6 +306,25 @@ def load(*args, **kwargs):
...
+        if (
+            hasattr(document, "meta")
+            and document.meta.get("title", None) is not None
```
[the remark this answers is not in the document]
kylesayrs: This attempts to account for users who write some preprocessing script that leaves document.meta["title"] = None. The proposed change would silently append "None {my content}"

src/deepsparse/transformers/haystack/nodes.py
corey-nm: Oh hmm I think sphinx will avoid things prepended with `_`? IDEs won't show the docstring to devs if you use `#` though. Is this a NM thing?

src/deepsparse/transformers/haystack/nodes.py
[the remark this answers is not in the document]
corey-nm: Ahhhh gotcha okay, thanks for the clarification

src/deepsparse/transformers/haystack/nodes.py
bfineran: we've hd some issues in the past - regardless - we standardize to not including pydoc strings in 'private' functions

src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -250,15 +244,11 @@ def __init__(
...
                 f"({max_seq_len_passage})"
             )
-        if embed_title:
-            raise ValueError(
-                "DeepSparseDensePassageRetriever does not support embedding titles"
-            )
 
         if self.context is None:
             self.context = Context(
                 num_cores=None, num_streams=4
```
bfineran, line 250:
```suggestion
                num_cores=None
```

src/deepsparse/transformers/haystack/nodes.py
```diff
@@ -250,15 +244,11 @@ def __init__(
...
-            )  # arbitrarily choose 4
+            )  # arbitrarily choose 4, to be replaced with dynamic batch
```
bfineran, line 251:
```suggestion
            )
```

kylesayrs:
Another thing that might be worth adding is the PR description from https://github.com/neuralmagic/deepsparse/pull/490 as a README.md in the Haystack directory

corey-nm (approved):
lgtm!

bfineran (approved):

PR #545, opened 2022-08-05
parse messy benchmark 'engine' output

horheynm:
Parse messy output from deepsparse.benchmark
Also added commit sha info

To run: `deepsparse.benchmark zoo:nlp/text_classification/bert-base/pytorch/huggingface/sst2/base-none -x benchmark.json`

Before:
https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/benchmark/img/json_output.png

After:
Screen Shot 2022-08-26 at 9 22 36 AM

### DIFF
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -414,9 +414,9 @@ def benchmark_model(
 def _parse_export_dict_engine_key(payload: Dict) -> Dict:
     """Extract metadata from str(Model)"""
 
-    formatted_metadata = "engine: " + payload.get("engine", "").replace("\t", "").replace(
-        ":", "", 1
-    )
+    formatted_metadata = "engine: " + payload.get("engine", "").replace(
+        "\t", ""
+    ).replace(":", "", 1)
     return {**payload, **yaml.load(formatted_metadata, Loader=Loader)}
 
 
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -95,6 +95,7 @@ import argparse
 import json
 import logging
 import os
+import subprocess
 from typing import Dict
 
 from deepsparse import Scheduler, __version__, compile_model
@@ -380,6 +381,7 @@ def benchmark_model(
     export_dict = {
         "engine": str(model),
         "version": __version__,
+        "commit": get_latest_commit_sha_for_path(),
         "orig_model_path": orig_model_path,
         "model_path": model_path,
         "batch_size": batch_size,
@@ -404,13 +406,34 @@ def benchmark_model(
 
 
 def _parse_export_dict_engine_key(payload):
-
+    """Extract metadata from str(Model)"""
     engine_split = payload["engine"].replace("\n\t", ":").split(":")
     engine = engine_split[0]
     payload["engine"] = engine
     for i in range(2, len(engine_split) - 1, 2):
         payload[engine_split[i]] = engine_split[i + 1].strip()
     return payload
+
+
+def get_latest_commit_sha_for_path(
+    relative_path: str = "",
+    repo_path: str = os.getcwd(),
+) -> str:
+    """Use git log to fetch the most recent commit sha that impacted the given
+    relative path."""
+    if relative_path:
+        args = [
+            "git",
+            "log",
+            "--pretty=format:%H",
+            "--max-count=1",
+            "--",
+            relative_path,
+        ]
+    else:
+        args = ["git", "log", "--pretty=format:%H", "--max-count=1"]
+    commit_sha = subprocess.check_output(args, cwd=repo_path).decode("utf-8")
+    return commit_sha
 
 
 def main():
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -95,10 +95,9 @@ import argparse
 import json
 import logging
 import os
-import subprocess
 from typing import Dict
 
-from deepsparse import Scheduler, __version__, compile_model
+from deepsparse import Scheduler, __version__, compile_model, generated_version
 from deepsparse.benchmark.ort_engine import ORTEngine
 from deepsparse.benchmark.stream_benchmark import model_stream_benchmark
 from deepsparse.cpu import cpu_architecture
@@ -381,7 +380,7 @@ def benchmark_model(
     export_dict = {
         "engine": str(model),
         "version": __version__,
-        "commit": get_latest_commit_sha_for_path(),
+        "commit": generated_version.revision,
         "orig_model_path": orig_model_path,
         "model_path": model_path,
         "batch_size": batch_size,
@@ -413,23 +412,6 @@ def _parse_export_dict_engine_key(payload):
     for i in range(2, len(engine_split) - 1, 2):
         payload[engine_split[i]] = engine_split[i + 1].strip()
     return payload
-
-
-def get_latest_commit_sha_for_path(
-    relative_path: str = "",
-    repo_path: str = os.getcwd(),
-) -> str:
-    """Use git log to fetch the most recent commit sha that impacted the given
-    relative path."""
-    args = ["git", "log", "--pretty=format:%H", "--max-count=1"]
-    if relative_path:
-        args += [
-            "--",
-            relative_path,
-        ]
-
-    commit_sha = subprocess.check_output(args, cwd=repo_path).decode("utf-8")
-    return commit_sha
 
 
 def main():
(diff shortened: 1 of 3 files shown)


src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -392,13 +394,42 @@ def benchmark_model(
...
         with open(export_path, "w") as out:
-            json.dump(export_dict, out, indent=2)
-
-    return export_dict
+            json.dump(parsed_export_dict, out, indent=2)
+
+    return parsed_export_dict
+
+
+def _parse_export_dict_engine_key(payload):
```
markurtz, line 408: @kazukageorge it would be good to migrate this over to a regex implementation to make it more obvious what the expected input and subsequent output are

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -392,13 +394,42 @@ def benchmark_model(
...
+    """Extract metadata from str(Model)"""
+    engine_split = payload["engine"].replace("\n\t", ":").split(":")
+    engine = engine_split[0]
+    payload["engine"] = engine
+    for i in range(2, len(engine_split) - 1, 2):
+        payload[engine_split[i]] = engine_split[i + 1].strip()
+    return payload
+
+
+def get_latest_commit_sha_for_path(
```
markurtz, line 418: So we're not going to have git available especially on the pip install case. The sha hash from the engine side should be printed out in the engine statement and we should just need to grab that

markurtz (changes requested):

src/deepsparse/benchmark/benchmark_model.py
kazukageorge: The sha from the engine side is not Wand?

markurtz: With the engine startup message it prints out a sha from the wand side. This is what we want to store and report. We do not have git available when we pip install so will not be able to run any git commands.

derekk-nm: FWIW, it would be nice to have the length of the commit id be consistent w/ the information previously provided, so the commit id is the first 8 characters of the full commit SHA string.

kazukageorge: Ok tried to capture stdout in terminal in two ways. The way Im capturing does not work (perhaps its bc its a different process? Not sure how deepsparse works in the backend)

Way 1. Context Manager -- will close property even if error is induced
Screen Shot 2022-08-08 at 10 01 12 AM

Output
Screen Shot 2022-08-08 at 10 05 02 AM

Way 2. Force output to sys stdout
Screen Shot 2022-08-08 at 10 01 30 AM

Output
Screen Shot 2022-08-08 at 10 01 37 AM
In the output 11 doesnt  get printed (gets captured), so everything in that scope should be captured, but deepsparse bypasses this. 

TLDR: Can we just add the sha info in str(model)?

derekk-nm: Am I correct that this is already available to you?
```
from deepsparse import generated_version
commit_id = generated_version.revision
```

mgoin: This is what I was going to ask, all of this information parsed out from the Engine str is information we can query from the Engine class itself. Especially if you aren't finding all the information in the str representation useful, we can just extract the values needed from the Engine

kazukageorge: Youre a life saver thanks

src/deepsparse/benchmark/benchmark_model.py
tdg5: OOO maybe you can get away with `yaml.load`... that would help to make it explicit what we expect to find in the engine key and make it more flexible in the event that the engine key is not normal (e.g. property order varies)

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -379,7 +388,6 @@ def benchmark_model(
 
     export_dict = {
         "engine": str(model),
```
corey-nm, line 390: Could you do most of the parsing from below here instead?
```suggestion
        **yaml.load("engine: " + str(model).replace("\t", "").replace(":", "", 1)),
```

corey-nm: Alternatively you could use the `model._properties_dict()` (https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/engine.py#L545) directly to get the dictionary

mgoin: +1 for the `_properties_dict` as the output structure of `str(model)` is subject to change

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -392,13 +400,24 @@ def benchmark_model(
...
+    return parsed_export_dict
+
+
+def _parse_export_dict_engine_key(payload: Dict) -> Dict:
+    """Extract metadata from str(Model)"""
+
+    formatted_metadata = "engine: " + payload["engine"].replace("\t", "").replace(
+        ":", "", 1
+    )
+    return {**payload, **yaml.load(formatted_metadata, Loader=Loader)}
```
rahul-tuli, line 420: could we use dict.update here instead of unpacking? I think it would make the intent more clearer

src/deepsparse/engine.py
```diff
@@ -540,6 +542,8 @@ def _validate_inputs(self, inp: List[numpy.ndarray]):
 
     def _properties_dict(self) -> Dict:
         return {
+            "version": __version__,
+            "commit": generated_version.revision,
```
rahul-tuli, line 546: nice!

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -95,9 +95,18 @@
 import json
 import logging
 import os
+
+import yaml
+
```
rahul-tuli, line 100: as yaml is in stdlib could we not add a line break before importing it?

src/deepsparse/benchmark/benchmark_model.py
kazukageorge: `_properties_dict` will populate the metadata from a result from a func, right now its populating straight from the input arg.

str(model) has engine key with messy output which is the reason for this pr.

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -392,13 +400,24 @@ def benchmark_model(
...
+            json.dump(parsed_export_dict, out, indent=2)
 
-    return export_dict
+    return parsed_export_dict
+
+
+def _parse_export_dict_engine_key(payload: Dict) -> Dict:
+    """Extract metadata from str(Model)"""
+
+    formatted_metadata = "engine: " + payload["engine"].replace("\t", "").replace(
```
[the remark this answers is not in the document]
kazukageorge: ok

src/deepsparse/benchmark/benchmark_model.py
kazukageorge: Is dict pass by reference? I did this so i wont mutate if pass by ref

src/deepsparse/benchmark/benchmark_model.py
kazukageorge: alright this is the result from lint.
Made without \n and still got the above

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -392,13 +400,24 @@ def benchmark_model(
...
         _LOGGER.info("Saving benchmark results to JSON file at {}".format(export_path))
         with open(export_path, "w") as out:
-            json.dump(export_dict, out, indent=2)
+            json.dump(parsed_export_dict, out, indent=2)
 
-    return export_dict
+    return parsed_export_dict
+
+
+def _parse_export_dict_engine_key(payload: Dict) -> Dict:
```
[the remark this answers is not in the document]
kazukageorge: thanks done

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -391,14 +390,15 @@ def benchmark_model(
         "num_streams": num_streams,
         "benchmark_result": benchmark_result,
     }
+    parsed_export_dict = {**model._properties_dict(), **export_dict}
```
kazukageorge, line 393: First populate kwargs in _properties_dict(), if there is an intersection in export_dict, because its values are closer to the source of truth (straight from the input args, instead of variable from class -- can be mutated by mistake), overwrite it.

corey-nm (approved):
nice!!

tdg5 (approved):

mgoin:
@kazukageorge is this PR still needed or has it been superseded by other things?

mgoin:
Stale/not needed

PR #548, opened 2022-08-09
Fix critical bug in YOLACT inference pipeline

dbogunowicz:
## Description

It turned out that the one line that converts from RGB -> BGR is not necessary - after all, OpenCV reads the image as BGR.
The bug was not salient when I used the image of elephants (`thailand.jpeg`) as my testing image. 
But now, when working with pictures of oranges, the bug has revealed itself, since the order of color channels for fruit classification does really matter. 

Also, the quality of predictions on `thailand.jpeg` has improved (also consistent with the prediction done by the original yolact repo).

Additionally, resetting the yolact parameters (in `schema`) to default parameters from `yolact` repository, to make future testing easier (we have consistent logic between deepsparse and yolact).

### Before the PR:
The classes of the detected objects are clearly wrong. 

image

### After the PR:
The classes of the detected objects are correct and the results are the same as the ones generated using `yolact` repository.

Video generated using `yolact` repository:

```bash
--trained_model
zoo:cv/segmentation/yolact-darknet53/pytorch/dbolya/coco/base-none
--engine
deepsparse
--video
/home/damian/deepsparse_copy/src/videos/pexels-thiago-zanutim-lucas-10576689.mp4:data/yolact_example_out_0.mp4
```

image

Video generated using deepsparse engine:

```bash
--model-filepath
zoo:cv/segmentation/yolact-darknet53/pytorch/dbolya/coco/base-none
--source
/home/damian/deepsparse_copy/src/videos/pexels-thiago-zanutim-lucas-10576689.mp4
```

image

NOTE:
The reviewers may be anxious about how bad the quality of predictions displayed above (even for the "baseline" model from yolact). Yolact inference is very sensitive to the choice of parameters. After playing with some parameters, we can get a much higher amount of predictions for instance:

image


tdg5 (approved):

corey-nm (approved):

mgoin (approved):
nice catches and writeup

PR #549, opened 2022-08-09
Add warning that elastic scheduler is deprecated

mgoin:
```
deepsparse.benchmark zoo:cv/classification/efficientnet-b0/pytorch/sparseml/imagenet/base-none -s elastic
...
2022-08-09 13:51:32 deepsparse.engine WARNING  Scheduler.elastic is an alias for Scheduler.multistream - will be deprecated in 1.3
```

### DIFF
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -18,9 +18,9 @@ Code related to interfacing with a Neural Network in the DeepSparse Engine using
 
 import logging
 import time
+import warnings
 from enum import Enum
 from typing import Dict, Iterable, List, Optional, Tuple, Union
-import warnings
 
 import numpy
 from tqdm.auto import tqdm
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -140,6 +140,11 @@ def _validate_scheduler(scheduler: Union[None, str, Scheduler]) -> Scheduler:
             "unsupported type for scheduler: {} ({})".format(scheduler, type(scheduler))
         )
 
+    if scheduler == Scheduler.elastic:
+        _LOGGER.warn(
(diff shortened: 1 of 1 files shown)


src/deepsparse/engine.py
```diff
@@ -140,6 +140,11 @@ def _validate_scheduler(scheduler: Union[None, str, Scheduler]) -> Scheduler:
             "unsupported type for scheduler: {} ({})".format(scheduler, type(scheduler))
         )
 
+    if scheduler == Scheduler.elastic:
+        _LOGGER.warn(
```
rahul-tuli, line 144: Could we use python's inbuilt `warnings` module instead of a LOGGER,  it is the recommended way after Python 3.6 and gives a nicer message, the code would look something like:
`import warnings`
`warnings.warn("Scheduler.elastic is an alias for Scheduler.multistream", DeprecationWarning, stacklevel=2)`


And the actual warning would look like:
```bash
/XXXXXX/projects/deepsparse/local-work.py:8: DeprecationWarning: Scheduler.elastic is an alias for Scheduler.multistream
```

corey-nm (dismissed):
lgtm!

src/deepsparse/engine.py
rahul-tuli: My bad, should be a `FutureWarning` instead

rahul-tuli (dismissed):

rahul-tuli (approved):

src/deepsparse/engine.py
mgoin: good call, thank you

mgoin:
Runtime team decided we can actually keep the elastic scheduler to repurpose it - stay tuned...

PR #553, opened 2022-08-09
Make artifact download use src/deepsparse/version.py as default

mgoin:
It was possible to be stuck in a loop because a previously downloaded generated_version.py would be used as the source of truth over the version.py that gets updated within deepsparse

### DIFF
--- a/utils/artifacts.py
+++ b/utils/artifacts.py
@@ -53,10 +53,16 @@ def get_release_and_version(package_path: str) -> Tuple[bool, str, str, str, str
     """
     Load version and release info from deepsparse package
     """
+    # deepsparse/src/deepsparse/version.py always exists, default source of truth
+    version_path = os.path.join(package_path, "version.py")
 
-    version_path = os.path.join(package_path, "generated_version.py")
-    if not os.path.exists(version_path):
-        version_path = os.path.join(package_path, "version.py")
+    # If deepsparse/engine-version.txt exists, use the generated_version
+    engine_version_path = os.path.abspath(
+        os.path.join(package_path, os.pardir, os.pardir, "engine-version.txt")
+    )
+    gen_version_path = os.path.join(package_path, "generated_version.py")
+    if os.path.exists(engine_version_path) and os.path.exists(gen_version_path):
+        version_path = gen_version_path
 
     # exec() cannot set local variables so need to manually
     locals_dict = {}
--- a/utils/artifacts.py
+++ b/utils/artifacts.py
@@ -30,7 +30,7 @@ __all__ = [
 
 def parse_args():
     parser = argparse.ArgumentParser(
-        description="Benchmark ONNX models in the DeepSparse Engine"
+        description="Download binaries for the engine"
     )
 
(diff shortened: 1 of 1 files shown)


utils/artifacts.py
```diff
@@ -53,10 +51,16 @@ def get_release_and_version(package_path: str) -> Tuple[bool, str, str, str, str
...
     """
+    # deepsparse/src/deepsparse/version.py always exists, default source of truth
+    version_path = os.path.join(package_path, "version.py")
 
-    version_path = os.path.join(package_path, "generated_version.py")
-    if not os.path.exists(version_path):
-        version_path = os.path.join(package_path, "version.py")
+    # If deepsparse/engine-version.txt exists, use the generated_version
+    engine_version_path = os.path.abspath(
+        os.path.join(package_path, os.pardir, os.pardir, "engine-version.txt")
```
rahul-tuli, line 59: Do we not rely on `version.py` at all going forward?

rahul-tuli (approved):

utils/artifacts.py
mgoin: We actually use `version.py` more often now after this change. The only way to get `deepsparse/engine-version.txt` is if you're working on the internal engine interface, which is how you get the `generated_version.py` as well (or through the artifact store).
Another approach to this would be to not gather `generated_version.py` from the artifact at all.. I'll need to make sure this doesn't break anything, might be better than this method

corey-nm, line 59: How does checking the presence of engine-version.txt break the cycle you mention in the PR description? Is it not possible to have an old engine-version.txt lying around?

mgoin: `generated_version.py` is needed for other functions of deepsparse to work, so not a valid solution

utils/artifacts.py
```diff
@@ -53,10 +51,16 @@ def get_release_and_version(package_path: str) -> Tuple[bool, str, str, str, str
     """
     Load version and release info from deepsparse package
     """
+    # deepsparse/src/deepsparse/version.py always exists, default source of truth
+    version_path = os.path.join(package_path, "version.py")
```
corey-nm, line 55: Also from looking at version.py, it seems like it also pulls version from generated_version.py if it exists - does that need to be changed as well?

utils/artifacts.py
mgoin: The key thing is `generated_version.py` is pulled down with the binaries you download, `engine-version.txt` is not. The only current way to get an `engine-version.txt` is if you are creating the binaries locally yourself. We don't want to disrupt this flow by always using the deepsparse repo as source of truth.
If we wanted to do the safest approach, I think that would be to compare the versions in `version.py` and `generated_version.py`, only using `version.py` if it is newer. It'll be a bit messier but should work

utils/artifacts.py
mgoin: This is the intended behavior still, look at my previous response to lay out the flow a bit more.

corey-nm (approved):

PR #556, opened 2022-08-10
Revert Haystack to Auto Install

kylesayrs:
This PR reverts the changes made in https://github.com/neuralmagic/deepsparse/pull/532. It was found that installing Haystack dependencies through extras was not feasible. Haystack requires HF transformers as a dependency. However, having HF transformers breaks the auto install of NM transformers, as there is no good way to dynamically reload modules in python. There is also no good way to install Haystack in extras without installing its dependencies.

Given these blocks, reverting back to dynamical install fixes these issues by controlling which dependencies are need by Haystack through a requirements.txt file. We can then remove HF transformers from this dependency list, thus avoiding the issue of dynamic module reload.

### DIFF
--- a/src/deepsparse/transformers/haystack/__init__.py
+++ b/src/deepsparse/transformers/haystack/__init__.py
@@ -28,14 +28,7 @@ import deepsparse as _deepsparse
 
 _HAYSTACK_PREFERRED_VERSION = "1.4.0"
 _HAYSTACK_EXTRAS = "[all]"
-_HAYSTACK_REQS_PATH = _os.path.join(
-    *[
-        f"{_deepsparse.__path__[0]}",
-        "transformers",
-        "haystack",
-        "haystack_reqs.txt",
-    ]
-)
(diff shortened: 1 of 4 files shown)


src/deepsparse/transformers/haystack/__init__.py
```diff
@@ -19,6 +19,112 @@
...
+_HAYSTACK_PREFERRED_VERSION = "1.4.0"
+_HAYSTACK_EXTRAS = "[all]"
+_HAYSTACK_REQS_PATH = _os.path.join(
+    *[
+        f"{_deepsparse.__path__[0]}",
+        "transformers",
+        "haystack",
+        "haystack_reqs.txt",
+    ]
+)
```
corey-nm, line 38: Does this work?
```suggestion
_HAYSTACK_REQS_PATH = _os.path.join(os.path.dirname(__file__), "haystack_reqs.txt")
```

kylesayrs: Seems more robust, making those changes

corey-nm (approved):
nice! 🎉

bfineran (approved):

PR #557, opened 2022-08-10
Pipeline can dynamic import task from arbitrary python file

corey-nm:
Tested with:
- The new unit tests added to PR
- running `deepsparse.server --task import:my_pipeline --model_path ...` where my_pipeline.py had this contents
```python
from deepsparse.pipeline import Pipeline
from deepsparse.transformers.pipelines.question_answering import (
    QuestionAnsweringPipeline,
)

TASK = "my_qa_task"
Pipeline.register(TASK)(QuestionAnsweringPipeline)
```
- running `deepsparse.server --task import:deepsparse.my_pipeline --model_path ...` from a different directory
- running `deepsparse.server --config_file config.yaml` where config.yaml had this contents:
```yaml
models:
    - task: "import:my_pipeline"
      model_path: zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none
      batch_size: 1
      alias: question_answering/base
```


KSGulin (approved):
Solid testing. LGTM

rahul-tuli (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -1353,3 +1361,76 @@ def zero_shot_text_classification_pipeline(*args, **kwargs) -> "Pipeline":
         with 2 streams to make use of parallel inference of labels
     """
     return Pipeline.create("zero_shot_text_classification", *args, **kwargs)
+
+
+def _dynamic_import_task(module: str) -> str:
```
bfineran, line 1366: let's get this moved to a separate file as a helper function

src/deepsparse/pipeline.py
```diff
@@ -1353,3 +1361,76 @@ def zero_shot_text_classification_pipeline(*args, **kwargs) -> "Pipeline":
...
+        )
+
+    # add parent_dir to sys.path so we can import the file as a module
+    sys.path.append(os.curdir)
+    if parent_dir:
+        print(f"Adding {parent_dir} to sys.path")
+        sys.path.append(parent_dir)
+
+    # do the import
+    print(f"Importing '{module_name}'")
```
bfineran, line 1403: nit: prefer info logs to `print` in api side code

tests/deepsparse/pipelines/dynamic_import_modules/ok.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: nit: file name - prefer `valid_pipeline.py` or something a bit more descriptive

src/deepsparse/pipeline.py
```diff
@@ -1353,3 +1361,76 @@ def zero_shot_text_classification_pipeline(*args, **kwargs) -> "Pipeline":
...
+    ```
+
+    NOTE: this modifies `sys.path`.
+
+    :raises FileNotFoundError: if path does not exist
+    :raises RuntimeError: if the imported module does not contain `TASK`
+    :raises RuntimeError: if the module doesn't register the task
+    :return: The task from the imported module.
+    """
+    parent_dir, module_name = _module_to_dir_and_name(module)
```
bfineran, line 1389: what we can do to allow for more flexible locations/ways to provide the custom module file.  Right now it seems like the file needs to be nested somewhere inside the directory the python process is called from? 

Thoughts on trying to import first (ie user would have already exported the custom file into the python path) and then trying to dynamically find and add the file to path on fail?

bfineran (changes requested):
LGTM overall, few comments.

Additionally - let's add some mechanism (ie global variable, environment variable, etc) that we can control to block the loading of `import:` pipelines.  Will give us an easy way in the future if there's any security considerations with certain deployment scenarios

src/deepsparse/pipeline.py
corey-nm: It should allow any path on the file system - that's what the `sys.path.append` will do. E.g. if you specify `import:/home/corey/deepsparse/a.py` parent_dir will be `/home/corey/deepsparse` and it will should add it to the python path

corey-nm: err i forgot we changed to module names instead of paths, so the above should be `import:home.corey.deepsparse.a`

bfineran: ah yeah makes sense - still though, it looks like that would translate to `home/corey/deepsparse` and `a.py` (rather than having the preceding `/` to home)

corey-nm: yeah great call out - missing that test case. in this case they'd have to add a '.' in front of path i think, which is a bit awkward. thoughts on going back to file paths? I think that makes it a bit more clear what a user can enter

bfineran: let detect and support either two cases:

1. user gives `module` path (ie `.` separated, does not end in `.py`) -> we assume the module is valid in the pythonpath and run importlib
2. user gives file path (ends in `.py`, optionally '/' separated) -> we append the file to python path and then run importlib

corey-nm: I like it, will do!

corey-nm: Right now it only keys off of `endswith(".py")` - depending on anything else seemed to have too many edge cases

bfineran: looks good

bfineran (approved):
LGTM - do we have a working readme for this pathway we can update? if not, lets get a task for a simple readme added to our backlog

bfineran:
@corey-nm take a look at the failing test

corey-nm:
> @corey-nm looks like a test is failing

Looks like a flaky test - rerunning pipeline

corey-nm:
> LGTM - do we have a working readme for this pathway we can update? if not, lets get a task for a simple readme added to our backlog

Added to asana backlog

rahul-tuli (approved):

PR #559, opened 2022-08-11
Update `deepsparse.server` to display task list programmatically

rahul-tuli:
Before this PR invoking `deepsparse.server --help` gave a hardcoded task list as follows:

```bash
  --task TEXT                     The task the model_path is serving. For
                                  example, one of: question_answering,
                                  text_classification, token_classification.
                                  Ignored if config file is supplied.
```

This PR updates `--task` field to a `click.Choice` type argument that gives good error handling for unsupported tasks + a nice display message with all tasks:

```bash
 --task [custom|question_answering|qa|text_classification|glue|sentiment_analysis|token_classification|ner|zero_shot_text_classification|embedding_extraction|image_classification|yolo|yolact|information_retrieval_haystack|haystack]
                                  The task the model_path is serving.
```

### DIFF
--- a/src/deepsparse/server/main.py
+++ b/src/deepsparse/server/main.py
@@ -48,10 +48,10 @@ Options:
   --log_level TEXT      Sets the logging level. Defaults to info.
   --config_file TEXT    Configuration file containing info on how to serve the
                         desired models.
-  --task TEXT           The task the model_path is serving. For example, one
-                        of: question_answering, text_classification,
-                        token_classification. Ignored if config file is
-                        supplied.
+  --task [custom|question_answering|qa|text_classification|glue|sentiment_analysis|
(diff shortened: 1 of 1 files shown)


src/deepsparse/server/main.py
```diff
@@ -273,11 +276,9 @@ def server_app_factory():
 )
 @click.option(
     "--task",
-    type=str,
+    type=click.Choice(SupportedTasks.task_names()),
```
corey-nm, line 279: Does this need to be optional or something to support using config file without specifying task?

corey-nm: Actually nvm i see default is none

corey-nm (approved):
easy change! 🎉

PR #573, opened 2022-08-15
Add embed_title arg to DPR (#544)

kylesayrs:
It was noticed that Tevatron embeds the passage title into the raw text input. This PR adds that functionality to the DeepSparseDensePassageRetriever as an argument embed_title, which allows a user to generate passage embeddings with embedded titles. This argument is also implemented in the Haystack version of this class.

https://github.com/neuralmagic/deepsparse/pull/544


rahul-tuli (approved):

KSGulin (approved):
Good stuff

PR #574, opened 2022-08-15
force download of tokenizer_config.json when available

bfineran:
tokenizer_config.json is needed to define special characters in the tokenizer (as opposed to tokenizer.json used for determining tokenizer class).  future refactor will make deployment directory a first class citizen across all pipelines without casework for any files. This PR adds support for downloading the tokenizer_config file in the manual parsing of the deployment directory

**test_plan:**
tested manually with https://github.com/neuralmagic/sparsezoo/pull/215 in env. all 3 deployment files correctly downloaded and used to load tokenizer

### DIFF
--- a/src/deepsparse/transformers/helpers.py
+++ b/src/deepsparse/transformers/helpers.py
@@ -45,6 +45,7 @@ _LOGGER = get_main_logger()
 _MODEL_DIR_ONNX_NAME = "model.onnx"
 _MODEL_DIR_CONFIG_NAME = "config.json"
 _MODEL_DIR_TOKENIZER_NAME = "tokenizer.json"
+_MODEL_DIR_TOKENIZER_CONFIG_NAME = "tokenizer_config.json"
 
 
 def get_onnx_path_and_configs(
@@ -109,6 +110,11 @@ def get_onnx_path_and_configs(
         tokenizer_path = _get_file_parent(
             zoo_model.deployment.default.get_file(_MODEL_DIR_TOKENIZER_NAME).path
         )
+        tokenizer_config_path = zoo_model.deployment.default.get_file(
+            _MODEL_DIR_TOKENIZER_CONFIG_NAME
+        )
+        if tokenizer_config_path is not None:
+            tokenizer_config_path.path  # trigger download of tokenizer_config
     elif require_configs and (config_path is None or tokenizer_path is None):
         raise RuntimeError(
             f"Unable to find model and tokenizer config for model_path {model_path}. "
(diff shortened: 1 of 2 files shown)


src/deepsparse/transformers/helpers.py
```diff
@@ -109,6 +110,11 @@ def get_onnx_path_and_configs(
         tokenizer_path = _get_file_parent(
             zoo_model.deployment.default.get_file(_MODEL_DIR_TOKENIZER_NAME).path
         )
+        tokenizer_config_path = zoo_model.deployment.default.get_file(
+            _MODEL_DIR_TOKENIZER_CONFIG_NAME
+        )
+        if tokenizer_config_path is not None:
+            tokenizer_config_path.path  # trigger download of tokenizer_config
```
corey-nm, line 117:
```suggestion
            tokenizer_config_path.download()
```

corey-nm (changes requested):

src/deepsparse/transformers/helpers.py
bfineran: don't want to force download, going through `.path` route to use cache, can update comment if confusing

rahul-tuli: Calling path also downloads the needed artifacts right?

rahul-tuli (dismissed):

corey-nm:
I think this stuff is unit testable if you mock.patch the download function. You could see how many things & what were downloaded then. You could probably also patch `zoo_model.deployment.default.get_file` for this specific use case. once to return None and once to return something

bfineran:
throwing up a quick test

src/deepsparse/transformers/helpers.py
corey-nm: The comment is clear, it just reads weird because the property access is being treated as a method. It makes sense because we all know that path is actually a method, but for those who don't know I imagine its confusing

bfineran: right, that's why the comment was added, since it may be unclear otherwise that accessing `path` may trigger a download

rahul-tuli: Does `.download()` not use the `cache`? 

For the scope of this PR it seems okay, but maybe using `.path` is adding more confusion than making the code clearer. For example just looking at `Model(...).path` it is not very clear whether this will download the model(we know it will download the model ofc) or just `return` the local path where model is/will be stored.

rahul-tuli (approved):

corey-nm (approved):
👍

PR #580, opened 2022-08-16
Add torch to haystack auto-install dependencies

kylesayrs:
This PR fixes tests which are failing due to the Haystack auto-install. Because torch is never listed as a dependency in DeepSparse, I guessed that the policy was to leave torch installation up to the user. However torch is required on the hot path of Haystack, so it must be installed for tests involving haystack to work.


kylesayrs:
Deferring to this solution: https://github.com/neuralmagic/deepsparse/pull/581

PR #583, opened 2022-08-16
Reducing slowness in yolact pipeline

corey-nm:
This is caused because pydantic will actually check every element of a list of floats to make sure it is valid. Masks are so large that it actually takes a not insignifant time to type check this


rahul-tuli (dismissed):

bfineran (commented):
LGTM - @dbogunowicz is it safe to remove this `flatten` call now that we are returning a `List[Any]`?

PR #588, opened 2022-08-18
[Commercial Server] InferenceTiming + TimingBuilder

dbogunowicz:
# UI

`TimingBuilder` is an object that aggregates the durations (time deltas in seconds) of various components of the inference pipeline. Once all the desired time deltas are aggregated, the object may build the data structure (dict: [str, float]) that summarizes the collected information.
Example flow:
  ```python
  builder = TimingBuilder()
  builder.start("total_inference")
  builder.start("pre_process")
  do_pre_process()
  builder.stop("pre_process")
  builder.start("engine_forward")
  do_engine_forward()
  builder.stop("engine_forward")
  builder.start("post_process")
  do_post_process()
  builder.stop("post_process")
  builder.stop("total_inference")
  ...
  summary = builder.build()
  print(summary)
  >> {'pre_process: 0.2, `post-process`: 0.2, 'engine_forward': 0.6, 'total_inference': 1}
  ```
 # Testing
 PR is covered by its respective unittest


src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+                "This will be only possible once the build() method is called."
+            )
+        self._post_process_delta = value
+
+    def start(self):
+        if self.started:
+            raise ValueError("The TimingBuilder instance has been already started")
+        self.started = True
+
+    def build(self):
```
dbogunowicz, line 149:
```suggestion
    def build(self) -> InferenceTimingSchema:
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    def engine_forward_delta(self, value):
+        if self._engine_forward_delta is not None:
+            raise ValueError(
+                "Attempting to overwrite the active time delta readout. "
+                "This will be only possible once the build() method is called."
+            )
+        self._engine_forward_delta = value
+
+    @post_process_delta.setter
+    def post_process_delta(self, value):
```
dbogunowicz, line 136:
```suggestion
    def post_process_delta(self, value: float):
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    def pre_process_delta(self, value):
+        if self._pre_process_delta is not None:
+            raise ValueError(
+                "Attempting to overwrite the active time delta readout. "
+                "This will be only possible once the build() method is called."
+            )
+        self._pre_process_delta = value
+
+    @engine_forward_delta.setter
+    def engine_forward_delta(self, value):
```
dbogunowicz, line 127:
```suggestion
    def engine_forward_delta(self, value: float):
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+            raise ValueError(
+                "Attempting to collect completion time information, "
+                "but the placeholder for start time has not "
+                "been reset. Make sure that the timing of the current"
+                "process has been started"
+            )
+        self._t1 = value
+
+    @pre_process_delta.setter
+    def pre_process_delta(self, value):
```
dbogunowicz, line 118:
```suggestion
    def pre_process_delta(self, value: float):
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+            raise ValueError(
+                "Attempting to collect start time information, "
+                "but the placeholder for completion time has not "
+                "been reset. Make sure that the timing of the previous"
+                "process has been completed"
+            )
+        self._t0 = value
+
+    @t1.setter
+    def t1(self, value):
```
dbogunowicz, line 101:
```suggestion
    def t1(self, value : float):
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def engine_forward_delta(self):
+        return self._engine_forward_delta
+
+    @property
+    def post_process_delta(self):
+        return self._post_process_delta
+
+    @t0.setter
+    def t0(self, value):
```
dbogunowicz, line 84:
```suggestion
    def t0(self, value: float):
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def pre_process_delta(self):
+        return self._pre_process_delta
+
+    @property
+    def engine_forward_delta(self):
+        return self._engine_forward_delta
+
+    @property
+    def post_process_delta(self):
```
dbogunowicz, line 80:
```suggestion
    def post_process_delta(self) -> float:
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def t1(self):
+        return self._t1
+
+    @property
+    def pre_process_delta(self):
+        return self._pre_process_delta
+
+    @property
+    def engine_forward_delta(self):
```
dbogunowicz, line 76:
```suggestion
    def engine_forward_delta(self) -> float:
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+    @property
+    def t0(self):
+        return self._t0
+
+    @property
+    def t1(self):
+        return self._t1
+
+    @property
+    def pre_process_delta(self):
```
dbogunowicz, line 72:
```suggestion
    def pre_process_delta(self) -> float:
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+        self._pre_process_delta = None
+        self._engine_forward_delta = None
+        self._post_process_delta = None
+
+    @property
+    def t0(self):
+        return self._t0
+
+    @property
+    def t1(self):
```
dbogunowicz, line 68:
```suggestion
    def t1(self) -> float:
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,194 @@
...
+        self._t0 = None
+        # generate placeholder for completion time
+        self._t1 = None
+
+        self._pre_process_delta = None
+        self._engine_forward_delta = None
+        self._post_process_delta = None
+
+    @property
+    def t0(self):
```
dbogunowicz, line 64:
```suggestion
    def t0(self) -> float:
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,197 @@
...
+    summary = builder.build()
+    ```
+
+    The object can be extended to aggregate further deltas.
+    To extend, four things are required:
+        - adding {process_name}_delta property (including the setter)
+        - adding {process_name}_start method
+        - adding {process_name}_complete method
+        - adding the {process_name_delta} to the arguments collected
+          inside the build() method.
```
bfineran, line 51: thoughts on the following?

1. Removing the specific `_start()` and `_complete()` methods in favor of `builder.start_phase(name)`, `builder.end_phase(name)`.  
2. Keeping `start_times` and `end_times` dicts internally that are populated by `start/end_phase` functions
3. measuring deltas between the dict keys on build, and passing in as kwargs to the timing object

don't think that users need access to the deltas before build and this makes it a bit more flexible.  we can either fail if invalid phase names are added, or pipe them in as extras to the timing object

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+
+    def __init__(self):
+        self.started = False
+
+        # general placeholder for start time
+        self._t0 = None
+        # generate placeholder for completion time
+        self._t1 = None
```
bfineran, line 60: why do we need this? looks like it adds the potential for annoying bugs if the builder methods are called out of order for some reason.  see longer comment above about refactoring, shouldn't have a need for this after.

src/deepsparse/timing/timing_schema.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Stores the information about time deltas
+    (in seconds) of certain processes within
+    the inference pipeline
+    """
+
+    pre_process_delta: float
+    engine_forward_delta: float
+    post_process_delta: float
+    total_inference_delta: float
```
bfineran, line 31: nit: add as pydantic fields w/ description per org standards

bfineran (changes requested):

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,133 @@
...
+                f"Attempting to grab the stop time of the phase: {phase_name},"
+                f"but is start time missing"
+            )
+        if "stop" in self._start_stop_times[phase_name]:
+            raise ValueError(
+                f"Attempting to overwrite the stop time of the phase: {phase_name}"
+            )
+        self._start_stop_times[phase_name]["stop"] = time.time()
+
+    def initialize(self):
```
KSGulin, line 106: Is there a reason for requiring initialize? Are there plans to expand the logic for initialize in the future?

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        Collect the finish time of the phase
+
+        :param phase_name: The name of an event (phase), which duration
+            we are measuring
+        """
+        if phase_name not in self._start_stop_times:
+            raise ValueError(
+                f"Attempting to grab the stop time of the phase: {phase_name},"
+                f"but is start time missing"
```
KSGulin, line 98:
```suggestion
                f"but its start time is missing"
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,133 @@
...
+            raise ValueError(
+                "Attempting to collect time information, "
+                "but the TimingBuilder instance not initialized. "
+                "Call initialize() method first"
+            )
+        if phase_name in self._start_stop_times:
+            raise ValueError(
+                f"Attempting to overwrite the start time of the phase: {phase_name}"
+            )
+        self._start_stop_times[phase_name] = {"start": time.time()}
```
KSGulin, line 86: I'd recommend using `perf_counter()` over `time()` here, as it provides higher precision. Going back to our HPC class notes, the IEEE standard for floating points allocates a lot more floating point representation density to values near 1, so the large values you get with `time()` fall in a low resolution range. `perf_counter()` gets around that having a much later epoch (and thus returning smaller values)

corey-nm: +1

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,133 @@
...
+    ```
+    The object may time the duration of an arbitrary number
+    of events (phases), but the names of phases need to
+    be consistent with the `schema` argument passed to
+    the build() method
+    """
+
+    def __init__(self):
+        self.initialized = False
+        self._start_stop_times = {}
```
corey-nm, line 67: Can we get this split into `self.start_times = {}` and `self.stop_times = {}`? I think it'd make it slightly easier to validate that something has started, and slightly easier to read

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def initialize(self):
+        """
+        Prime the builder, so it is ready to collect time measurements
+        """
+        if self.initialized:
+            raise ValueError("The TimingBuilder instance has been already initialized")
+        self.initialized = True
+
+    def build(self, schema: BaseModel = InferenceTimingSchema) -> BaseModel:
```
corey-nm, line 114: What's the reasoning for using pydantic here? This is all an internal api, so assuming its tested, it should never create fields that shouldn't be there. 

An alternative to this approach is to have a `def query(self, phase) -> time` method. Then you wouldn't have to return anything and wouldn't need this extra step.

tests/deepsparse/timing/test_timing_builder.py
```diff
@@ -0,0 +1,87 @@
...
+import pytest
+from deepsparse.timing import TimingBuilder
+
+
+def _sleep(sleep_time):
+    time.sleep(sleep_time)
+
+
+@pytest.mark.parametrize("s1, s2, s3", [(0.1, 0.2, 0.3)])
+class TestTimingBuilder:
```
corey-nm, line 26: ❤️ 😁

src/deepsparse/timing/timing_builder.py
dbogunowicz: Good point, I do not really see a reason why it is required, will remove it.

src/deepsparse/timing/timing_builder.py
dbogunowicz: Not entirely sure what you mean here. My goal was to use TimerBuilder to be able to:
-  create time deltas for phases with different names
-  once the deltas area is created, put them into a schema to be returned for further consumption

I feel that using Pydantic schema here prevents us from e.g. having a typo in one of our `phase_name`s.
E.g. `pre_processing` could be erronously written as `pre_processing_time`.  Pydantic would catch that.

corey-nm: Ahh composition gotcha, that makes sense. Right cause these will be summed/averaged together from the different runs? Okay yeah this makes more sense then!

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,120 @@
...
+    pipeline.
+    Once all the desired time deltas are aggregated, the
+    object may build the data structure that summarizes the
+    collected information.
+
+    Example flow:
+
+    ```
+    builder = TimingBuilder()
+    builder.initialize()
```
dbogunowicz, line 39:
```suggestion
```

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,119 @@
...
+                f"Attempting to grab the stop time of the phase: {phase_name},"
+                f"but its start time is missing"
+            )
+        if phase_name in self._stop_times:
+            raise ValueError(
+                f"Attempting to overwrite the stop time of the phase: {phase_name}"
+            )
+        self._stop_times[phase_name] = time.perf_counter()
+
+    def build(self, schema: BaseModel = InferenceTimingSchema) -> BaseModel:
```
corey-nm, line 99: Can we just return InferenceTimingSchema here instead of passing in a schema?

corey-nm: Alternatively going back to dictionary comment - this class could just return the delta dictionary, and then whatever class is using the TimingBuilder could build it's schema using the dictionary. That way this TimingBuilder class doesn't have to know anything about schemas, but we are still capturing that type checking outside of this class

KSGulin (dismissed):
LGTM pending @corey-nm's comment

src/deepsparse/timing/timing_builder.py
dbogunowicz: Why don't you like the current state of the `build` method?

I could agree on returning the `InferenceTimingSchema` and not passing a `schema` as an arg.
We will be effectively limiting the flexibility of the object (you cannot name your phases however you like), but maybe for a good cause.
And then if we want to extend the usability of the TimingBuilder in the future, we can do it (and then perhaps return dicts as you proposed).

corey-nm: I think because this method is doing some validating on fields you used earlier, whereas the start/stop methods don't do that validation. So there is a little bit of inconsistency.

The phase names and therefore fields of the schema are controlled outside of this class, so in my head it makes sense for building the schema to be outside of this class as well.

Agree that returning InferenceTimingSchema from this function does seem too limiting with that in mind (since the phase names are not required to be valid for InferenceTimingSchema).

dbogunowicz: Ok, now I am leaning towards returning the dict and putting it into the schema outside of the class methods.

bfineran: agreed, I like this flow - it doesn't intrude on anything we'll need in the server and keeps the builder as something we can use generically for other timings

bfineran (dismissed):

bfineran (approved):
LGTM, please update PR description with goal, ux, test plan before landing

KSGulin (approved):

PR #589, opened 2022-08-18
[Commercial Server] InferenceTiming + pipeline integration

dbogunowicz:
This is the PR that integrates the `TimerBuilder` introduced in https://github.com/neuralmagic/deepsparse/pull/588 into the deepsparse base `Pipeline` class.

## Testing
Covered by appropriate unittests.

Note: This PR depends on the `unify-run-split-join` branch, and will be landed once the parent branch is merged to main.


tests/deepsparse/pipelines/test_inference_timing.py
```diff
@@ -0,0 +1,78 @@
...
+            "zoo:cv/segmentation/yolact-darknet53/pytorch/dbolya/coco/pruned82_quant-none",  # noqa E501
+        ),
+        (
+            "yolo",
+            "zoo:cv/detection/yolov5-s/pytorch/ultralytics/coco/pruned_quant-aggressive_94",  # noqa E501
+        ),
+    ],
+)
+class TestInferenceTiming:
+    @pytest.fixture
```
corey-nm, line 45: Add scope=class, otherwise this setup call gets called for each test.

src/deepsparse/pipeline.py
```diff
@@ -638,6 +661,9 @@ def _run_with_static_batch(self, pipeline_inputs: BaseModel) -> BaseModel:
                 f"Outputs of {self.__class__} must be instances of "
                 f"{self.output_schema} found output of type {type(pipeline_outputs)}"
             )
+        self.timer.post_process_complete()
+
+        self.inference_timing = self.timer.build()
```
corey-nm, line 666: How does this stuff interact with dynamic batch which uses async?

Perhaps _run_with_static batch should create & return a timer inside of it, and _run_with_dynamic_batch can sum them all together or take max or something else

src/deepsparse/pipeline.py
```diff
@@ -175,6 +176,8 @@ def __init__(
 
         self.onnx_file_path = self.setup_onnx_file_path()
         self.engine = self._initialize_engine()
+        self._inference_timing = None
+        self.timer = TimingBuilder()
```
corey-nm, line 180: What about just having a `self.call_times: List[TimingBuilder] = []`, and then `__call__` can append/extend this list?

Then in server we can just dip into this call_times list.

Having a inference timing schema feels overkill unless i'm missing some use case for it!

bfineran, line 180: timing builder's scope should just be within `call` not a property of the class (ie if there are multiple calls to call in async how might this affect the builder?)

src/deepsparse/pipeline.py
```diff
@@ -175,6 +176,8 @@ def __init__(
 
         self.onnx_file_path = self.setup_onnx_file_path()
         self.engine = self._initialize_engine()
+        self._inference_timing = None
```
bfineran, line 179: inference timing should not be a class property since it relates to a single inference - should be returned with the specific inference call

src/deepsparse/pipeline.py
mgoin: I think the way to do this is to gather the time of the start of the first dynamic inference and and the end of the last dynamic inference. This is how `deepsparse.benchmark` tracks time. Computing this from just deltas seems difficult.

src/deepsparse/pipeline.py
```diff
@@ -617,17 +619,29 @@ def _run_with_dynamic_batch(
         ]
         # wait for all inferences to complete before joining outputs
         concurrent.futures.wait(futures)
-        outputs = [future.result() for future in futures]
-        return self.output_schema.join(outputs)
+        collected_results = [future.result() for future in futures]
+        outputs, inference_timing_batch = zip(*collected_results)
+        return self.output_schema.join(outputs), InferenceTimingSchema.aggregate(
```
dbogunowicz, line 624: Wondering, whether we should include the inference timing object inside the output schema. i.e create field `InferenceTiming` inside every `OutputSchema`

src/deepsparse/pipeline.py
dbogunowicz: @mgoin I would lean towards @corey-nm suggestion. The recent commit collects all the separate inference timing readouts in an `Iterable[InferenceTimingSchema]` and sums over all the labels. I like it because :
1. we are not losing any information (e.g. we can still see roughly how much time the input spent in pre-processing step alone)
2. `total_time` still is equal to the sum of the `pre_processing`, `post_processing` and `engine_forward`

src/deepsparse/pipeline.py
```diff
@@ -176,29 +177,34 @@ def __init__(
...
         if "engine_inputs" in kwargs:
             raise ValueError(
                 "invalid kwarg engine_inputs. engine inputs determined "
                 f"by {self.__class__.__qualname__}.parse_inputs"
             )
+        timer = TimingBuilder()
+        timer.initialize()
 
         # parse inputs into input_schema
+        timer.start("total_inference")
```
corey-nm, line 190: Can we move all these strings to global constants so it'll be easy to change if necessary?

```
TOTAL_INFERENCE = "total_inference"
...
timer.start(TOTAL_INFERENCE)
```

Wanted to do that with all the sample_inputs/sample-inputs too 😁

src/deepsparse/pipeline.py
```diff
@@ -176,29 +177,34 @@ def __init__(
 
         self._batch_size = self._batch_size or 1
 
-    def __call__(self, *args, **kwargs) -> BaseModel:
+    def __call__(self, monitoring: bool = False, *args, **kwargs) -> BaseModel:
```
corey-nm, line 180: a kwargs should go after `*args` in this case
```suggestion
    def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
```

src/deepsparse/pipeline.py
```diff
@@ -220,8 +228,30 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
+        if not monitoring:
+            return pipeline_outputs
+
+        else:
+            data = None
+            timing = timer.build()
+            return pipeline_outputs, timing, data
 
-        return pipeline_outputs
+    def run_with_monitoring(self, *args, **kwargs) -> Tuple[BaseModel, BaseModel, Any]:
```
corey-nm, line 242: Is there a case when we wouldn't want to be monitoring? wondering if we can just always monitor and delete this extra function

src/deepsparse/pipeline.py
```diff
@@ -220,8 +228,30 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
+        timer.stop("post_process")
+        timer.stop("total_inference")
+
+        if not monitoring:
+            return pipeline_outputs
+
+        else:
+            data = None
+            timing = timer.build()
+            return pipeline_outputs, timing, data
```
corey-nm, line 240: Thoughts on storing the timing as `self._last_timing = timer.build`, and then maybe adding a method for `def get_times_for_last_run(self) -> Optional[InferenceTimes]`?

Then we wouldn't need the additional flag in this method, but still have a clear way to access it.

corey-nm (changes requested):

src/deepsparse/pipeline.py
corey-nm: Agree with you there - then I vote for the idea in my next comment down about additional method for retrieiving timings instead of the monitoring flag 😁

src/deepsparse/pipeline.py
corey-nm: Oooh yeah that's a great call out about reliably returning the last timings and the data stuff.

Will the PipelineLogger class be outside of Pipeline? Or will all this timing data be passed directly to PipelineLogger from inside Pipeline (maybe PipelineLogger is passing into constructor of Pipeline)?

src/deepsparse/pipeline.py
```diff
@@ -29,6 +29,13 @@
 from deepsparse.benchmark import ORTEngine
 from deepsparse.cpu import cpu_details
 from deepsparse.tasks import SupportedTasks
+from deepsparse.timing import (
+    ENGINE_FORWARD,
+    POST_PROCESS,
+    PRE_PROCESS,
+    TOTAL_INFERENCE,
```
bfineran, line 36: let's nest these under a `PipelineTimings` enum or something like that

src/deepsparse/pipeline.py
bfineran: yup, just want to call out here that we should never keep state in a pipeline that relates to a single inference run (esp the latest) as this will lead to race conditions in async runs

tests/deepsparse/pipelines/test_pipeline.py
```diff
@@ -21,9 +21,10 @@
 
 import pytest
 from deepsparse import Pipeline
+from deepsparse.schemas import DeepSparseSchema
 
 
-class FakeSchema(BaseModel):
+class FakeSchema(DeepSparseSchema):
```
bfineran, line 27: let's get this refactored to use the `mock_pipeline`

src/deepsparse/timing/timing_builder.py
```diff
@@ -0,0 +1,133 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: need to make sure to properly merge/rebase with #588

bfineran (changes requested):

src/deepsparse/pipeline.py
```diff
@@ -220,8 +227,30 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
+        if not monitoring:
+            return pipeline_outputs
+
+        else:
+            data = None
+            timer_summary = timer.build()
+            return pipeline_outputs, InferenceTimingSchema(**timer_summary), data
 
-        return pipeline_outputs
+    def run_with_monitoring(self, *args, **kwargs) -> Tuple[BaseModel, BaseModel, Any]:
```
dbogunowicz, line 241:
```suggestion
    def run_with_monitoring(self, *args, **kwargs) -> Tuple[BaseModel, InferenceTimingSchema, Any]:
```

src/deepsparse/timing/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# flake8: noqa
+
+from .helpers import *
```
dbogunowicz, line 17:
```suggestion
```

src/deepsparse/pipeline.py
```diff
@@ -220,8 +227,30 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
+            timer_summary = timer.build()
+            return pipeline_outputs, InferenceTimingSchema(**timer_summary), data
 
-        return pipeline_outputs
+    def run_with_monitoring(self, *args, **kwargs) -> Tuple[BaseModel, InferenceTimingSchema, Any]:
+        """
+        Runs the inference with monitoring.
+
+        This means setting the `monitoring` flag to True for __call__()
+        and as result receiving additional information from the forward pass:
```
dbogunowicz, line 246: add additional return

bfineran (changes requested):
needs rebase + update PR description

src/deepsparse/pipeline.py
```diff
@@ -176,29 +177,33 @@ def __init__(
...
+    def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
         if "engine_inputs" in kwargs:
             raise ValueError(
                 "invalid kwarg engine_inputs. engine inputs determined "
                 f"by {self.__class__.__qualname__}.parse_inputs"
             )
+        timer = TimingBuilder()
 
         # parse inputs into input_schema
+        timer.start(InferencePhases.TOTAL_INFERENCE.value)
```
bfineran, line 189: let's add support in timer.start/end to unwrap the enums so a user calling in won't have to. (ie `if isinstance(phase, Enum)...`

src/deepsparse/pipeline.py
```diff
@@ -220,8 +227,36 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
                 f"{self.output_schema} found output of type {type(pipeline_outputs)}"
             )
+        timer.stop(InferencePhases.POST_PROCESS.value)
+        timer.stop(InferencePhases.TOTAL_INFERENCE.value)
+
+        if not monitoring:
+            return pipeline_outputs
+
+        else:
+            data = None
```
bfineran, line 237: let's return the input schema as well as engine inputs here. should not be pushing empty functionality to main

bfineran (commented):
two quick comments then LGTM

src/deepsparse/pipeline.py
corey-nm: Alternatively, I think we don't even need the `.value`?

```python
>>> from pydantic import BaseModel
>>> from enum import Enum
>>> class A(str, Enum):
...     a = "a"
...     b = "b"
... 
>>> class Schema(BaseModel):
...     a: int
...     b: int
... 
>>> {A.a: 1, A.b: 2}
{<A.a: 'a'>: 1, <A.b: 'b'>: 2}
>>> Schema(**{A.a: 1, A.b: 2})
Schema(a=1, b=2)
```

KSGulin (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -176,29 +177,33 @@ def __init__(
 
         self._batch_size = self._batch_size or 1
 
-    def __call__(self, *args, **kwargs) -> BaseModel:
+    def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
```
corey-nm, line 180: Should monitoring be a pipeline constructor arg? Will there be pipelines that sometimes monitor their forward calls and sometimes not?

src/deepsparse/pipeline.py
```diff
@@ -220,8 +227,39 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
             )
+        timer.stop(InferencePhases.POST_PROCESS)
+        timer.stop(InferencePhases.TOTAL_INFERENCE)
+
+        if not monitoring:
+            return pipeline_outputs
+
+        else:
+            inference_timing = InferenceTimingSchema(**timer.build())
+            return pipeline_outputs, pipeline_inputs, engine_inputs, inference_timing
```
corey-nm, line 238: Do we know how exactly will pipeline_inputs/engine_inputs be used? I know monitoring data distribution was discussed previously, but what does that actually look like?

src/deepsparse/pipeline.py
```diff
@@ -176,29 +177,33 @@ def __init__(
...
         self._batch_size = self._batch_size or 1
 
-    def __call__(self, *args, **kwargs) -> BaseModel:
+    def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
         if "engine_inputs" in kwargs:
             raise ValueError(
                 "invalid kwarg engine_inputs. engine inputs determined "
                 f"by {self.__class__.__qualname__}.parse_inputs"
             )
+        timer = TimingBuilder()
```
corey-nm, line 186: Have you tested this in the server with multiple requests happening concurrently? This is local to this function, so I _think_ we shouldn't have interference from other requests? I'm not even really sure how to tell if there is interference

src/deepsparse/pipeline.py
dbogunowicz: This remains to be defined.

src/deepsparse/pipeline.py
bfineran: this was the intended effect of keeping the builder local to this function, we can look more into this after we integrate with the server

src/deepsparse/pipeline.py
bfineran: I'm okay with keeping this at the call level for now, nothing inherently changes about the pipeline object based on this variable and allows us to not have to reinitialize if for some reason we want to run with and without for the same pipeline

bfineran (dismissed):

tests/deepsparse/pipelines/test_pipeline.py
```diff
@@ -141,3 +142,21 @@ def sleep_for_10ms(xs):
         end = time.time_ns()
         dur_ms = (end - start) * 1e-6
         assert abs(dur_ms - 30) < 10
+
+
+@mock_engine(rng_seed=0)
+def test_run_with_monitoring(engine_mock):
```
corey-nm, line 148: nit: probably would want to make each of the pre/post/forward functions take a specific amount of time known by the unit test to truly verify that our timings are correct.

src/deepsparse/pipeline.py
```diff
@@ -220,8 +227,39 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
+                elapsed during the inference steps: pre-processing,
+                engine-forward, post-processing, as well as
+                the total elapsed time
+        """
+        (
+            pipeline_outputs,
+            pipeline_inputs,
+            engine_inputs,
+            inference_timing,
+        ) = self.__call__(monitoring=True, *args, **kwargs)
```
corey-nm, line 261:
```suggestion
        ) = self.__call__(*args, monitoring=True, **kwargs)
```

src/deepsparse/pipeline.py
corey-nm: Yeah but it makes the caller much easier if its an attribute. Probably don't even need run_with_monitoring in that case. And then in the server we can just pass that into constructor instead of having to specify monitoring=True or call run_with_monitoring.

corey-nm (changes requested):

src/deepsparse/pipeline.py
```diff
@@ -220,8 +227,39 @@ def __call__(self, *args, **kwargs) -> BaseModel:
...
+        return pipeline_outputs, pipeline_inputs, engine_inputs, inference_timing
```
corey-nm, line 262:
```suggestion
        return self(*args, monitoring=True, **kwargs)
```

corey-nm (approved):

bfineran (approved):

PR #590, opened 2022-08-18
Fixed condition where a single numpy array was passed

rahul-tuli:
This PR fixes a bug; and makes `YOLO` pipeline more robust, essentially when a `numpy` array was passed via annotate script it is send in a channels first format, when we try to resize this image using opencv, an error was thrown; the current PR always makes the channels last before resizing hence fixing the bug. Tested with the following:

```bash
deepsparse.object_detection.annotate --source test_images/basilica.jpg
```

output:
```bash
❯ deepsparse.object_detection.annotate --source test_images/basilica.jpg
2022-08-18 11:27:05 deepsparse.utils.annotate INFO     Results will be saved to annotation-results/deepsparse-annotations
DeepSparse Engine, Copyright 2021-present / Neuralmagic, Inc. version: 1.1.0.20220816 COMMUNITY EDITION (c1e95a75) (release) (optimized) (system=avx512, binary=avx512)
2022-08-18 11:27:11 deepsparse.yolo.annotate INFO     Results saved to annotation-results/deepsparse-annotations
INFO:deepsparse.yolo.annotate:Results saved to annotation-results/deepsparse-annotations
```

### DIFF
--- a/src/deepsparse/yolo/pipelines.py
+++ b/src/deepsparse/yolo/pipelines.py
@@ -178,7 +178,7 @@ class YOLOPipeline(Pipeline):
         # Noting that if numpy arrays are passed in, we assume they are
         # already the correct shape
 
-        if isinstance(inputs.images, str):
+        if isinstance(inputs.images, (str, numpy.ndarray)):
             inputs.images = [inputs.images]
 
         image_batch = []
@@ -191,6 +191,7 @@ class YOLOPipeline(Pipeline):
             if isinstance(image, str):
                 image = cv2.imread(image)
 
+            image = self._make_channels_last(image)
             image = cv2.resize(image, dsize=tuple(reversed(self.image_size)))
             image = self._make_channels_first(image)
             image_batch.append(image)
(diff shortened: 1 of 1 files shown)


corey-nm (approved):

src/deepsparse/yolo/pipelines.py
```diff
@@ -285,6 +286,22 @@ def _make_channels_first(self, image: numpy.ndarray) -> numpy.ndarray:
 
         return image
 
+    def _make_channels_last(self, image: numpy.ndarray) -> numpy.ndarray:
```
dbogunowicz, line 289: I have the exact same function in my yolact utils. It hurts to see how we duplicate the CV code. Just to reinforce your ask from yesterday @rahul-tuli (about refactoring the vision pipelines).

dbogunowicz (approved):

src/deepsparse/yolo/pipelines.py
```diff
@@ -285,6 +286,22 @@ def _make_channels_first(self, image: numpy.ndarray) -> numpy.ndarray:
...
+        # return a numpy array with channels first
+        is_single_image = image.ndim == 3
+        is_batch = image.ndim == 4
```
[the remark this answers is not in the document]
rahul-tuli: I feel the variable names are more intuitive than comments, leaving this as is for now.

src/deepsparse/yolo/pipelines.py
rahul-tuli: @dbogunowicz I agree but, I did not want to import `yolact` utils inside `yolo` pipelines, in my opinion these helper functions need to be refactored out into a parent package and they need to be tested for all three cv pipelines which is non-trivial work;

My ask yesterday was to refactor CV Input Schemas which is comparatively very trivial, all CV input schemas had the exact same function with same lines of code, that was moved to a parent class.

I propose to leave this as is here, and take up refactoring as a separate bigger task. Thoughts?

PR #595, opened 2022-08-19
[Commercial Server] (Base) PipelineLogger + PrometheusPipelineLogger

dbogunowicz:
This PR introduces the pipeline loggers (located in `src/deepsparse/loggers`).
In total, there are three new objects created:
- `BaseLogger` - an abstract class meant to define interfaces for the loggers that support various monitoring services APIs.
- `PrometheusLogger` - a logger that uses the Prometheus client to monitor the inference pipeline
- `ManagerLogger` - a logger manager that oversees all the pipeline loggers

Testing is covered by unit tests

Example use of Prometheus logger.

```python
pipeline_yolo = ...
logger = PrometheusLogger( port = 8000)
while True:
    input_image = ... # from external source
    out, input_schema, engine_input, inference_timing = pipeline_yolo.run_with_monitoring(images=input_image)
    logger.log_latency(pipeline_yolo.name, inference_timing)
    logger.log_data(pipeline_yolo.name, ...) # interface for logging data to be defined
```

Example use of Manager logger:

```python
loggers = [PrometheusLogger(), CloudWatchLogger(), ...]
logger_manager = ManagerLogger(loggers)
pipeline_yolo = ...
while True:
    input_image = ... # from external source
    out, input_schema, engine_input, inference_timing = pipeline_yolo.run_with_monitoring(images=input_image)
    logger.log_latency(pipeline_yolo.name, inference_timing)
    logger.log_data(pipeline_yolo.name, ...) # interface for logging data to be defined
```


src/deepsparse/pipeline_loggers/prometheus_pipeline_logger.py
```diff
@@ -0,0 +1,121 @@
...
+            # will be run on the first digestion of
+            # inference_timing data
+            self._setup_metrics(inference_timing)
+
+        self._assert_incoming_data_consistency(inference_timing)
+
+        for metric_name, value_to_log in dict(inference_timing).items():
+            self._log_latency(metric_name=metric_name, value_to_log=value_to_log)
+
+        _LOGGER.info("Prometheus client: logged latency metrics")
```
bfineran, line 71: definitely don't want to log to console every time we log to prometheus, especially with no extra info

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,50 @@
...
+__all__ = ["PipelineLogger"]
+
+
+class PipelineLogger(ABC):
+    """
+    Generic PipelineLogger abstract class meant to define interfaces
+    for the loggers that support various monitoring services APIs.
+    """
+
+    def __init__(self):
```
bfineran, line 28: need infrastructure to get a different logger instance for each pipeline

bfineran (changes requested):
great first pass. seeing as how the implementations need to do some global setup, I think it doesn't make sense for us to initialize new loggers for each pipeline anymore. Instead, let's make `pipeline_name` a required argument of each of the `log_` methods (ie `log_latency("pipeline_1", ...)`)

this will also make the integration at the server level easier as we don't need to reconcile creation and management of logger objects for every combination of logger impl and pipeline entrypoint

let's aim for to support a UX for something like this:

```python
# create logger and pipelines
logger = LoggerManager(["prometheus", "elastic_search"])
pipelines = [Pipeline.from_config(config) for config in pipeline_configs]

def create_routes(pipelines, logger):
    # define route for each pipeline, pipelines share the global logger
    for pipeline in pipelines:
        @get(ROUTE_NAME)
         def pipeline_route(inputs):
                # run pipeline
                results, timings, data = pipeline.run_with_monitoring(inputs)  # pathway name here tbd
                # logger manager handles logging to each integration by pipeline name
                logger.log_timings(pipeline.name, timings)
                logger.log_data(pipeline.name, data)
                return results
```

src/deepsparse/pipeline_loggers/prometheus_pipeline_logger.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        While self._counter collects the number of forward passes per
+        pipeline inference, self.counter returns a sum of all the calls
+
+        :return: The total amount of times the logger has been called by
+            any pipeline
+        """
+        return sum(self._counter.values())
+
+    def _setup_client(self):
```
dbogunowicz, line 86: private method, move down

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,65 @@
...
+    for the loggers that support various monitoring services APIs.
+
+    :param identifier: The name of the monitoring service that the
+        PipelineLogger uses to log the inference data
+    """
+
+    def __init__(self, identifier: str):
+        self._identifier = identifier
+
+    @property
```
corey-nm, line 34: Related to my other comment.
Let's make this an abstract property for subclasses to implement, then it doesn't have to given to this class' constructor. And then the prometheus one can just define it as a property instead of having a global variable for it.

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,65 @@
...
+    def identifier(self) -> str:
+        return self._identifier
+
+    def __str__(self):
```
corey-nm, line 38: Let's move this into the subclasses, esp since this is an abstract class, no one will have an instance of just this

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,65 @@
...
+        Logs the inference latency information to the appropriate monitoring service
+        :param pipeline_name: The name of the inference pipeline from which the
+            logger consumes the inference information to be monitored
+        :param inference_timing: pydantic model that holds the information about
+            the inference latency of a forward pass
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def log_data(self, pipeline_name: str, inputs: Any, outputs: Any):
```
corey-nm, line 55: What's the expected order of calls for log_data and log_latency? Are they both required to be called or is log_data optional?

src/deepsparse/pipeline_loggers/manager.py
```diff
@@ -0,0 +1,132 @@
...
+    :param supported_loggers: A set of supported logger identifiers; listed in
+        `SUPPORTED_LOGGERS` variable
+    """
+
+    def __init__(
+        self,
+        logger_identifiers: Union[str, List[str]],
+        supported_loggers: Set[str] = SUPPORTED_LOGGERS,
+    ):
+        self._supported_loggers = supported_loggers
```
corey-nm, line 71: Are we expecting people to pass in their own set of supported loggers? I think this should not be a parameter and should just be a instance variable.

Additionally if you make the identifier a abstract class property you could do:
```python
loggers = [PrometheusLogger]
self._supported_loggers: Dict[str, Type[PipelineLogger]] = {logger.identifier: logger}
```

src/deepsparse/pipeline_loggers/manager.py
```diff
@@ -0,0 +1,132 @@
...
+        self._loggers = self._setup_loggers(logger_identifiers)
```
corey-nm, line 72: In tandem with the above comment you could remove _setup_loggers for this:
```python
# check loggers
for identifier in logger_identifiers:
    if identifier not in self._supported_loggers:
                raise ValueError(
                    "Attempting to create a pipeline logger with an "
                    f"unknown identifier: {identifier}. Supported "
                    f"identifiers are: {self._supported_loggers}"
                )

self._loggers = list(map(self._supported_loggers.get, logger_identifiers))
```

src/deepsparse/pipeline_loggers/manager.py
```diff
@@ -0,0 +1,132 @@
...
+    :param logger_identifiers: The identifier(s) of the logger types that
+        will be created within the scope of the manager. This can be either a
+        single identifier, or a list of identifiers (for multiple loggers).
+    :param supported_loggers: A set of supported logger identifiers; listed in
+        `SUPPORTED_LOGGERS` variable
+    """
+
+    def __init__(
+        self,
+        logger_identifiers: Union[str, List[str]],
```
corey-nm, line 68: I'm going to recommend a larger change that is to just pass in the list of PipelineLoggers directly here. That will do two things:
1. You can probably just remove the identifier all together in that class (which would remove a big portion of this)
2. Make it more clear that LoggerManager is a "meta" PipelineLogger that is built up of a list of sub pipeline loggers.

src/deepsparse/pipeline_loggers/manager.py
```diff
@@ -0,0 +1,132 @@
...
+from deepsparse.pipeline_loggers.prometheus_pipeline_logger import PrometheusLogger
+from deepsparse.timing.timing_schema import InferenceTimingSchema
+
+
+__all__ = ["LoggerManager"]
+
+SUPPORTED_LOGGERS = {PROMETHEUS_IDENTIFIER}
+
+
+class LoggerManager:
```
corey-nm, line 34: Suggestions:
1. Subclass from PipelineLogger
2. Rename to something like `MultiLogger` or `SequentialLogger` or something that captures that this is a pipeline logger built up of a list of other pipeline loggers

corey-nm (changes requested):
I think this needs a bit more work structurally. Seems like it will be straightforward to use prometheus though!

src/deepsparse/pipeline_loggers/manager.py
dbogunowicz: I like the idea, but I wonder how this would work, given that we use human-readable (writable) config to create the loggers. What I am saying is, that it would be natural to pass identifiers (string), directly parsed from the config, rather then logger classes, to create the manager.

src/deepsparse/pipeline_loggers/manager.py
corey-nm: Mainly to make it clear to readers & others that that is what it is. I see it the same as if PrometheusLogger didn't subclass PipelineLogger

src/deepsparse/pipeline_loggers/manager.py
corey-nm: I think in that case the server should do the mapping from identifier or name to the class and construct the classes. This keeps the human readable stuff isolated into server, so it doesn't infect this code.

As far as how the server does the mapping it could either be strings like this, or we could even have users use the class name and then do a `getattr(deepsparse.pipeline_loggers, name)`, where name could be `PrometheusLogger`.

src/deepsparse/pipeline_loggers/manager.py
dbogunowicz: In general, your whole set of bits of advice helps tremendously. However, the code above would not work. Properties are "properties" (duh!) of instances and not classes. There is a hack to make them global though, will try that since I really like the simplifications you are introducing

src/deepsparse/pipeline_loggers/manager.py
dbogunowicz: Ok, this makes sense.

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
dbogunowicz: Keeping it one level above, so that it will be harder to run into circular dependency import issue when creating new loggers

src/deepsparse/pipeline_loggers/multi_logger.py
```diff
@@ -0,0 +1,124 @@
...
+LOGGERS = [PrometheusLogger]
+SUPPORTED_LOGGERS = {logger.identifier: logger for logger in LOGGERS}
+
+
+class MultiLogger(PipelineLogger):
+    """
+    Object that contains multiple loggers for
+    the given inference pipeline.
+
+    The envisioned lifecycle of a logger
```
dbogunowicz, line 37:
```suggestion
    The envisioned lifecycle of a multilogger
```

src/deepsparse/pipeline_loggers/multi_logger.py
```diff
@@ -0,0 +1,124 @@
...
+    # log the data for the particular inference pipeline
+
+    timings, data = ...
+    logger.log_timings(pipeline_name, timings)
+    logger.log_data(pipeline_name, data)
+
+    ```
+
+    :param logger_identifiers: The identifier(s) of the logger types that
+        will be created within the scope of the manager. This can be either a
```
dbogunowicz, line 58:
```suggestion
        will be created within the scope of the logger. This can be either a
```

tests/deepsparse/pipeline_loggers/test_multilogger.py
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from deepsparse.pipeline_logger import PipelineLogger
+from deepsparse.pipeline_loggers import MultiLogger
+
+
+def test_logger_manager(identifier="prometheus"):
```
dbogunowicz, line 18: There are some naming problems that I will fix after the PR is approved.

src/deepsparse/pipeline_loggers/multi_logger.py
```diff
@@ -0,0 +1,124 @@
...
+SUPPORTED_LOGGERS = {logger.identifier: logger for logger in LOGGERS}
+
+
+class MultiLogger(PipelineLogger):
+    """
+    Object that contains multiple loggers for
+    the given inference pipeline.
+
+    The envisioned lifecycle of a logger
+    manager:
```
dbogunowicz, line 38: There are some naming problems that I will fix after the PR is approved.

src/deepsparse/pipeline_loggers/multi_logger.py
```diff
@@ -0,0 +1,124 @@
...
+        :param inputs: the data received and consumed by the inference
+            pipeline
+        :param outputs: the data returned by the inference pipeline
+        """
+        for name, logger in self.loggers.items():
+            logger.log_data(pipeline_name, inputs, outputs)
+
+    def _setup(
+        self, logger_identifiers: Union[str, List[str]]
+    ) -> Dict[str, PipelineLogger]:
```
dbogunowicz, line 105:
```suggestion
    ) -> Dict[str, Type[PipelineLogger]]:
```

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,70 @@
...
+
+class PipelineLogger(ABC):
+    """
+    Generic PipelineLogger abstract class meant to define interfaces
+    for the loggers that support various monitoring services APIs.
+    """
+
+    @classproperty
+    @abstractmethod
+    def identifier(self) -> str:
```
corey-nm, line 38: Maybe we can just use a method instead of property?
```suggestion
    @staticmethod
    @abstractmethod
    def identifier(self) -> str:
```

src/deepsparse/pipeline_loggers/multi_logger.py
```diff
@@ -0,0 +1,124 @@
...
+
+    :param logger_identifiers: The identifier(s) of the logger types that
+        will be created within the scope of the manager. This can be either a
+        single identifier, or a list of identifiers (for multiple loggers).
+    """
+
+    def __init__(
+        self,
+        logger_identifiers: Union[str, List[str]],
+    ):
```
corey-nm, line 65: We could just pass in the list of instantiated logger objects directly into this class `loggers: List[PipelineLogger]`, and move the creation/parsing logic (`self._setup`, `self._start`) outside of this class to somewhere else (the server I guess?)

Then this class would be a really tightly scoped util class to easily wrap multiple loggers. It'd be easier to test to since you won't have to go through strings and could just pass in the loggers you want to use.

dbogunowicz: I like this idea a lot!

src/deepsparse/pipeline_logger.py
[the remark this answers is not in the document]
bfineran: I don't think we'd run into dependency issues as long as we use the full module path.

let's also rename the `pipeline_loggers` directory to `logging` or `loggers`

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+__all__ = ["PipelineLogger", "classproperty"]
+
+
+class classproperty(object):
```
bfineran, line 22: nit: docstring

src/deepsparse/pipeline_logger.py
```diff
@@ -0,0 +1,70 @@
...
+    def log_latency(
+        self, pipeline_name: str, inference_timing: "InferenceTimingSchema"  # noqa F821
+    ):
+        """
+        Logs the inference latency information to the appropriate monitoring service
+        :param pipeline_name: The name of the inference pipeline from which the
+            logger consumes the inference information to be monitored
+        :param inference_timing: pydantic model that holds the information about
+            the inference latency of a forward pass
+        """
```
bfineran, line 57:
```suggestion
        """
        raise NotImplementedError()
```

src/deepsparse/pipeline_loggers/multi_logger.py
bfineran: sounds good, not a fan of `MultiLogger` name

bfineran, line 38: nit: pretty short line breaks here

src/deepsparse/pipeline_loggers/multi_logger.py
```diff
@@ -0,0 +1,124 @@
...
+            logger consumes the data information to be monitored
+        :param inputs: the data received and consumed by the inference
+            pipeline
+        :param outputs: the data returned by the inference pipeline
+        """
+        for name, logger in self.loggers.items():
+            logger.log_data(pipeline_name, inputs, outputs)
+
+    def _setup(
+        self, logger_identifiers: Union[str, List[str]]
```
bfineran, line 104: need to think if we can do something more extensible here than string names to the desired loggers - ie if there's stuff we need to configure with an integration like prometheus's port, save directory, etc - we'd like to be able to control that at the server (and even user) level. 

so just taking a list of constructed logger objects here and ignoring identifiers might be the way to go, leaving the onus of construction to the server

src/deepsparse/pipeline_loggers/prometheus_pipeline_logger.py
```diff
@@ -0,0 +1,189 @@
...
+        Sets up the set of metrics that the logger will expect
+        to receive continuously
+
+        :param inference_timing: Pydantic model that contains information
+            about time deltas processes within the inference pipeline
+        """
+        # set multiple Histograms to track latencies
+        # per Prometheus docs:
+        # Histograms track the size and number of events in buckets
+        self.metrics[pipeline_name] = {}
```
bfineran, line 151: can use a default dict (of dict) here

bfineran (changes requested):

src/deepsparse/pipeline_logger.py
dbogunowicz: Corey pointed out: "For abstract methods, you can actually just have the docstring and it'll auto do the not implemented thing."
Tbh, I could not find the confirmation of this statement on the internet, but it's working fine, so I guess this may be true.

src/deepsparse/pipeline_loggers/multi_logger.py
dbogunowicz: good point!

src/deepsparse/pipeline_logger.py
bfineran: would prefer to make this explicit

bfineran (commented):
lgtm pending comment and PR description

bfineran (approved):
LGTM, please add examples of prometheus usage/access in the description

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -0,0 +1,188 @@
...
+    @property
+    def counter(self) -> int:
+        """
+        While self._counter collects the number of forward passes per
+        pipeline inference, self.counter returns a sum of all the calls
+
+        :return: The total amount of times the logger has been called by
+            any pipeline
+        """
+        return sum(self._counter.values())
```
corey-nm, line 87: Is `self._counter` only used here to get the number of times logger has been called? Can self._counter just be a single value instead of a dictionary?

corey-nm (approved):
LGTM 🎉

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: self._counter counts how many inferences were done per pipeline
self.counter sums them up and returns the total count of interences of the logger

corey-nm: It looks like self._counter[pipeline_name] is incremented once per log_latency call - can you just have an integer attribute that you increment instead of the dictionary? Then you could just use `self._counter` directly instead of doing the sum of values

PR #596, opened 2022-08-19
Bfloat16

ProExpertProg:
Added support for bfloat16, as we can now detect it on the architecture.

### DIFF
--- a/src/deepsparse/engine.py
+++ b/src/deepsparse/engine.py
@@ -310,7 +310,7 @@ class Engine(object):
         """
         :return: True if bfloat16 support was detected on the cpu, False otherwise.
         """
-        return self._cpu_vnni
+        return self._cpu_bf16
 
     def run(
         self,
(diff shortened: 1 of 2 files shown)


src/deepsparse/engine.py
```diff
@@ -184,6 +185,7 @@ def __init__(
         self._input_shapes = input_shapes
         self._cpu_avx_type = AVX_TYPE
         self._cpu_vnni = VNNI
+        self._cpu_bf16 = BF16
```
mgoin, line 188: could we keep the `ARCH` object as the member and query that for the property functions?

src/deepsparse/engine.py
```diff
@@ -303,6 +305,13 @@ def cpu_vnni(self) -> bool:
         """
         return self._cpu_vnni
 
+    @property
+    def cpu_bf16(self) -> bool:
+        """
+        :return: True if bfloat16 support was detected on the cpu, False otherwise.
+        """
+        return self._cpu_bf16
+
```
mgoin, line 314: it would be nice if we could forward the properties from the `ARCH` class somehow instead of needing to duplicate  them each time

mgoin (commented):
Not saying we have to do this now, but this method of duplicating cpu properties across `cpu.py` and `engine.py` seems like a foot-gun as options like `cpu_neon_compatible()` have been left behind already. We should strive for a single interface for the cpu architecture @bnellnm

PR #601, opened 2022-08-22
[Commercial Server] New server config & server code

corey-nm:
Summary of changes:
- Add new EndpointConfig and ServerConfig objects
    - Removed integration field - this should be handled outside of these scripts
    - Added bucketing configuration
- Add conversion of EndpointConfig to PipelineConfig
- Move server app building logic into `server/server.py`
- Move cli logic into `server/cli.py`
- Split cli into two subcommands:
    - `task` which takes CLI options and a task and runs a server with a single model
    - `config` which only takes a config file
- endpoints for files are automatically added if `from_files` exists on input_schema
- `/invocations` is automatically added if `from_files` exists and there's only 1 model
- `Context` is now initialized with num_streams and scheduler set.
- Removed passing configuration through environment variables
- uvicorn hard coded to only use 1 worker
- setting ThreadPoolExecutor to have same number of workers as `Context.num_streams`
- Make model_path optional for `task` subcommand

Example configuration:
```yaml
num_cores: 16
num_workers: 8
endpoints:
  - name: main qa static model
    endpoint: /predict/static
    task: question-answering
    batch_size: 1
    bucketing:
      sequence_lengths: [10, 128]
```

Example commands:
```
deepsparse.server config --host 0.0.0.0 --port 8080 server-config.yaml
```
```
deepsparse.server task qa
```


src/deepsparse/server_v2/main.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+
+
+def _endpoint_config_to_pipeline_config(model: EndpointConfig) -> PipelineConfig:
+    if model.batch_size == 1 and model.accept_multiples_of_batch_size:
+        # dynamic batch
+        batch_size = None
+    elif model.batch_size > 1 and model.accept_multiples_of_batch_size:
+        # should still be dynamic batch
+        raise NotImplementedError
```
corey-nm, line 185: This can be handled in another PR for commercial server. In theory we should be able to split up input that is divisible by batch size

src/deepsparse/server_v2/config.py
```diff
@@ -0,0 +1,182 @@
...
+
+    task: str = Field(description="Task this endpoint performers")
+
+    model: str = Field(description="Location of the underlying model to use.")
+
+    batch_size: int = Field(
+        default=1, description="The batch size to compile the model for."
+    )
+
+    accept_multiples_of_batch_size: bool = Field(
```
corey-nm, line 67: This field can be removed since #604 adds dynamic batching for any batch size

tests/test_server/test_endpoints.py
```diff
@@ -0,0 +1,200 @@
...
+                #     batch_size=1,
+                # ),
+                # EndpointConfig(
+                #     name="test endpoint 3",
+                #     endpoint="/predict/bucketed",
+                #     task="text-classification",
+                #     model=Pipeline.default_model_for("text-classification"),
+                #     batch_size=1,
+                #     bucketing=SequenceLengthsConfig(sequence_lengths=[2, 4]),
+                # ),
```
rahul-tuli, line 160: Remove or add back in?

tests/test_server/test_endpoints.py
```diff
@@ -0,0 +1,200 @@
...
+    #         },
+    #     )
+    #     assert response.status_code == 200
+    #     assert response.json() == {
+    #         "labels": ["LABEL_1", "LABEL_0", "LABEL_1"],
+    #         "scores": [0.9998027682304382, 0.9995161890983582, 0.9998027682304382],
+    #     }
+
+    # def test_bucketed_request(self, client):
+    #     ...
```
rahul-tuli, line 200: remove or add back in?

src/deepsparse/engine.py
```diff
@@ -597,6 +598,9 @@ def num_streams(self):
     def scheduler(self):
         return self._scheduler
 
+    def __repr__(self) -> str:
+        return f"Context(num_cores={self.num_cores}, num_streams={self.num_streams})"
```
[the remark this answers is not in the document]
corey-nm: Good idea!

tests/test_server/test_endpoints.py
corey-nm: Oh yeah thanks for reminding me, I wanted to add more tests here. Ideally we could get the mock config PR before this so they aren't super slow.

tests/test_server/test_endpoints.py
corey-nm: Will add back in thanks

src/deepsparse/server/cli.py
```diff
@@ -0,0 +1,194 @@
...
+    help="The batch size to serve the model from model_path with",
+)
+@click.option(
+    "--num-cores",
+    type=int,
+    default=1,
+    help="The number of cores the server should have access to.",
+)
+@click.option(
+    "--num-workers",
```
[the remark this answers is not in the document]
corey-nm: I just tried, it actually does already support this thanks to that context thing we added before!!

tests/test_server/test_endpoints.py
corey-nm: Okay so I'm taking out bucketing & dynamic batch testing at this server level. dynamic batch test is below, but will require the #604 PR to be merged. Bucketing seems to have some issues so we can address that and add tests in a separate PR.

```python
                EndpointConfig(
                    name="test endpoint 3",
                    endpoint="/predict/bucketed",
                    task="text-classification",
                    model=stub,
                    batch_size=1,
                    bucketing=SequenceLengthsConfig(sequence_lengths=[2, 4]),
                ),

    def test_dynamic_batch_handles_any_batch_size(self, client):
        sentences = [
            "today is great",
            "today is terrible",
            "today is average",
            "just okay today, thanks",
            "is this thing working??",
        ]
        for num_items in range(1, len(sentences) + 1):
            response = client.post(
                "/predict/dynamic-batch",
                json={"sequences": sentences[:num_items]},
            )
            assert response.status_code == 200
            output = response.json()
            assert len(output["labels"]) == num_items
            assert len(output["scores"]) == num_items
```

corey-nm: Added a task in asana for this

src/deepsparse/engine.py
```diff
@@ -572,9 +572,10 @@ def __init__(
         self,
         num_cores: int = None,
         num_streams: int = None,
+        scheduler: Scheduler = Scheduler.elastic,
```
bfineran, line 575: mentioned offline - we should only be supporting elastic scheduler with `Context`

src/deepsparse/pipeline.py
```diff
@@ -312,6 +298,17 @@ def create(
             **kwargs,
         )
 
+    @classmethod
+    def default_model_for(cls, task: str) -> Optional[str]:
```
bfineran, line 302: what's the intended usage for this? note with auto install it may have some unintended side effects like installing the nm-transformers fork a bit early. if it's just to enable optional `model_path` in the cli, let's support a `"default"` special string that will let us get through the config schema

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,159 @@
...
+        # NOTE: only want to have 1 server process so models aren't being copied
+        workers=1,
+    )
+
+
+def _build_app(server_config: ServerConfig) -> FastAPI:
+    context = Context(
+        num_cores=server_config.num_cores,
+        num_streams=server_config.num_workers,
+        scheduler=Scheduler.multi_stream,
```
bfineran, line 73: should remove and require elastic

src/deepsparse/pipeline.py
```diff
@@ -696,10 +693,11 @@ class PipelineConfig(BaseModel):
             "specifies all available cores. Default is None"
         ),
     )
-    scheduler: str = Field(
-        default="async",
+    scheduler: Scheduler = Field(
+        default=Scheduler.single_stream,
```
bfineran, line 697: this is fine in controlled environments for getting the best single model numbers - also since we're always overriding with a context (elastic) in the server. however might want to think about how we can detect misuse and warn (ie initiating multiple single stream pipelines that are intended to run asynchronously )

src/deepsparse/server/config.py
```diff
@@ -12,152 +12,151 @@
...
+    if isinstance(bucketing, SequenceLengthsConfig):
+        if not SupportedTasks.is_nlp(task):
+            raise ValueError(f"SequenceLengthConfig specified for non-nlp task {task}")
+
+        return _unpack_nlp_bucketing(bucketing)
+    elif isinstance(bucketing, ImageSizesConfig):
+        if (
+            not SupportedTasks.is_image_classification(task)
+            and not SupportedTasks.is_yolo(task)
+            and not SupportedTasks.is_yolact(task)
```
bfineran, line 124: let's add an `is_cv` helper, probably won't be the last time we use this if statement / add another cv pipeline

src/deepsparse/server/config.py
```diff
@@ -12,152 +12,151 @@
...
-        config.models.append(
-            PipelineConfig(
-                task=config_dict["task"],
-                model_path=config_dict["model_path"],
-                batch_size=config_dict["batch_size"],
-            )
+    def to_pipeline_config(self) -> PipelineConfig:
+        input_shapes, kwargs = _unpack_bucketing(self.task, self.bucketing)
+
+        return PipelineConfig(
```
bfineran, line 82: we should set the pipeline `alias` to the `name` or `endpoint` value here, that way the logger can have something easy to log the pipeline name as

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,159 @@
...
+    @app.get("/ping", tags=["general"], response_model=bool)
+    @app.get("/health", tags=["general"], response_model=bool)
+    @app.get("/healthcheck", tags=["general"], response_model=bool)
+    @app.get("/status", tags=["general"], response_model=bool)
+    def _health():
+        return True
+
+    app._deepsparse_pool = ThreadPoolExecutor(max_workers=context.num_streams)
+
+    add_invocations_endpoint = len(server_config.endpoints) == 1
```
bfineran, line 97: the logic for re-routing the endpoint path with /invocations for sagemaker support was taken out. need to get some functionality for this added back in and update the existing sagemaker documentation/dockerfile if the location of where to set integration has changed. ping if any questions

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,159 @@
...
+        @app.post(file_route, tags=["predict"], response_model=output_schema)
+        async def _predict_from_files_func(request: List[UploadFile]):
+            request = pipeline.input_schema.from_files(
+                (file.file for file in request), from_server=True
+            )
+            return await get_running_loop().run_in_executor(pool, pipeline, request)
+
+        _LOGGER.info(f"Added '{file_route}' endpoint")
+
+    if add_invocations_endpoint and hasattr(input_schema, "from_files"):
```
bfineran, line 153: we need `invocations` for situations besides from_files, such as nlp which doesn't currently support from_files

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,159 @@
...
+
+    pool: ThreadPoolExecutor = app._deepsparse_pool
+
+    route = endpoint_config.endpoint or "/predict"
+    if not route.startswith("/"):
+        route = "/" + route
+
+    @app.post(route, tags=["predict"], response_model=output_schema)
+    async def _predict_func(request: pipeline.input_schema):
+        return await get_running_loop().run_in_executor(pool, pipeline, request)
```
bfineran, line 137: will using the same threadpool for the run loop as well as inside the pipeline (ie for dyamic) give us an inherent "minus 1" on concurrent inferences since one worker will be dedicated to this main predict func?

bfineran (changes requested):

bfineran:
additionally, docs seem fine now (outside of sagemaker) in terms of maintianing existing functionality. but should start thinking about how we want to structure docs around configuring for features such as bucketing, dynamic, custom w/ server, etc

src/deepsparse/server/server.py
corey-nm: Gotcha, I was translating the logic from the existing server, which only allows file uploads with sagemaker. You okay with doing sagemaker cleanup in a separate PR?

src/deepsparse/server/server.py
corey-nm: Hmm good question. Unclear how this all works with multiple concurrent server requests. We may also just be able to remove this async run_in_executor thing and do `pipeline(request)`. I think this needs some benchmarking to say.

src/deepsparse/server/server.py
corey-nm: Can you expand? The server will add an /invocations endpoint if there's only 1 endpoint, was there something else I missed bringing over?

src/deepsparse/pipeline.py
corey-nm: I can revert these changes from the PR - don't set the scheduler field of PipelineConfig anymore

src/deepsparse/pipeline.py
corey-nm: Yeah just for optional model path, good idea

src/deepsparse/server/server.py
bfineran: surprising, our advertised examples use the nlp pipelines which don't have file support IIRC, maybe this was a late introduction post the last time we tested sagemaker

src/deepsparse/server/server.py
bfineran: missing functionality expanded on in thread below. and would be good to add/keep in a pathway for specifying deployment scenario (can be above the server config level, like CLI) because for something like multi-model sagemaker we'll want to know and detect when to add in the extra sagemaker routes. having knowledge of integrations is also important for choosing which logger integration(s) to use/add in

additionally, don't think it's necessary to always add in `invocations` for single model servers since for non sagemaker deployments it will just be extra spam in the auto docs/swagger ui

src/deepsparse/server/server.py
corey-nm: Ahh I was interpretting code incorrectly - nevermind!

src/deepsparse/server/server.py
corey-nm: Okay I added integration field back in, and now the logic for sagemaker should match what was there before

src/deepsparse/server/config.py
corey-nm: Can we do this in another PR? don't want to add an additional file to this PR to review

src/deepsparse/server/server.py
corey-nm: I just ran some benchmarking, and this version performed the same as just calling `pipeline(request)`, so I replaced it to not use the async version.

src/deepsparse/engine.py
bfineran: we should just remove this, no?

corey-nm: Oh yeah - will revert now

src/deepsparse/server/config.py
bfineran: it should be a minimal change with basically this same code no?  Just worried about this falling through the cracks. open an asana ticket if you feel strongly

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,186 @@
...
+    if server_config.integration not in INTEGRATIONS:
+        raise ValueError(
+            f"Unknown integration field {server_config.integration}. "
+            f"Expected one of {INTEGRATIONS}"
+        )
+
+    context = Context(
+        num_cores=server_config.num_cores,
+        num_streams=server_config.num_workers,
+        scheduler=Scheduler.elastic,
```
bfineran, line 97: was removed - let's make sure to find any other usages of this kwarg

bfineran (commented):
lgtm overall pending comment

src/deepsparse/server/config.py
corey-nm: Added

src/deepsparse/server/server.py
corey-nm: Updated - no other usages

.github/workflows/test-check.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: "Clean sparsezoo directory"
         run: rm -r sparsezoo/
       - name: ⚙️ Install dependencies
-        run: pip3 install .[dev] opencv-python
```
dbogunowicz, line 34: I think it would be cleaner to separate rudimentary deepsparse tests from server tests, especially now when the codebase for the server will grow quickly. Not a blocking comment, but something to think about.

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,185 @@
...
+
+    app = _build_app(server_config)
+
+    uvicorn.run(
+        app,
+        host=host,
+        port=port,
+        log_level=log_level,
+        log_config=log_config,
+        # NOTE: only want to have 1 server process so models aren't being copied
```
dbogunowicz, line 70: Could you elaborate? What would happen if `workers=2`?

src/deepsparse/server/server.py
```diff
@@ -0,0 +1,185 @@
...
+    @app.get("/status", tags=["general"], response_model=bool)
+    def _health():
+        return True
+
+    # fill in names if there are none
+    for idx, endpoint_config in enumerate(server_config.endpoints):
+        if endpoint_config.name is None:
+            endpoint_config.name = f"endpoint-{idx}"
+
+    # creat pipelines & endpoints
```
dbogunowicz, line 125:
```suggestion
    # create pipelines & endpoints
```

src/deepsparse/server/main.py
```diff
@@ -1,352 +0,0 @@
-# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#    http://www.apache.org/licenses/LICENSE-2.0
```
dbogunowicz, line 7: Are we sure that landing that PR will be identical to replacing the old server? Maybe it makes sense to have them both live concurrently (something like server and server_v2) for a while.

src/deepsparse/server/config.py
```diff
@@ -12,152 +12,161 @@
...
+
+    task: str = Field(description="Task this endpoint performs")
+
+    model: str = Field(description="Location of the underlying model to use.")
+
+    batch_size: int = Field(
+        default=1, description="The batch size to compile the model for."
+    )
+
+    bucketing: Optional[Union[ImageSizesConfig, SequenceLengthsConfig]] = Field(
```
dbogunowicz, line 75: nit: not `Union[Optional[ImageSizesConfig], Optional[SequenceLenghtsConfig]]?

src/deepsparse/server/config.py
```diff
@@ -12,152 +12,161 @@
...
-        config_dict = json.loads(config_file.replace(ENV_SINGLE_PREFIX, ""))
-        config = ServerConfig()
-        config.models.append(
-            PipelineConfig(
-                task=config_dict["task"],
-                model_path=config_dict["model_path"],
-                batch_size=config_dict["batch_size"],
+        return _unpack_nlp_bucketing(bucketing)
+    elif isinstance(bucketing, ImageSizesConfig):
+        if not SupportedTasks.is_cv(task):
```
dbogunowicz, line 135: Good call on unifying the set of cv tasks.

src/deepsparse/pipeline.py
```diff
@@ -177,21 +177,7 @@ def __init__(
         self.engine = self._initialize_engine()
 
     def __call__(self, *args, **kwargs) -> Union[BaseModel, Future]:
```
dbogunowicz, line 179: I understand that this predates the `unify-run-split-join` branch and will be updated once the branch lands?

dbogunowicz (commented):
Great job, very clean implementation.
Left a few nits, ping me aggressively for further reviews.

src/deepsparse/pipeline.py
corey-nm: Yes, though I believe we'll be landing this PR first, will make sure its updated either way

src/deepsparse/server/server.py
corey-nm: num_workers for uvicorn or other server things usually means multiple processes handling requests. In our case this is bad because *I think* each server process would have it's own copy of context/engine. Whereas we want to ensure that there's only 1 context/engine

src/deepsparse/server/main.py
corey-nm: Yeah good question. I think this is another big reason to have unit/integration tests, otherwise all you can *really* do is try to side-by-side compare things haha. Are there specific cases that you are wondering about?

.github/workflows/test-check.yaml
corey-nm: Yeah I could even see the server being in a separate repo

src/deepsparse/server/config.py
corey-nm: I prefer having the optional on the outside because the whole field is optional. Otherwise it could be valid to have a non-optional field in the union (e.g. `Union[Optional[ImageSizesConfig], SequenceLengthsConfig]`), which is not what the intention is.

bfineran:
this command needs update as well (might be good to do a code search on invocations of the server in general): https://github.com/neuralmagic/deepsparse/blob/8b3426d2ff2551b0c07faad1a8c79797d314b633/examples/aws-sagemaker/Dockerfile#L27

bfineran:
LGTM after docs update, merge conflict, and some simple benchmarks added to PR description. Great work @corey-nm

src/deepsparse/pipeline.py
```diff
@@ -369,7 +369,7 @@ def create(
         pipeline_constructor = Pipeline._get_task_constructor(task)
 
         if (
-            model_path is None
+            (model_path is None or model_path == "default")
```
rahul-tuli, line 372: nice!

src/deepsparse/server/config.py
```diff
@@ -12,152 +12,161 @@
...
-                "model_path": model_path,
-                "batch_size": batch_size,
-                "integration": integration,
-            }
+def _unpack_cv_bucketing(cfg: ImageSizesConfig):
+    if len(cfg.image_sizes) == 0:
+        raise ValueError("Must specify at least one image size under bucketing")
+
+    if len(cfg.image_sizes) == 1:
+        # NOTE: convert from List[Tuple[int, int]] to List[List[int]]
```
rahul-tuli, line 164: nit:
```suggestion
        # NOTE: convert from List[Tuple[int]] to List[List[int]]
```

src/deepsparse/tasks.py
```diff
@@ -166,6 +166,14 @@ def is_nlp(cls, task: str) -> bool:
         """
         return any([nlp_task.matches(task) for nlp_task in cls.nlp])
 
+    @classmethod
+    def is_cv(cls, task: str) -> bool:
+        return (
+            cls.is_yolo(task)
+            or cls.is_yolact(task)
+            or cls.is_image_classification(task)
```
rahul-tuli, line 174: Python can be so beautiful

rahul-tuli (approved):
Approved! Nice tests.

src/deepsparse/tasks.py
corey-nm: there's nothing better than writing some beautiful code 😁

bfineran (approved):
LGTM @corey-nm quick thing to note / get in - we standardize to putting `--help` text(s) at the top docstring of the python files where their scripts are created (ie for the `cli.py` file)

PR #605, opened 2022-08-22
Twitter NER important token extraction example

mgoin:
Adapt `deepsparse/examples/twitter-nlp/analyze_sentiment.py` for NER/token classification.

The example below is trying to extract locations from a set of tweets.

Screen Shot 2022-08-22 at 12 43 48 PM
Screen Shot 2022-08-22 at 12 44 28 PM
Screen Shot 2022-08-22 at 12 44 41 PM


kylesayrs (dismissed):

examples/twitter-nlp/analyze_sentiment.py
```diff
@@ -124,30 +130,41 @@ def _display_results(batch, sentiments):
...
     tweets = _load_tweets(tweets_file)
     tweets = _prep_data(tweets, total_tweets)
     tot_sentiments = []
+    times = []
 
     while True:
         batch = _batched_model_input(tweets, batch_size)
         if batch is None:
             break
+        start = time.time()
```
KSGulin, line 165: Would recommend using time.perf_counter() here instead. It uses a much more recent epoch to increase floating point resolution on the nanosecond scale. Planning to replace all time.time() calls in deepsparse early this week

rahul-tuli (dismissed):

examples/twitter-nlp/analyze_tokens.py
```diff
@@ -0,0 +1,230 @@
...
+    tweets = _load_tweets(tweets_file)
+    tweets = _prep_data(tweets, total_tweets)
+    tot_tokens = []
+    times = []
+
+    while True:
+        batch = _batched_model_input(tweets, batch_size)
+        if batch is None:
+            break
+        start = time.time()
```
KSGulin, line 209: Same as above on using `time.perf_counter()`

examples/twitter-nlp/analyze_tokens.py
```diff
@@ -0,0 +1,230 @@
...
+    with open(tweets_file, "r") as file:
+        for line in file.readlines():
+            tweets.append(json.loads(line))
+
+    return tweets
+
+
+def _prep_data(tweets: List[Dict], total_num: int) -> List[str]:
+    if total_num:
+        tweets = islice(cycle(tweets), total_num)
```
KSGulin, line 94: Nice

KSGulin (changes requested):
Clean and satisfying example. Other than a comment on timing looks good to me

examples/twitter-nlp/analyze_sentiment.py
mgoin: Didn't know about this, I'm all for this. It does seem to be slower in a microbenchmark but not by much.
https://stackoverflow.com/a/71068984

KSGulin (approved):

examples/twitter-nlp/analyze_sentiment.py
KSGulin: Interesting find, that's good to know

spacemanidol (approved):

PR #606, opened 2022-08-23
Adding `@mock_engine()` to avoid compiling models in tests

corey-nm:
This adds a feature to patch `LIB.deepsparse_engine()` with our own fake class that reads onnx models and returns numpy arrays in the expected shape.

This will enable unit tests to avoid expensive compiles, however they will still download the `.onnx` files.

### DIFF
--- a/tests/deepsparse/pipelines/test_async_pipeline.py
+++ b/tests/deepsparse/pipelines/test_async_pipeline.py
@@ -17,6 +17,7 @@ from concurrent.futures import Future, ThreadPoolExecutor
 
 import pytest
 from deepsparse import Pipeline
+from tests.utils import mock_engine
 
 
 @pytest.fixture(scope="module")
@@ -24,7 +25,8 @@ def pipeline():
     """
     Auto-del fixture for Sequential Pipeline
     """
-    yield Pipeline.create(task="question-answering")
+    with mock_engine(rng_seed=0):
+        yield Pipeline.create(task="question-answering")
 
 
 @pytest.fixture(scope="module")
@@ -40,7 +42,8 @@ def threaded_pipeline(executor):
     """
     Auto-del fixture for Threaded Pipeline
     """
-    yield Pipeline.create(task="question-answering", executor=executor)
+    with mock_engine(rng_seed=0):
+        yield Pipeline.create(task="question-answering", executor=executor)
 
 
 @pytest.fixture(scope="module")
--- a/tests/utils/engine_mocking.py
+++ b/tests/utils/engine_mocking.py
@@ -42,16 +42,24 @@ def mock_engine(*, rng_seed: int, mode: SampleMode = SampleMode.rand, **kwargs):
     Use like you would a regular `unittest.mock.patch`
     ```python
     @mock_engine(rng_seed=0)
-    def test_something(engine):
+    def test_something(engine_mock: MagicMock):
         ...
     ```
 
     If you want the arrays to be zeros instead of random data:
     ```python
     @mock_engine(rng_seed=0, model=SampleMode.zeros)
-    def test_something(engine):
+    def test_something(engine_mock: MagicMock):
         ...
     ```
+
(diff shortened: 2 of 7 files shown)


corey-nm:
I think we could actually use this in some of the other tests like test_engine.py or test_benchmark.py. Since mock_engine patches the underlying LIB.deepsparse_engine, it can be used to test the Engine/MultiModelEngine classes without compiling stuff.

tests/utils/engine_mocking.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+
+    def deepsparse_engine(*args, **kwargs):
+        return _FakeDeepsparseLibEngine(*args, **kwargs, rng_seed=rng_seed, mode=mode)
+
+    return mock.patch(
+        "deepsparse.engine.LIB.deepsparse_engine",
+        side_effect=deepsparse_engine,
+        **kwargs,
+    )
```
rahul-tuli, line 64: nice!

rahul-tuli (approved):
Nice! could we update the description with an example on how to use the mocked engine in a dummy test? It will be a nice `TLDR;` resource. Thanks!

tests/utils/engine_mocking.py
```diff
@@ -0,0 +1,181 @@
...
+    zeros = "zeros"
+    rand = "rand"
+
+
+def mock_engine(*, rng_seed: int, mode: SampleMode = SampleMode.rand, **kwargs):
+    """
+    Intended to create a fake engine instead of compiling the model.
+
+    This patches the `LIB.deepsparse_engine()` that is used in deepsparse.Engine
+    with `_FakeDeepsparseLibEngine`.
```
mgoin, line 40: Python is an amazing evil

corey-nm: 😁 I like it for unit testing since it lets you isolate parts of the code that you want the test to focus on even if the code's api doesn't allow for that. Of course I also think code should be written so that its easy to test without doing stuff like this, but that's not always feasible.

rahul-tuli (dismissed):

tests/utils/engine_mocking.py
```diff
@@ -0,0 +1,181 @@
...
+            # fix the batch dimension in outputs.
+            #
+            # Assumes the first dimension is batch dimension!!
+            # However in general we cannot assume that all outputs have
+            # a batch dimension, that's why we need onnxruntime here.
+            for input in inputs:
+                input.type.tensor_type.shape.dim[0].dim_value = batch_size
+
+            batched_model_path = str(Path(tmp_dir) / "batched_model.onnx")
+            onnx.save(model, batched_model_path)
```
mgoin, line 120: is there a reason to not use `override_onnx_batch_size` here? It seems to do a similar action. Understood why you need to put it through ORT later. Note that this will not give you valid shapes if some input/output dimensions are not static or the model's output shape is based on input (like variable num boxes in object detection models)

corey-nm: Yes its the same functionality, but I wanted more control over the path that override_onnx_batch_size saved to (so it was under `tmp_dir`). Probably not a good enough reason to duplicate here in retrospect, I'll use the existing function!

tests/deepsparse/pipelines/test_async_pipeline.py
```diff
@@ -40,7 +42,8 @@ def threaded_pipeline(executor):
     """
     Auto-del fixture for Threaded Pipeline
     """
-    yield Pipeline.create(task="question-answering", executor=executor)
+    with mock_engine(rng_seed=0):
+        yield Pipeline.create(task="question-answering", executor=executor)
```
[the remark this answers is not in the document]
corey-nm: There's some weird interaction with the pytest.fixture decorator. When used together the unit tests below receive a the generator object instead of the pipeline.

mgoin (approved):

rahul-tuli (approved):

PR #607, opened 2022-08-23
Remove click versions from everywhere except `setup.py`

rahul-tuli:
This PR removes `click` versions from everywhere except `setup.py` to have one source of truth where `click` `version` is controlled


examples/aws-sagemaker/requirements.txt
```diff
@@ -1,3 +1,3 @@
 rich==12.3.0
 boto3==1.22.5
-click<8.1
\ No newline at end of file
+click
```
bfineran, line 3: why not delete?

rahul-tuli:
```suggestion
```

markurtz:
@rahul-tuli let's get all references deleted now given that all installs should require it now

bfineran (commented):
actually, thinking about tutorials like `aws-sagemaker`, deepsparse is not actually installed by the user (installed within a container in the tooling) so `click` won't be available to run the examples

examples/aws-sagemaker/requirements.txt
```diff
@@ -1,3 +1,3 @@
 rich==12.3.0
 boto3==1.22.5
-click<8.1
\ No newline at end of file
+click>=8.0
```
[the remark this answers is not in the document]
bfineran: net net, click is required for this example and not installed via deepsparse. if a user were to install deepsparse as well, the pinned version would take priority. when click is upgraded in deepsparse we shouldn't need to update this line later

bfineran (approved):

corey-nm (dismissed):

bfineran (approved):

KSGulin (approved):

corey-nm (approved):

PR #609, opened 2022-08-23
Unpin Haystack dependencies

kylesayrs:
After encountering issues attempting to install Haystack and its dependencies with different versions of python, it was decided to unpin the specific versions and instead match the dependency versioning provided by Haystack 1.4.0.

Installation was tested and successful with python3.8 and python3.9. The readme docs will specify that deepsparse[haystack] should only be installed with python3.8 pending better torch support in sparseml for other python versions.


dhuangnm:
Just wondering do we really need that many number of dependencies?

kylesayrs:
@dhuangnm These are all the dependencies that are required for `haystack[all]`. Theoretically we could get by with the core haystack dependencies, but then users would have to reinstall haystack to get features like Faiss and Weaviate.

dhuangnm:
I see, can we also get the cherry-pick PR up and landed today so we can pick up the fix in the engine?

bfineran (approved):

src/deepsparse/transformers/haystack/haystack_reqs.txt
```diff
@@ -8,280 +8,76 @@
...
-alembic==1.8.0
-appdirs==1.4.4
-astroid==2.11.7
-asttokens==2.0.5
-async-generator==1.10
-async-timeout==4.0.2
-attrs==21.4.0
-audioread==2.1.9
+importlib-metadata
+#torch>1.9,<1.11
```
rahul-tuli, line 19: Nit: remove

src/deepsparse/transformers/haystack/haystack_reqs.txt
```diff
@@ -8,280 +8,76 @@
...
+torch==1.9.1
+requests
+pydantic
+#transformers==4.13.0
```
rahul-tuli, line 23: Nit: remove

rahul-tuli (approved):

rahul-tuli (approved):

bfineran (approved):

PR #617, opened 2022-08-29
Using executor in cv pipeline preprocessing

corey-nm:
As part of #604, preprocessing was moved out of async threadpool execution. This PR changes the CV pipelines (image classification, yolo, yolact), to use `Pipeline.executor` in their preprocessing scripts. This is a performance improvement for static batch, because previously async preprocessing was not done for static batch.

Note: This only makes a difference when the pipelines need to load images. If pipelines receive all numpy arrays, then there is no performance gain when using async preprocessing.

### DIFF
--- a/src/deepsparse/image_classification/pipelines.py
+++ b/src/deepsparse/image_classification/pipelines.py
@@ -148,55 +148,10 @@ class ImageClassificationPipeline(Pipeline):
         if isinstance(inputs.images, numpy.ndarray):
             image_batch = inputs.images
         else:
-
-            image_batch = []
-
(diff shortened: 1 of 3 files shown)


corey-nm:
Note: this is blocked by #604 because that PR makes it so executor always exists. Currently on main it's possible executor is None

bfineran:
looks like the rebase should also unblock the tests

rahul-tuli (commented):
Lgtm pending rebase!

KSGulin (approved):

bfineran (approved):

PR #618, opened 2022-08-29
Updating documentation to use new deepsparse.server syntax

corey-nm:


examples/aws-sagemaker/config.yaml
```diff
@@ -1,7 +1,10 @@
-models:
+num_cores: 1
+num_workers: 1
+integration: sagemaker
```
corey-nm, line 3: btw @bfineran, do we even need this if we can just directly set the route below to `/invocations`?

PR #620, opened 2022-08-30
Make YOLACT return masks by default

dbogunowicz:
The user expects to receive masks by default, the code should reflect it. This patch also makes annotation for YOLACT work again.


KSGulin (approved):

rahul-tuli (approved):

rahul-tuli:
LGTM, pending two test failures

PR #623, opened 2022-08-31
Change time.time() -> time.perf_counter() for higher precision

KSGulin:
This PR replaces all calls to `time.time()` with calls to `time.perf_counter()`. `perf_counter()` provides a higher resolution time reading by using CPU counter relative time, rather than absolute time since UNIX epoch as is done by `time()`

**Testing**
Covered by existing tests


corey-nm (approved):
🎉

mgoin (approved):
nice!

PR #632, opened 2022-09-06
Adding hooks for config monitoring functionality

corey-nm:
This enables you to spin up a separate process/thread that will watch the configured file for changes, and if changes are detected will send requests to dynamically add/remove endpoints.

Integration into server/cli will be a separate PR.

Tested via unit tests & manual editing of server config.

### DIFF
--- a/src/deepsparse/server/monitoring.py
+++ b/src/deepsparse/server/monitoring.py
@@ -1,0 +1,131 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import multiprocessing
+import time
+from pathlib import Path
+from typing import List, Optional, Tuple
+
+import pydantic
+import requests
+import yaml
+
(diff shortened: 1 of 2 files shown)


src/deepsparse/server/monitoring.py
```diff
@@ -0,0 +1,131 @@
...
+            endpoints_url=endpoints_url,
+            check_interval_s=check_interval_s,
+        ),
+        daemon=True,
+    )
+    proc.start()
+    return proc
+
+
+def _file_watcher(config_path: str, endpoints_url: str, check_interval_s: float):
```
corey-nm, line 59: Unsure of how to test this, thoughts? Open to restructuring this function to make it easier to test. I think the `while True` currently makes it a bit hard to test, and there's no output from this function, so you can't black box it.

src/deepsparse/server/monitoring.py
```diff
@@ -0,0 +1,131 @@
...
+import yaml
+
+from deepsparse.server.config import EndpointConfig, ServerConfig
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def start_file_watcher(
+    config_path: str, endpoints_url: str, check_interval_s: float
```
corey-nm, line 32: @bfineran i'm proposing that we monitor the yaml config path instead of a "model directory". I think allowing a model directory in addition to yaml config is confusing from user standpoint. A side benefit is that it's easier to support code wise.

bfineran (commented):
notes from offline discussion:
* server needs faithful to config - ie endpoint comparator should look at all fields and update accordingly
* we should be saving in memory or to local file timestamped history of configs that have been deployed for debugging
* in ^ that vein for logging, etc should we ever allow the model/params for a route to change, and if we do, how does that affect pipeline route logging

src/deepsparse/server/monitoring.py
bfineran: could  test through `start_file_watcher` with the update method mocked as long as there's a way we can verify that the mocked method is called

corey-nm:
Will pick this up at a later point

PR #637, opened 2022-09-08
Removing predictions from YOLOOutput

corey-nm:
This field just contains what all the other fields in the output contains, so its very redundant. 

Unit tested and searched codebase for any use of "predictions"


bfineran (changes requested):
looks like this doc needs update: https://github.com/neuralmagic/deepsparse/blob/87b2b5d900ac8de1ad348be72fc30f739b06c545/src/deepsparse/yolo/README.md?plain=1#L137

corey-nm:
> looks like this doc needs update:
> 
> https://github.com/neuralmagic/deepsparse/blob/87b2b5d900ac8de1ad348be72fc30f739b06c545/src/deepsparse/yolo/README.md?plain=1#L137

Fixed, great catch. That should've been using boxes anyway!

bfineran (approved):

KSGulin (approved):

PR #638, opened 2022-09-08
Adding thread pinning and torch.set_num_threads to ServerConfig

corey-nm:
Both of these values are very useful to tuning server performance. torch.set_num_threads(1) is a good default to make sure torch isn't contending with engine. threading pinning to cores for engine is enabled by default as well.

Tested with unit tests.

### DIFF
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -120,16 +120,19 @@ class ServerConfig(BaseModel):
         description="The kind of integration to use. local|sagemaker",
     )
 
-    engine_thread_pinning: bool = Field(
-        default=True,
-        description="Whether to pin engine threads to cores.",
+    engine_thread_pinning: str = Field(
+        default="core",
+        description=(
+            "Enable binding threads to cores ('core' the default), "
+            "threads to cores on sockets ('numa'), or disable ('none')"
+        ),
     )
 
     pytorch_num_threads: Optional[int] = Field(
         default=1,
         description=(
-            "Configures number of threads that pytorch is allowed to use. "
-            "Useful to reduce resource contention. "
+            "Configures number of threads that pytorch is allowed to use during"
+            "pre and post-processing. Useful to reduce resource contention. "
             "Set to `None` to place no restrictions on pytorch."
         ),
     )
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -120,6 +120,20 @@ class ServerConfig(BaseModel):
         description="The kind of integration to use. local|sagemaker",
     )
 
+    engine_thread_pinning: bool = Field(
+        default=True,
+        description="Whether to pin engine threads to cores.",
+    )
+
+    pytorch_num_threads: Optional[int] = Field(
+        default=1,
+        description=(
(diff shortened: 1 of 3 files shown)


src/deepsparse/server/config.py
```diff
@@ -120,6 +120,20 @@ class ServerConfig(BaseModel):
...
+    engine_thread_pinning: bool = Field(
+        default=True,
+        description="Whether to pin engine threads to cores.",
+    )
+
+    pytorch_num_threads: Optional[int] = Field(
+        default=1,
+        description=(
+            "Configures number of threads that pytorch is allowed to use. "
+            "Useful to reduce resource contention. "
```
bfineran, line 132: nit: let's specify this is for pre/post processing

src/deepsparse/server/config.py
```diff
@@ -120,6 +120,20 @@ class ServerConfig(BaseModel):
         description="The kind of integration to use. local|sagemaker",
     )
 
+    engine_thread_pinning: bool = Field(
```
bfineran, line 123: let's actually match here (see logic in the same file): https://github.com/neuralmagic/deepsparse/blob/87b2b5d900ac8de1ad348be72fc30f739b06c545/src/deepsparse/benchmark/benchmark_model.py#L203

forgot there's also the numa option

tests/server/test_endpoints.py
```diff
@@ -293,3 +295,42 @@ def test_dynamic_add_and_remove_endpoint(engine_mock):
...
+
+
+@patch.dict(os.environ, deepcopy(os.environ))
+def test_thread_pinning():
+    _build_app(
+        ServerConfig(
+            num_cores=1, num_workers=1, engine_thread_pinning=False, endpoints=[]
+        )
+    )
+    assert "NM_BIND_THREADS_TO_CORES" not in os.environ
```
bfineran, line 329: hm, a user could have this already set in their environment

corey-nm: oh true, ill pop it at the beginning of the test - the patch makes a deepcopy anyway so it shouldn't affect

tests/server/test_endpoints.py
```diff
@@ -293,3 +295,73 @@ def test_dynamic_add_and_remove_endpoint(engine_mock):
     )
     assert response.status_code == 200
     assert 404 == client.post("/predict", json=dict(sequences="asdf")).status_code
+
+
+def test_pytorch_num_threads():
+    try:
+        import torch
```
KSGulin, line 302: I recommend using `torch = pytest.importorskip("torch")` here ([docs](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip-ref)). That way it's clear the test is skipped rather than successfully passed

KSGulin (commented):
LGTM pending the one comment

tests/server/test_endpoints.py
corey-nm: OMG i could not figure out how to do this, thanks!!

corey-nm: Added! beautiful 😁

KSGulin (approved):

bfineran (approved):

PR #639, opened 2022-09-08
Small readability improvement for Pipeline.__call__

corey-nm:
Moving to use `self.executor.map` for consistency with preprocessing. Internally executor.map does the same thing.

### DIFF
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -18,7 +18,7 @@ inference engine and include pre/postprocessing
 """
 import os
 from abc import ABC, abstractmethod
-from concurrent.futures import ThreadPoolExecutor, wait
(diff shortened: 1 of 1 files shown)


bfineran (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -207,15 +207,10 @@ def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
         batches = self.split_engine_inputs(engine_inputs, self._batch_size)
 
         # submit to engine
```
bfineran, line 209:
```suggestion
        # submit split batches to engine threadpool
```

bfineran (approved):

rahul-tuli (approved):

PR #640, opened 2022-09-08
[deepsparse.server] utility functions for instantiating loggers

bfineran:
utility functions for instantiating loggers - main contribution is the ability to specify and instantiate loggers from a simple yaml config. yaml config should be the integration name followed by it's init kwargs

example config:
```yaml
prometheus:
        port: 8001
        text_log_save_dir: /home/deepsparse-server/prometheus
        text_log_save_freq: 30
```

**test_plan:**
pytest included

PR #641, opened 2022-09-08
[deepsparse.server] integrate pipeline logging into deepsparse server

bfineran:
integrates pipeline logging into `deepsparse.server` by adding options for logging config on instantiation, logging to the logger manager on route prediction with `run_with_monitoring`

**updated config** - a `loggers` field is added to the `ServerConfig` which can be a raw yaml logging config as defined in #640

**test_plan:**
pytests included for config update
currently manually verified default prometheus server is spun up from example server run

server command and logger instantiation log -
```bash
$ deepsparse.server     task sentiment-analysis     --model_path "zoo:nlp/sentiment_analysis/bert-base/pytorch/huggingface/sst2/12layer_pruned80_quant-none-vnni"
2022-09-08 18:37:56 deepsparse.server.server INFO     config_path: /tmp/tmpw7emk_vk/server-config.yaml
...
2022-09-08 18:37:56 deepsparse.loggers.prometheus_logger INFO     Prometheus client: started. Using port: 8000.
...
INFO:     Uvicorn running on http://0.0.0.0:5543 (Press CTRL+C to quit)
```

curl into prometheus server -
```bash
$ curl http://localhost:8000# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 5931.0
...
```

### DIFF
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -158,7 +158,7 @@ def config(config_path: str, host: str, port: int, log_level: str):
 )
 @click.option(
     "--no-loggers",
-    is_flag=bool,
+    is_flag=True,
     default=False,
     help=(
         "Set to not use any inference logging integration. Defaults to using "
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -158,7 +158,7 @@ def config(config_path: str, host: str, port: int, log_level: str):
 )
 @click.option(
     "--no-loggers",
-    type=bool,
+    is_flag=bool,
     default=False,
     help=(
         "Set to not use any inference logging integration. Defaults to using "
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -187,7 +187,7 @@ def task(
         endpoints=[
             EndpointConfig(
                 task=task,
-                name=f"{task} inference model",
+                name=f"{task}",
                 route="/predict",
                 model=model_path,
                 batch_size=batch_size,
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -156,6 +156,15 @@ def config(config_path: str, host: str, port: int, log_level: str):
     default="info",
     help="Sets the logging level.",
 )
+@click.option(
+    "--loggers",
+    type=str,
+    default="",
+    help=(
+        "Optional path to a config file defining server logger integrations "
+        "to initialize."
+    ),
+)
 def task(
     task: str,
     model_path: str,
@@ -165,11 +174,13 @@ def task(
     host: str,
     port: int,
     log_level: str,
+    loggers: str,
 ):
     """
     Run the server using configuration with CLI options,
     which can only serve a single model.
     """
+    loggers = loggers or None
     cfg = ServerConfig(
         num_cores=num_cores,
         num_workers=num_workers,
@@ -182,6 +193,7 @@ def task(
                 batch_size=batch_size,
             )
         ],
+        loggers=loggers,
     )
 
     with TemporaryDirectory() as tmp_dir:
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -157,12 +157,12 @@ def config(config_path: str, host: str, port: int, log_level: str):
     help="Sets the logging level.",
 )
 @click.option(
-    "--loggers",
-    type=str,
-    default="",
+    "--no-loggers",
+    type=bool,
+    default=False,
     help=(
-        "Optional path to a config file defining server logger integrations "
-        "to initialize."
+        "Set to not use any inference logging integration. Defaults to using "
+        "a default integration such as Prometheus."
     ),
 )
 def task(
@@ -174,13 +174,12 @@ def task(
     host: str,
     port: int,
     log_level: str,
-    loggers: str,
+    no_loggers: bool,
 ):
     """
     Run the server using configuration with CLI options,
     which can only serve a single model.
     """
-    loggers = loggers or None
     cfg = ServerConfig(
         num_cores=num_cores,
         num_workers=num_workers,
@@ -193,7 +192,7 @@ def task(
                 batch_size=batch_size,
             )
         ],
-        loggers=loggers,
+        loggers=None if no_loggers else "default",
     )
 
     with TemporaryDirectory() as tmp_dir:
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -125,7 +125,8 @@ class ServerConfig(BaseModel):
         default=None,
         description=(
             "Optional dictionary of logger integration names to initialization kwargs "
-            "or path to a yaml config file of the same specification. Default is None."
+            "or path to a yaml config file of the same specification. Example: "
+            "{'prometheus': {'port': 8001}}. Default is None."
         ),
     )
 
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -120,6 +120,23 @@ class ServerConfig(BaseModel):
         description="The kind of integration to use. local|sagemaker",
     )
 
+    engine_thread_pinning: str = Field(
+        default="core",
+        description=(
+            "Enable binding threads to cores ('core' the default), "
+            "threads to cores on sockets ('numa'), or disable ('none')"
+        ),
+    )
+
+    pytorch_num_threads: Optional[int] = Field(
+        default=1,
+        description=(
+            "Configures number of threads that pytorch is allowed to use during"
+            "pre and post-processing. Useful to reduce resource contention. "
+            "Set to `None` to place no restrictions on pytorch."
+        ),
+    )
+
     endpoints: List[EndpointConfig] = Field(description="The models to serve.")
     loggers: Union[Dict[str, Dict[str, Any]], str, None] = Field(
         default=None,
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -84,8 +84,14 @@ class EndpointConfig(BaseModel):
         ),
     )
 
+    kwargs: Dict[str, Any] = Field(
+        default={}, description="Additional arguments to pass to the Pipeline"
+    )
+
     def to_pipeline_config(self) -> PipelineConfig:
         input_shapes, kwargs = _unpack_bucketing(self.task, self.bucketing)
+
+        kwargs.update(self.kwargs)
 
         return PipelineConfig(
             task=self.task,
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -138,12 +138,14 @@ class ServerConfig(BaseModel):
     )
 
     endpoints: List[EndpointConfig] = Field(description="The models to serve.")
+
     loggers: Union[Dict[str, Dict[str, Any]], str, None] = Field(
-        default=None,
+        default="default",
         description=(
-            "Optional dictionary of logger integration names to initialization kwargs "
-            "or path to a yaml config file of the same specification. Example: "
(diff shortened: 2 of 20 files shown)


KSGulin (approved):

src/deepsparse/server/cli.py
```diff
@@ -156,6 +156,15 @@ def config(config_path: str, host: str, port: int, log_level: str):
     default="info",
     help="Sets the logging level.",
 )
+@click.option(
+    "--loggers",
+    type=str,
+    default="",
```
corey-nm, line 162:
```suggestion
    default=None,
```
And change loggers to Optional[str]?

Seems like this could remove the `loggers = loggers or None` line too

src/deepsparse/server/config.py
```diff
@@ -121,6 +121,13 @@ class ServerConfig(BaseModel):
     )
 
     endpoints: List[EndpointConfig] = Field(description="The models to serve.")
+    loggers: Union[Dict[str, Dict[str, Any]], str, None] = Field(
+        default=None,
+        description=(
+            "Optional dictionary of logger integration names to initialization kwargs "
+            "or path to a yaml config file of the same specification. Default is None."
+        ),
```
corey-nm, line 129:
```suggestion
    logger_config: Dict[str, Dict[str, Any]] = Field(
        default={},
        description=(
            "Optional dictionary of logger integration names to initialization kwargs. "
            "Default is empty dictionary."
        ),
```

src/deepsparse/server/cli.py
```diff
@@ -165,11 +174,13 @@ def task(
     host: str,
     port: int,
     log_level: str,
+    loggers: str,
 ):
     """
     Run the server using configuration with CLI options,
     which can only serve a single model.
     """
+    loggers = loggers or None
```
corey-nm, line 183: Recommend doing the loading of the yaml config in this function and passing the dictionary to server config.

```
loggers = {}
if loggers_path is not None:
    loggers = yaml.safe_load(...)
```

src/deepsparse/server/cli.py
```diff
@@ -156,6 +156,15 @@ def config(config_path: str, host: str, port: int, log_level: str):
     default="info",
     help="Sets the logging level.",
 )
+@click.option(
+    "--loggers",
```
corey-nm, line 160: To make it more explicit that its a file?
```suggestion
    "--loggers-path",
```

src/deepsparse/server/server.py
```diff
@@ -164,42 +170,54 @@ def _add_endpoint(
...
+
+    def _predict(request: pipeline.input_schema):
+        (
+            pipeline_outputs,
+            pipeline_inputs,
+            engine_inputs,
+            inference_timing,
+        ) = pipeline.run_with_monitoring(request)
+        pipeline_logger.log_latency(pipeline_name, inference_timing)
+        pipeline_logger.log_data((pipeline_inputs, engine_inputs), pipeline_outputs)
```
corey-nm, line 197: This needs pipeline_name from the other change in this PR

src/deepsparse/server/server.py
```diff
@@ -164,42 +170,54 @@ def _add_endpoint(
...
-        return pipeline(request)
+    pipeline_name = pipeline.alias or pipeline.task
+
+    def _predict(request: pipeline.input_schema):
+        (
+            pipeline_outputs,
+            pipeline_inputs,
+            engine_inputs,
+            inference_timing,
+        ) = pipeline.run_with_monitoring(request)
```
corey-nm, line 195: nit: figure out how to format this on less lines 😁

src/deepsparse/server/server.py
```diff
@@ -164,42 +170,54 @@ def _add_endpoint(
...
+        pipeline_logger.log_latency(pipeline_name, inference_timing)
+        pipeline_logger.log_data((pipeline_inputs, engine_inputs), pipeline_outputs)
+        return pipeline_outputs
 
     def _predict_from_files(request: List[UploadFile]):
         request = pipeline.input_schema.from_files(
             (file.file for file in request), from_server=True
         )
-        return pipeline(request)
+        return _predict(request)
```
corey-nm, line 204: +1 nice

corey-nm (changes requested):
I think the ServerConfig field could be simplified a bit to make it easier to understand from user perspective (and also easier code wise!)

Also have you run the server tests locally? Annoying that the workflows don't run if you aren't doing a PR into main

src/deepsparse/server/cli.py
bfineran: right, had this originally, but click didn't seem to like the optionals, will take another look

src/deepsparse/server/server.py
bfineran: I did try quite a few line breaks, but black keeps forcing back to here, and we've agreed to keep that our source of truth for formatting / breaks.  Don't want to keep the outputs as a tuple to unpack later as there's a more than likely change the outputs of `run_with_monitoring` will change and explicitly naming will make it easier to catch and fix issues when that happens rather than accessing via indexing

corey-nm: 👍

src/deepsparse/server/cli.py
bfineran: want to keep the name consistent with the server config field since the rest of the args are

src/deepsparse/server/config.py
bfineran: was planning on empty dict meaning no loggers, `None` meaning use the default logger for the integration

src/deepsparse/server/cli.py
bfineran: was planning on empty dict meaning no loggers, `None` meaning use the default logger for the integration so not sure we'd want to unwrap like that here

src/deepsparse/server/config.py
corey-nm: Ahh I see so then just the Optional[Dict]? Still think `str` should not be one of the options

src/deepsparse/server/cli.py
corey-nm: `None` being default loggers is really confusing from user perspective. recommend finding another way to specify default

src/deepsparse/server/config.py
```diff
@@ -121,6 +121,14 @@ class ServerConfig(BaseModel):
     )
 
     endpoints: List[EndpointConfig] = Field(description="The models to serve.")
+    loggers: Union[Dict[str, Dict[str, Any]], str, None] = Field(
```
corey-nm, line 124: Still think str should not be allowed here. Parsing of this should happen at a different level. Otherwise anything that uses ServerConfig.loggers has to handle parsing the str

bfineran: my thinking here is that if a user has a common logging config on their system they can reference it here. loggers by design already is not useful without being unwrapped by the `logger_manager_from_config` function, ie any use of `ServerConfig.loggers` would need to call into it

src/deepsparse/server/cli.py
bfineran: fair, the way to get the default would just be to omit `loggers` from the config, but agreed `null` vs empty dict isn't intuitive

src/deepsparse/server/config.py
corey-nm: Okay sure this makes sense then

src/deepsparse/server/cli.py
corey-nm: omitting loggers from config would set it as the default value right? which atm is None.

Thoughts on:
- loggers is omitted or set to None (null in yaml): no loggers `ManagerLogger(loggers=[])`
- loggers is a str or dict: current behavior of doing logger_manager_from_config
- ServerConfig.loggers Field.default is {"prometheus": {}}
- task cli loggers is an Optional[str] that defaults to None, meaning no loggers

bfineran: thinking through this proposal, need something that 1) gives a user an easy/default way to specify the default 2) let's us use the get default function (we'll need to change the default based on other info from the server config like the integration)

bfineran: updated, possible values:

`None` - no loggers
str - must be "default" for default (this is the default)
dict - config literal

additionally for simplicity in task CLI mode, logging is always default, added `no_logging` option to set loggers to None. (Otherwise would have required file support for a reasonable UX, can look into adding back in file support later as needed, but as discussed offline, probably not necessary for a while)

src/deepsparse/server/cli.py
```diff
@@ -156,6 +156,15 @@ def config(config_path: str, host: str, port: int, log_level: str):
     default="info",
     help="Sets the logging level.",
 )
+@click.option(
+    "--no-loggers",
+    type=bool,
```
corey-nm, line 161:
```suggestion
    is_flag=True,
```
https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags

src/deepsparse/server/cli.py
```diff
@@ -176,12 +186,13 @@ def task(
...
             EndpointConfig(
                 task=task,
-                name=f"{task} inference model",
+                name=f"{task}",
                 route="/predict",
                 model=model_path,
                 batch_size=batch_size,
             )
         ],
+        loggers=None if no_loggers else "default",
```
corey-nm, line 195: +1 I like this formulation from cli perspective a lot!

tests/server/test_app.py
```diff
@@ -125,6 +160,22 @@ def test_invalid_thread_pinning():
...
+                num_workers=1,
+                engine_thread_pinning="asdf",
+                endpoints=[],
+                loggers=None,
             )
         )
+
+
+@mock_engine(rng_seed=0)
+def test_default_logging_server_launches(engine_mock):
```
corey-nm, line 173: Can you move this test into the test_endpoints file? And then also use a `TestClient(_build_app(server_config))` to send requests instead of requests lib

tests/server/test_app.py
```diff
@@ -125,6 +160,22 @@ def test_invalid_thread_pinning():
...
+    server_config = ServerConfig(
+        num_cores=1, num_workers=1, endpoints=[], loggers="default"
+    )
+    _build_app(server_config)
+
+    # check positive ping to expected prometheus server on port 6100
+    response = requests.get("http://127.0.0.1:6100")
```
corey-nm, line 180: Are there other endpoints we are expecting to exist that we could throw in a quick test for?

corey-nm (changes requested):
lgtm, just some small moving around of unit tests

src/deepsparse/server/cli.py
corey-nm: Nice, i like it!

tests/server/test_app.py
bfineran: moving, can't use TestClient here because the prometheus server does not run on FastAPI (the request here isn't going to deepsparse)

tests/server/test_app.py
bfineran: there are, working out what the ones we want to hit are and will update in the graphana PR

corey-nm: Okay sounds good

corey-nm (dismissed):
👍

corey-nm (approved):

src/deepsparse/server/cli.py
```diff
@@ -156,6 +156,15 @@ def config(config_path: str, host: str, port: int, log_level: str):
     default="info",
     help="Sets the logging level.",
 )
+@click.option(
+    "--no-loggers",
+    is_flag=bool,
```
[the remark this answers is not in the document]
bfineran: Good catch

rahul-tuli (approved):

corey-nm (approved):

PR #644, opened 2022-09-09
Moving PIL and opencv-python to install requirements

corey-nm:
This moves PIL and cv2 to required dependencies to install deepsparse. They are used in multiple pipelines and the server.

test plan: unit tests


KSGulin (approved):

corey-nm:
Closing for now since we don't want to add opencv as a core dependency. Probably recommend some `deepsparse[cv]` feature for the future.

PR #651, opened 2022-09-15
Fixing bug with prometheus metric name

corey-nm:
https://prometheus.io/docs/practices/naming/

hyphens are not allowed in metric names.

test plan: added regression test, tested multi model example in readme


bfineran (approved):
LGTM - one thing just cross my mind we should add is shutting down the prometheus server at the end of the unit test. I think they may be staying alive longer than the test scope

corey-nm:
> LGTM - one thing just cross my mind we should add is shutting down the prometheus server at the end of the unit test. I think they may be staying alive longer than the test scope

It looks like its a daemon thread so I think it should be cleaned up? Do you see resources leftover?

rahul-tuli:
> https://prometheus.io/docs/practices/naming/
> 
> Underscores are not allowed in metric names.
> 
> test plan: added regression test, tested multi model example in readme

Could we update the description, I think underscores are allowed but hyphens are not

corey-nm:
> > https://prometheus.io/docs/practices/naming/
> > Underscores are not allowed in metric names.
> > test plan: added regression test, tested multi model example in readme
> 
> Could we update the description, I think underscores are allowed but hyphens are not

Oh yeah haha reversed it, fixed!

rahul-tuli (approved):
LGTM pending the description update!

PR #653, opened 2022-09-15
Adding example with model deployment directory in docker

corey-nm:
This shows users how to use the `/deployment` directory of a model inside docker.

Test plan: Run example docker build command from readme

### DIFF
--- a/examples/docker-deployment/Dockerfile
+++ b/examples/docker-deployment/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.9
+
+# install sparsezoo from github
+ARG SPARSEZOO_BRANCH=main
+RUN git clone https://github.com/neuralmagic/sparsezoo.git --depth 1 -b $SPARSEZOO_BRANCH
+RUN pip3 install --no-cache-dir --upgrade -e "./sparsezoo"
+
+# install deepsparse from github
+ARG DEEPSPARSE_BRANCH=main
+RUN git clone https://github.com/neuralmagic/deepsparse.git --depth 1 -b $DEEPSPARSE_BRANCH
+RUN pip3 install --no-cache-dir --upgrade -e "./deepsparse[server]"
+
+# download the deployment directory
+ARG MODEL_STUB
+RUN sparsezoo.download --save-dir /model $MODEL_STUB
+
+ARG TASK
+ENV TASK=$TASK
+ENTRYPOINT deepsparse.server task $TASK --model_path /model/deployment
--- a/examples/docker-deployment/README.md
+++ b/examples/docker-deployment/README.md
@@ -16,11 +16,7 @@ It is a directory with at least a `model.onnx` file in it, but usually other con
 
 ## How to create a Model Deployment?
 
-For the sake of examples, we will use this `question-answering` stub:
-
-```
-zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/pruned95_obs_quant-none
-```
+There are many options!
 
 #### Use sparsezoo.download
 
--- a/examples/docker-deployment/README.md
+++ b/examples/docker-deployment/README.md
@@ -1,0 +1,95 @@
+# Wrapping a Model Deployment with Docker
+
+This README outlines the many options you have for storing a deepsparse model inside
+a docker container.
+
+The [Dockerfile](Dockerfile) shows one case of this: running `sparsezoo.download`
+when building the image, and passing to `deepsparse.server task`. Example usage:
+
+```bash
+docker build --build-arg TASK=qa --build-arg MODEL_STUB=zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/pruned95_obs_quant-none -t qa .
+```
+
+## What is a Model Deployment?
+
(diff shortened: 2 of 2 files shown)


rahul-tuli (approved):
really like the clean `Dockerfile`, no unnecessary git history downloads. Kudos!

bfineran (commented):
any reason not to use a mount or volume when running our existing container instead of needing to edit the dockerfile with copy? mounting + updating the server config to point to the mounted location should give a pathway where users don't need to edit the out of the box dockerfile

corey-nm:
> any reason not to use a mount or volume when running our existing container instead of needing to edit the dockerfile with copy? mounting + updating the server config to point to the mounted location should give a pathway where users don't need to edit the out of the box dockerfile

Can you expand? Are you thinking of mounting a deployment directory to docker? I can add details of that to the README if so.

The dockerfile is intended to be an example that covers a decent amount of use cases, but is not exhaustive. IMO Dockerfile should be easy to understand so users can use it as a starting point.

corey-nm:
> any reason not to use a mount or volume when running our existing container instead of needing to edit the dockerfile with copy? mounting + updating the server config to point to the mounted location should give a pathway where users don't need to edit the out of the box dockerfile

Added mounting directory section in dockerfile

examples/docker-deployment/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM python:3.9
+
+# install sparsezoo from github
+ARG SPARSEZOO_BRANCH=main
+RUN git clone https://github.com/neuralmagic/sparsezoo.git --depth 1 -b $SPARSEZOO_BRANCH
```
dbogunowicz, line 5: why not pip install?

corey-nm: Yeah probably should... Right now the server command below only works for 1.2, which isn't released on pypi yet. So pip install would fail currently. After 1.2 is released we can go back and update this to use pip install

rahul-tuli (approved):

KSGulin (approved):

examples/docker-deployment/README.md
```diff
@@ -0,0 +1,122 @@
...
+# Wrapping a Model Deployment with Docker
+
+This README outlines the many options you have for storing a deepsparse model inside
+a docker container.
+
+The [Dockerfile](Dockerfile) shows one case of this: running `sparsezoo.download`
+when building the image, and passing to `deepsparse.server task`. Example usage:
+
+```bash
+docker build --build-arg TASK=qa --build-arg MODEL_STUB=zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/pruned95_obs_quant-none -t qa .
```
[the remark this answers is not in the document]
corey-nm: As far as I can tell yes, you have to specify them separately. Too bad there's no shorthand for it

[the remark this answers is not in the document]
dbogunowicz: :(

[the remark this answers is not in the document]
corey-nm: @dbogunowicz do you think this dockerfile is useful for the main example? there's a bunch of different options we could go with (as outlined in the readme), not sure what's the most clear/concise.

examples/docker-deployment/README.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+## What is a Model Deployment?
+
+It is a directory with at least a `model.onnx` file in it, but usually other configuration files.
```
dbogunowicz, line 31: In the future it would be great to link to a sample deployment folder in sparsezoo.

examples/docker-deployment/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+## How to create a Model Deployment?
+
+There are many options!
```
dbogunowicz, line 35: We will have at some point a doc for that I suppose...

dbogunowicz:
@corey-nm Can we also add a tutorial for the user to add the downloaded `deployment` folder as a volume? I could imagine that some users might not want to keep the bulky deployment folder (e.g. model.onnx) inside the image, but would rather keep the image light and just mount the `deployment` on top of the image.

examples/docker-deployment/Dockerfile
bfineran: would suggest going with pip when branch is not set, otherwise from git using the given branch

examples/docker-deployment/README.md
bfineran: might be good to give the user some idea when to choose a specific option

bfineran (commented):
lgtm pending comments

bfineran (approved):

corey-nm:
Wondering if the example should be split into two smaller examples so the readmes describe what the dockerfile contains, instead of being documentation:

1. `examples/docker/build-time-deployment/Dockerfile` (this would be the current Dockerfile that downloads as part of the build process)
2. `examples/docker/run-time-deployment/Dockerfile` (this would be one where the deployment directory is volume mounted)

corey-nm:
Closing for now - will reopen other version as discussed in asana ticket

PR #660, opened 2022-09-19
Adding file monitoring hooks

corey-nm:
Note: Not integrated into server yet.

Main hook is `start_file_watcher` for server to call into to kick off a watcher process. Everything else is just helpers for that.

The file watcher at a high level does the following:
1. Check every X seconds if the file has been modified
2. If so, check if any endpoints were added/removed/modified
3. If endpoints have changed, send requests to server to update it live
4. Save the old config as `<path>.v<version>`

test plan: unit tests

### DIFF
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -153,10 +153,9 @@ class ServerConfig(BaseModel):
     )
 
 
-def _endpoint_diff(
+def endpoint_diff(
     old: ServerConfig, new: ServerConfig
 ) -> Tuple[List[EndpointConfig], List[EndpointConfig]]:
-    """ """
     old_routes = {e.route: e for e in old.endpoints if e.route is not None}
     new_routes = {e.route: e for e in new.endpoints if e.route is not None}
 
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -152,6 +152,26 @@ class ServerConfig(BaseModel):
         ),
     )
 
+    def endpoint_diff(
+        self, new: "ServerConfig"
+    ) -> Tuple[List[EndpointConfig], List[EndpointConfig]]:
+        old_routes = {e.route: e for e in self.endpoints if e.route is not None}
+        new_routes = {e.route: e for e in new.endpoints if e.route is not None}
+
+        added_routes = set(new_routes) - set(old_routes)
+        removed_routes = set(old_routes) - set(new_routes)
+
+        # for any routes that are in both, check if the config object is different
+        # if so, then we do modification by adding the route to both remove & add
+        for route in set(new_routes) & set(old_routes):
+            if old_routes[route] != new_routes[route]:
+                removed_routes.add(route)
+                added_routes.add(route)
+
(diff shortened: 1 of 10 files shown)


src/deepsparse/server/config.py
```diff
@@ -153,6 +153,28 @@ class ServerConfig(BaseModel):
     )
 
 
+def _endpoint_diff(
+    old: ServerConfig, new: ServerConfig
+) -> Tuple[List[EndpointConfig], List[EndpointConfig]]:
+    """ """
```
rahul-tuli, line 159: nit: remove empty docstring

src/deepsparse/server/config.py
```diff
@@ -153,6 +153,28 @@ class ServerConfig(BaseModel):
...
+    old_routes = {e.route: e for e in old.endpoints if e.route is not None}
```
rahul-tuli, line 160: nit: rename `e`?

src/deepsparse/server/monitoring.py
```diff
@@ -0,0 +1,151 @@
...
+        ),
+        daemon=True,
+    )
+    proc.start()
+    return proc
+
+
+def _watch_file(config_path: str, endpoints_url: str, check_interval_s: float):
+    for _ in _diff_generator(config_path, endpoints_url, check_interval_s):
+        ...
```
rahul-tuli, line 61: nice!

rahul-tuli (approved):

src/deepsparse/server/monitoring.py
```diff
@@ -0,0 +1,151 @@
...
+            continue
+
+        try:
+            _update_endpoints(endpoints_url, old_config, new_config)
+        except requests.RequestException:
+            yield None
+            _LOGGER.error("Requests to server failed, not updating.", exc_info=1)
+            continue
+
+        path = config_path + f".v{version}"
```
KSGulin, line 110: Thoughts on saving previous versions of into a subdir next to the active config file, named "previous_configs" or something of that nature. Also, I think adding the date and time of modification to the config file name or as a markdown comment at the top of the config itself would be useful to users

corey-nm: Like both of those, will do that!

corey-nm: Added `<config path>.versions/` and a comment at the top of the file

KSGulin (dismissed):
LGTM 🙂

src/deepsparse/server/config.py
bfineran: +1 along with `old` and `new`

src/deepsparse/server/monitoring.py
```diff
@@ -0,0 +1,156 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
bfineran, line 1: let's rename this file to something more specific to the functionality given monitoring also has (more relevant) connotations to our logging functionality

bfineran (commented):
lgtm, two minor comments. like the use of proc + requesting into the server

KSGulin (approved):

PR #674, opened 2022-09-28
Reworking hot reload config tests

corey-nm:
The previous way of spinning up a separate server process had a couple problems:
1. it was flaky
2. it was slow
3. it didn't really work well on github actions

This PR removes that test and replaces it with:
1. Tests that the hot_config_reload flag of start_server properly calls start_config_watcher
2. Tests that the cli commands pass the proper flag to start_server

All of this is done with mocking and click CLIRunner, so its really fast

### DIFF
--- a/tests/server/test_hot_reloading.py
+++ b/tests/server/test_hot_reloading.py
@@ -13,15 +13,15 @@
 # limitations under the License.
 
 
-import multiprocessing
 import time
 from pathlib import Path
+from unittest import mock
 from unittest.mock import MagicMock, patch
 
-import requests
 import yaml
+from click.testing import CliRunner
 
-import pytest
+from deepsparse.server.cli import config, task
 from deepsparse.server.config import EndpointConfig, ImageSizesConfig, ServerConfig
 from deepsparse.server.config_hot_reloading import (
     _ContentMonitor,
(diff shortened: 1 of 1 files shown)


tests/server/test_hot_reloading.py
```diff
@@ -166,82 +164,48 @@ def test_file_monitoring(delete_mock, post_mock, tmp_path: Path):
...
-    assert resp.status_code == 404
+    start_server(cfg_path, hot_reload_config=False)
+    run_patch.assert_called_once()
+    watcher_patch.assert_not_called()
+
+    run_patch.reset_mock()
+    watcher_patch.reset_mock()
+    start_server(cfg_path, hot_reload_config=True)
+    run_patch.assert_called_once()
+    watcher_patch.assert_called_once()
```
bfineran, line 185: let's add in some inline comments for what's being tested

bfineran (commented):
quick comment

tests/server/test_hot_reloading.py
corey-nm: Good point, added comments and improved skimmability of them

rahul-tuli (approved):
To the moon!

KSGulin (approved):

PR #678, opened 2022-09-30
Setting ServerConfig.loggers default to None

corey-nm:
This makes it so prometheus server is only spun up when specified.

Testing: unit tests


dbogunowicz (approved):
:)

bfineran (approved):

PR #686, opened 2022-10-04
[ent] add is_enterprise option to build version

bfineran:


dhuangnm:
Should we also update src/deepsparse/version.py as well?

Also do we need to have the version string separate for deepsparse/deepsparse-nightly and deepsparse-ent although they're same for now but may be different in later releases?

bfineran:
> Should we update Makefile as well where BUILD_ARGS is used for nightly?
> 
> Do we need to have the version string separate for deepsparse/deepsparse-nightly and deepsparse-ent although they're same for now but may be different in later releases?

Discussed offline - we will merge in as is for now as the build args in the makefile are not used in setup anymore

corey-nm (approved):
lgtm

setup.py
```diff
@@ -34,14 +34,21 @@
...
 version_base = f"{version_major}.{version_minor}.0"
 
-_PACKAGE_NAME = "deepsparse" if is_release else "deepsparse-nightly"
+_PACKAGE_NAME = (
+    "deepsparse-ent"
+    if is_enterprise
+    else "deepsparse"
+    if is_release
+    else "deepsparse-nightly"
+)
```
rahul-tuli, line 51: nit: In my opinion, this is difficult to read, could we refactor into a function call maybe and use normal `if ... elif .. else` construct, won't block the PR on this though

rahul-tuli (changes requested):

rahul-tuli (approved):
Removed change request, cause the requested changes weren't blocking.

KSGulin (approved):

PR #698, opened 2022-10-06
[Data Logging] Basic implementation for logging `pipeline_inputs`

dbogunowicz:
Implementing the scaffolding for the data logging framework. The scope of this feature is marked in colour on the general architecture diagram:

image

Testing: 
A set of basic unit tests has been implemented

### DIFF
--- a/src/deepsparse/loggers/__init__.py
+++ b/src/deepsparse/loggers/__init__.py
@@ -15,4 +15,5 @@
 # flake8: noqa
 from .base_logger import *
 from .manager_logger import *
+from .monitoring import *
 from .prometheus_logger import *
--- a/src/deepsparse/loggers/monitoring/__init__.py
+++ b/src/deepsparse/loggers/monitoring/__init__.py
@@ -1,0 +1,13 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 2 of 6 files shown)


corey-nm:
this shows how the internals of estimators/data extraction might work, but how will the server or a user use these?

here's a couple questions:
- what do you want the configuration to look like if fully customized? i see some snippets in the image, but the proposed structure is not clear to me at least
- how does a user pass that to deepsparse server? or a pipeline if not using deepsparse server?
- what function calls do you want the server to make?
    - is the server just passing the output of run_with_monitoring to extract logging data? What does it do with the logging data after its returned? maybe there should be a class to handle all of that logic
    - what structure/function is owning the estimators?
- how does all this work for multiple pipelines? is configuration shared across pipelines or can each pipeline have its own custom metrics config?

So IMO the most important things to figure out first are how to integrate this with existing code (i.e. how server or pipeline code will change). the internals of new features can come after the higher level interface we want is more stable. additionally, focusing on the interplay will help us make the pieces more modular and easy to use

To kick the high level code interface off, i think my dream for metric dashboard would be:
```python
metrics: MetricsConfig = ...
dashboard = MetricsDashboard.from_config(metrics)

# each pipeline would receive its own dashboard object to log to
pipeline = Pipeline.create(..., dashboard=dashboard)

# pipeline is called like normal, and internally forwards data to dashboard
inputs = ...
outputs = pipeline(inputs)
```

the nice thing here is:
1. you can use it when not using the server
2. the diff to the server will be very small because we just have to add a new parameter to Pipeline.create and configuration objects for the metrics

under the hood MetricsDashboard might look like:

```python
class MetricsDashboard:
    logger: Logger # where to log the observed metrics to
    estimators: Dict[str, Estimator] # something mapping one of the passes/parts to what to use to estimate it

    def observe_engine_forward_time(self, ...):
        ...
    def observe_preprocessing_time(self, ...):
        ...
    ...
    def observe_pipeline_inputs(self, ...):
        ...
    def observe_pipeline_outputs(self, ...):
        ...
    def observe_engine_inputs(self, ...):
        ...
    def observe_engine_outputs(self, ...):
        ...
```

(in this example using the observer pattern https://en.wikipedia.org/wiki/Observer_pattern)

bfineran:
great questions @corey-nm, most will be addressed in the internal design doc for data logging. interesting point about having different metrics for different pipelines - this definitely needs to be taken into account for configs and isn't already.

Ideally, I'd like to not couple metrics collection with general pipeline behavior to keep pipelines focused on end to end inference. 
 Would rather have the server, metrics logging, and pipeline working separately but independently. Metrics collection should only require a pipeline and config for what to collect, should not need the server to run, however in the server scenario (our main target) the server will control both pipeline and metrics

PR #701, opened 2022-10-11
multi label support for text classification pipeline

bfineran:
adds support for multi label text classification inference with the following changes (matching transformers repo pipeline)
* adds a `top_k` arg for specifying number of labels to return
* adds logic to deprecate `return_all_scores`
* runs sigmoid instead of softmax when model config is set to `multi_label_classification` mode

**test_plan**
manual for now, further testing to be added in pipelines unit tests work. ran eval_downstream against existing qqp and sst2 models, verified expected number of outputs returns for various top_k values

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_classification.py
+++ b/src/deepsparse/transformers/pipelines/text_classification.py
@@ -42,6 +42,7 @@ from pydantic import BaseModel, Field
 from transformers.tokenization_utils_base import PaddingStrategy, TruncationStrategy
 
 from deepsparse import Pipeline
+from deepsparse.log import get_main_logger
 from deepsparse.transformers.pipelines import TransformersPipeline
 
 
@@ -50,6 +51,9 @@ __all__ = [
     "TextClassificationOutput",
     "TextClassificationPipeline",
 ]
+
+
+_LOGGER = get_main_logger()
 
 
 class TextClassificationInput(BaseModel):
@@ -86,7 +90,10 @@ class TextClassificationOutput(BaseModel):
 )
 class TextClassificationPipeline(TransformersPipeline):
     """
-    transformers text classification pipeline
+    transformers text classification pipeline.
+    Scores are returned as sigmoid over logits if one label is given or
+    the model config is set to "multi_label_classification". Softmax
+    returned otherwise
 
     example instantiation:
     ```python
@@ -132,7 +139,9 @@ class TextClassificationPipeline(TransformersPipeline):
         If a list of lengths is provided, then for each length, a model and
         tokenizer will be compiled capable of handling that sequence length
         (also known as a bucket). Default is 128
-    :param return_all_scores: if True, instead of returning the prediction as the
+    :param top_k: number of labels with the highest score to return. Default is 1
+    :param return_all_scores: [DEPRECATED] set top_k to desired value instead.
+        If True, instead of returning the prediction as the
         argmax of model class predictions, will return all scores and labels as
         a list for each result in the batch. Default is False
     """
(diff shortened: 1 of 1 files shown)


src/deepsparse/transformers/pipelines/text_classification.py
```diff
@@ -132,20 +139,31 @@ class TextClassificationPipeline(TransformersPipeline):
         If a list of lengths is provided, then for each length, a model and
         tokenizer will be compiled capable of handling that sequence length
         (also known as a bucket). Default is 128
-    :param return_all_scores: if True, instead of returning the prediction as the
+    :param top_k: number of labels with the highest score to return. Default is 1
+    :param return_all_scores: [DEPRECATED] set top_k to desired value instead.
```
dbogunowicz, line 143: How do we go about the deprecated argument? Can't we remove it in such a way, that the change is invisible to the user?

src/deepsparse/transformers/pipelines/text_classification.py
```diff
@@ -86,7 +90,10 @@ class TextClassificationOutput(BaseModel):
 )
 class TextClassificationPipeline(TransformersPipeline):
     """
-    transformers text classification pipeline
+    transformers text classification pipeline.
+    Scores are returned as sigmoid over logits if one label is given or
```
dbogunowicz, line 94: Why one label, what's the logic here?

src/deepsparse/transformers/pipelines/text_classification.py
```diff
@@ -267,3 +290,27 @@ def route_input_to_bucket(
...
+    if return_all_scores and top_k not in [1, num_labels]:
+        # top_k is not set to default, but does not match num_labels
+        # for return_all_scores
+        raise ValueError(
+            f"Mismatch: return_all_scores set to {return_all_scores} with "
+            f"{num_labels} in model config, but top_k set to {top_k}. "
+            "return_all_scores is deprecated, set top_k instead"
+        )
+
+    if top_k == 1 and return_all_scores:
```
KSGulin, line 308: It feels a bit counter-intuitive to say an argument is deprecated (v.s. going to be deprecated) while it affects the flow of logic. Do we have a standardized way for handling deprecation?

src/deepsparse/transformers/pipelines/text_classification.py
KSGulin, line 143: Under the `return_all_scores` `@property`, I'd recommend adding `warnings.warn('The use of return_all_scores is deprecated. Please use top_k instead ', DeprecationWarning)`

src/deepsparse/transformers/pipelines/text_classification.py
bfineran: generally we continue support the best we can while raising a warning that will remind users (internal and external) to update their usage

src/deepsparse/transformers/pipelines/text_classification.py
bfineran: we only want to warn for deprecation once - if return all scores is set, then the warning will happen at construction time

src/deepsparse/transformers/pipelines/text_classification.py
bfineran: sotmax would always return 1 for a single label problem. this additionally matches the transformers logic flow

src/deepsparse/transformers/pipelines/text_classification.py
bfineran: we want to keep it in for backwards compatibility with any existing flows for at least a release or two. transformers is doing the same with this specific argument

KSGulin (dismissed):

src/deepsparse/transformers/pipelines/text_classification.py
dbogunowicz: Oh, got it.

dbogunowicz (dismissed):

src/deepsparse/transformers/pipelines/text_classification.py
```diff
@@ -223,25 +241,30 @@ def process_engine_outputs(self, engine_outputs: List[numpy.ndarray]) -> BaseMod
...
-            # return all scores and labels for each item in batch
-            labels = [
-                [self.config.id2label[idx] for idx in range(scores.shape[1])]
-            ] * len(scores)
-            label_scores = [score.reshape(-1).tolist() for score in scores]
+            else:
+                score = score.reshape(-1).tolist()
+                ranked_idxs = sorted(
+                    range(len(score)), reverse=True, key=lambda idx: score[idx]
+                )
```
corey-nm, line 263: optional nit: In other parts of deepsparse this is done with
```python
(-score).argsort()[: self.top_k]
```

corey-nm (changes requested):
Request for different deprecation function

corey-nm (approved):

KSGulin (approved):

PR #703, opened 2022-10-12
[Tutorial] DeepSparse Server Monitoring via Prometheus/Grafana

dbogunowicz:
This is an example repository that goal is to quickly instruct the user how to couple DeepSparse Server with Prometheus/Grafana stack.

- [ ] Wait for DataLogging functionality, so we can showcase a larger array of logs in the dashboard
- [ ] Talk to the Product team on the aesthetics of the sample Dashboard


bfineran (changes requested):
tutorial looks great - overall theme of comments is making the message less from running this specific example (using the artifacts added in for convenience), but teaching the user the tools to adapt these steps to their own flows via those artifacts

tutorial/deepsparse_server_config.yaml
```diff
@@ -0,0 +1,10 @@
+num_cores: 2
+num_workers: 2
+loggers:
```
corey-nm, line 3: I guess this will need to be changed after our loggers refactor? Want to get this merged in before all that or after?

tutorial/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+```bash
+pip install virtualenv
+virtualenv deepsparse_venv
+
+source {$virtual_env_path}/bin/activate
+pip install deepsparse[server]
+```
+
+### Running the Instructions of this README
```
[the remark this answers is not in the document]
dbogunowicz: The script was very useful for me for bugging purposes and this is why I kept it. Feels good to have the contents of the tutorial condensed into one command. 
This script was written for the first time for the other tutorial (Server + K8 + Prometheus/Grafana) because this tutorial is more complex. However, in the context of this tutorial, I think I am fine with removing the bash script.

tutorial/deepsparse_server_config.yaml
dbogunowicz: Afaik, not at all. We have agreed that loggers remain a global dictionary of the config file:
image

PR #704, opened 2022-10-12
[YoloV5] Expose multi_label control for Pipeline

KSGulin:
This PR adds support for multi-label inference with Pipelines by exposing the `multi_label` kwargs

**Testing**
```python

images = ["basilica.jpg"]
pipeline = Pipeline(task="yolo")

outputs_implicit_sl = pipeline(images=images, iou_thres=0.6, conf_thres=0.001)[0]
outputs_explicit_sl = pipeline(images=images, iou_thres=0.6, conf_thres=0.001, multi_label=False)[0]
outputs_ml = pipeline(images=images, iou_thres=0.6, conf_thres=0.001, multi_label=True)[0]

assert all(out1 == out2 for out1, out2 in zip(outputs_implicit_sl, outputs_explicit_sl))
assert any(out1 != out2 for out1, out2 in zip(outputs_explicit_sl, outputs_ml))
```

### DIFF
--- a/src/deepsparse/yolo/pipelines.py
+++ b/src/deepsparse/yolo/pipelines.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import json
-from typing import Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import numpy
 import onnx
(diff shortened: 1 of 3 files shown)


bfineran (commented):
what's the reason for choosing at the input level rather than constructor?

KSGulin:
@bfineran good question. I see `multi_label` as being in the same family of parameters as `iou_thres` and `conf_thres` as its value doesn't impact the engine (unlike `image_size`) and so the user should be able to provide different values between calls. Do you see it differently?

bfineran:
we had discussed in the past both giving the user the option to set the defaults via constructor as well as override for a particular inference, I think this is fine for now, but potentially will add options to the constructor in the future

bfineran (approved):

anmarques (approved):
LGTM

PR #708, opened 2022-10-14
Adding example for custom pipeline and dynamic import

corey-nm:
Could use some input on best thing to include in readme


mgoin (commented):
@corey-nm nice! can you add an example test like `deepsparse/tests/examples/test_twitter_nlp.py`

PR #709, opened 2022-10-17
Adding backwards compatibility for server cli

corey-nm:
This adds support for using the old version of deepsparse.server cli with a deprecation message:

```
deepsparse.server --task qa
deepsparse.server task qa
deepsparse.server --config_file /some/path
deepsparse.server config /some/path
```

Test plan:
- Since this is not replacing the new way of the api, none of the QA or unit tests need to be updated
- Unit tests
- Manual cli testing

### DIFF
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -164,17 +164,17 @@ def main(
     """
     Start a DeepSparse inference server for serving the models and pipelines.
 
-        1. `deepsparse.server config [OPTIONS] <config path>`
+        1. `deepsparse.server --config_file <config path> [OPTIONS]`
 
-        2. `deepsparse.server task [OPTIONS] <task>
+        2. `deepsparse.server --task <task> [OPTIONS]`
 
     Examples for using the server:
 
-        `deepsparse.server config server-config.yaml`
+        `deepsparse.server --config_file server-config.yaml`
 
-        `deepsparse.server task question_answering --batch-size 2`
+        `deepsparse.server --task question_answering --batch-size 2`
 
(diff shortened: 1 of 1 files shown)


src/deepsparse/server/cli.py
```diff
@@ -166,30 +106,185 @@ def config(
...
+    log_level: str,
+    hot_reload_config: bool,
+    model_path: str,
+    batch_size: int,
+    num_cores: int,
+    num_workers: int,
+    integration: str,
+):
+    """
+    Start a DeepSparse inference server for serving the models and pipelines.
```
bfineran, line 165: let's update help text to include the pathway w/o the named args

bfineran (commented):
LGTM overall, as a follow up let's get ahead on identifying readmes that use the `task`/`config` pathways, there shouldn't be too many

src/deepsparse/server/cli.py
corey-nm: Updated

bfineran (approved):

src/deepsparse/server/cli.py
```diff
@@ -166,30 +106,185 @@ def config(
...
+)
+@HOST_OPTION
+@PORT_OPTION
+@LOG_LEVEL_OPTION
+@HOT_RELOAD_OPTION
+@MODEL_OPTION
+@BATCH_OPTION
+@CORES_OPTION
+@WORKERS_OPTION
+@INTEGRATION_OPTION
```
rahul-tuli, line 148: nice

rahul-tuli (approved):

PR #713, opened 2022-10-20
[license] support for non-throwing license_status call

bfineran:
changes:
* adds python API wrapper for `license_status`
* updates `validate_license` to call into `license_status` as source of truth
* warning messages no longer displayed on first call to validate for `deepsparse.license`

**test_plan**
manual from QA after engine build/cherry-pick

### DIFF
--- a/src/deepsparse/license.py
+++ b/src/deepsparse/license.py
@@ -37,7 +37,7 @@ import shutil
 import sys
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Optional
+from typing import Optional, Tuple
 
 import click
 
@@ -62,7 +62,9 @@ def add_deepsparse_license(token_or_path):
 
     # do not print splash message on successful validation
     # only print on re-validation after license file copy
-    validate_license(candidate_license_file_path, print_splash=False)
+    validate_license(
+        candidate_license_file_path, print_splash=False, print_warning=False
+    )
     _LOGGER.info("DeepSparse license successfully validated")
 
     # copy candidate file to {LICENSE_FILE} in same directory as NM engine binaries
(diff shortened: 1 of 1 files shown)


src/deepsparse/license.py
```diff
@@ -74,7 +75,31 @@ def add_deepsparse_license(token_or_path):
...
+    Returns the status of a license as tuple of 
+    (is_exception, error/warning, splash message)
+    
+    License should be passed as a text file containing only the JWT. If no path is
+    provided the expected file path of the token will be validated. Default
+    path is ~/.config/neuralmagic/license.txt
+
+    :param license_path: file path to text file of token to check status of.
+        Default is None, expected token path will be validated
+    :return: tuple of (is_exception, error/warning, splash message)
```
bnellnm, line 89: fyi, if is_exception is true then the splash will have the same value as the error message.   there's no deep reason for this but it's just for convenience because of how the c++ code is written.

i don't think it should matter but thought i would mention it.

bfineran: updated - so only warnings go in the second slot?

src/deepsparse/license.py
```diff
@@ -85,22 +110,20 @@ def validate_license(license_path: Optional[str] = None, print_splash: bool = Tr
...
         successful validation. Default True
+    :param print_splash: set False to not print warning message on
+        successful validation with a returned warning. Default True
     """
+    is_exception, error_message, splash_message = license_status(license_path)
 
-    deepsparse_lib = init_deepsparse_lib()
+    if is_exception:
+        # exception would be raised on compilation with license_path, raise
+        raise ValueError(splash_message)
```
bnellnm: sorry, this was right the original way.  it should always be `error_message`

bnellnm (dismissed):
LGTM

rahul-tuli (approved):

bnellnm (approved):

bnellnm (approved):

PR #715, opened 2022-10-21
[Data Logging] BaseLogger + PythonLogger

dbogunowicz:
This PR implements:
- the new BaseLogger class 
 -the PythonLogger class (inherits from BaseLogger)
- the changes to the pipeline: removes `run_with_monitoring` and instead allows the pipeline to accept a `BaseLogger` as an attribute and collect inference information by the logger on the `__call__()`

Example usage:

```python
from deepsparse import PythonLogger, Pipeline

python_logger = PythonLogger()
pipeline = Pipeline.create("token_classification", batch_size=1, alias="python_logger", logger=python_logger)
pipeline("all_your_base_are_belong_to_us")
```
results in:

```bash
24/10/2022 05:43:50 Identifier: python_logger.pipeline_inputs | Category: data | Logged Data: inputs='hehe' is_split_into_words=False
24/10/2022 05:43:50 Identifier: python_logger.engine_inputs | Category: data | Logged Data: [array([[ 101, 2002, 5369,  102,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0,    0,    0,    0,    0,
           0,    0,    0,    0,    0,    0,    0]]), array([[1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]), array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]])]
24/10/2022 05:43:50 Identifier: python_logger.pre_process | Category: system | Logged Data: 0.002111834939569235
24/10/2022 05:43:51 Identifier: python_logger.engine_outputs | Category: data | Logged Data: [array([[[ 7.00468   , -1.2516016 , -3.2124548 , ..., -3.225307  ,
         -2.0040963 , -2.2855673 ],
        [-0.1511735 ,  3.6688154 , -2.6576076 , ..., -2.8445685 ,
         -0.13903134, -2.276656  ],
        [ 2.2755554 ,  1.1635886 , -0.13754019, ..., -2.1952467 ,
         -2.3797224 , -1.0511494 ],
        ...,
        [ 1.9636225 ,  2.2182174 , -2.1439798 , ..., -2.553689  ,
         -1.2473056 , -2.4919841 ],
        [ 4.231438  ,  0.2521807 , -2.6963062 , ..., -2.8192189 ,
         -1.7863538 , -2.4372022 ],
        [ 2.828379  ,  1.5764933 , -2.3594143 , ..., -2.6540926 ,
         -1.4270239 , -2.6137607 ]]], dtype=float32)]
24/10/2022 05:43:51 Identifier: python_logger.engine_forward | Category: system | Logged Data: 0.3400956797413528
24/10/2022 05:43:51 Identifier: python_logger.pipeline_outputs | Category: data | Logged Data: predictions=[[TokenClassificationResult(entity='LABEL_1', score=0.548909604549408, index=1, word='he', start=0, end=2, is_grouped=False), TokenClassificationResult(entity='LABEL_0', score=0.5486404895782471, index=2, word='##he', start=2, end=4, is_grouped=False)]]
24/10/2022 05:43:51 Identifier: python_logger.post_process | Category: system | Logged Data: 0.0003222692757844925
24/10/2022 05:43:51 Identifier: python_logger.total_inference | Category: system | Logged Data : 0.34388848999515176
```


src/deepsparse/loggers/base_logger.py
```diff
@@ -13,10 +16,24 @@
...
 
 
-__all__ = ["BaseLogger"]
+__all__ = ["BaseLogger", "MetricCategories"]
+
+
+class MetricCategories(Enum):
+    """
+    Metric Taxonomy [for reference]
+        CATEGORY - category of metric (System/Latency/Data)
```
dbogunowicz, line 29:
```suggestion
        CATEGORY - category of metric (System/Performance/Data)
```

src/deepsparse/loggers/base_logger.py
```diff
@@ -25,40 +42,21 @@ class BaseLogger(ABC):
...
+        the monitoring service
+
+        :param identifier: The identifier of the log
+             By convention should have the following structure:
+             {pipeline_name}.{target_name}.{optional_identifier_1}.{optional_identifier_2}.{...}
+        :param value: The data structure that is logged
+        :param category: The metric category that the log belongs to
         """
-        raise NotImplementedError()
+        raise NotImplementedError
```
dbogunowicz, line 62:
```suggestion
        raise NotImplementedError()
```

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,52 @@
...
+"""
+Implementation of the Python Logger that logs to the stdout
+"""
+import logging
+from typing import Any
+
+from deepsparse.loggers import BaseLogger, MetricCategories
+
+
+logging.basicConfig(format="%(asctime)s %(message)s", datefmt="%m/%d/%Y %I:%M:%S %p")
```
dbogunowicz, line 24: I feel that PythonLogger should have a bit of configurability when it comes to the stdout. Maybe having a method to setup the `logging.basicConfig(...)` from `config.yaml`? Something like
```python
def __init__(self, config: Optional[Dict] = None):
    if config:
        self.setup_built_in_logger(config)
        
def setup_build_in_logger(self, config):
    # do something
    args = ...
    logging.basicConfig(args)

File: src/deepsparse/pipeline.py [hunk 4]
```diff
@@ -129,6 +130,8 @@ class PipelineImplementation(Pipeline):
         synchronous execution - if running in dynamic batch mode a default
         ThreadPoolExecutor with default workers equal to the number of available
         cores / 2
+    :param logger: An optional BaseLogger object. If provided, it will be logging
+        the appropriate data (according to logger's internal state) on __call__()
```
dbogunowicz, line 134:
```suggestion
        the appropriate information (according to logger's internal state) from the pipeline on __call__()
```

src/deepsparse/pipeline.py
```diff
@@ -185,10 +198,19 @@ def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
             )
         timer = TimingBuilder()
 
-        # parse inputs into input_schema
         timer.start(InferencePhases.TOTAL_INFERENCE)
+
+        # ------ PREPROCESSING ------
```
dbogunowicz, line 203: Personally found it useful to visually separate the __call__() into three parts: preprocessing, inference and postprocessing.

src/deepsparse/pipeline.py
```diff
@@ -211,6 +240,13 @@ def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
...
         # join together the batches of size `self._batch_size`
         engine_outputs = self.join_engine_outputs(batch_outputs)
+        if self.logger:
+            self.logger.log(
+                identifier=f"{self._alias}.engine_outputs",
+                value=engine_outputs,
+                category=self.logger.metric_categories.DATA,
+            )
+
         timer.stop(InferencePhases.ENGINE_FORWARD)
```
dbogunowicz, line 251:
```suggestion

# ------ PREPROCESSING ------
```

src/deepsparse/pipeline.py
```diff
@@ -222,39 +258,23 @@ def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
...
-        if not monitoring:
-            return pipeline_outputs
-
-        else:
-            inference_timing = InferenceTimingSchema(**timer.build())
-            return pipeline_outputs, pipeline_inputs, engine_inputs, inference_timing
+        if self.logger:
+            self.logger.log(
+                identifier=f"{self._alias}.inference_timing",
+                value=timer.build(),
```
dbogunowicz, line 273: Instead of passing the result of `timer.build(): dict` to `InferenceTimingSchema` (like we did before), we can pass the dict to the logger and then further alter it inside the logger, not the pipeline.

src/deepsparse/pipeline.py
```diff
@@ -211,8 +240,16 @@ def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
...
+        if self.logger:
+            self.logger.log(
+                identifier=f"{self._alias}.engine_outputs",
+                value=engine_outputs,
+                category=self.logger.metric_categories.DATA,
+            )
+
         timer.stop(InferencePhases.ENGINE_FORWARD)
 
+# ------ POSTPROCESSING ------
```
dbogunowicz, line 252:
```suggestion
        # ------ POSTPROCESSING ------
```

src/deepsparse/pipeline.py
```diff
@@ -142,13 +145,23 @@ def __init__(
...
     ):
         self._model_path_orig = model_path
         self._model_path = model_path
         self._engine_type = engine_type
         self._batch_size = batch_size
         self._alias = alias
         self.context = context
+        self.logger = logger
+
+        if self.logger and not self._alias:
```
dbogunowicz, line 158: My philosophy here is: is somebody wants to enable logging, they should be explicit in the naming of their pipelines. Feel free to challenge this opinion.

src/deepsparse/loggers/base_logger.py
```diff
@@ -25,40 +42,21 @@ class BaseLogger(ABC):
...
-    ):
-        """
-        Logs the inference latency information to the appropriate monitoring service
-        :param pipeline_name: The name of the inference pipeline from which the
-            logger consumes the inference information to be monitored
-        :param inference_timing: pydantic model that holds the information about
-            the inference latency of a forward pass
-        """
-        raise NotImplementedError()
+    def __init__(self, config: Optional[dict] = None):
```
bfineran, line 45: let's remove config from the base class, if the base class isn't doing anything with it, then we shouldn't keep it at this level

src/deepsparse/loggers/python_logger.py
bfineran: Disagreed here for now, let's keep things simple and add in features as needed rather than tying to functionality that may or may not be used (and need to make a breaking change on later)

src/deepsparse/pipeline.py
bfineran: Disagreed here, let's just default to using the task name - the server will always handle alias and any other user using a logger shouldn't have to worry about setting one unless dealing with multiple models

src/deepsparse/pipeline.py
bfineran: we should be getting rid of inference timing schema now - this is because we want to be logging the timings in real time as they happen in case there is an error

bfineran (changes requested):
took a quick first pass - additionally, let's get the actual calls to the logger moved to a separate PR, there's further discussion we should do here that shouldn't block this base PR

src/deepsparse/loggers/python_logger.py
dbogunowicz: Ok, not having strong feelings about this, was just a loose proposal

src/deepsparse/pipeline.py
dbogunowicz: Yeah, makes sense.

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,52 @@
...
+        Collect information from the pipeline and print them to the console
+
+        :param identifier: The identifier of the log
+             By convention should have the following structure:
+             {pipeline_name}.{target_name}.{optional_identifier_1}.{optional_identifier_2}.{...}
+        :param value: The data structure that the logger is logging
+        :param category: The metric category that the log belongs to
+        """
+        msg = (
+            f"Identifier: {identifier} | Category: {category.value} "
```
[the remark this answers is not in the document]
dbogunowicz: Given that the goal of the `PythonLogger` is to log to stdout, does it make sense to print those huge data structures?

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,52 @@
...
+            f"| Logged Data Type: {type(value)}"
+        )
+        logging.info(msg=msg)
```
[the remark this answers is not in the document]
dbogunowicz: why not?

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,52 @@
...
+                e.g. identifier = "pipeline_name"
+            If identifier is to be more complex, optional strings are to be concatenated
+            using dot as a separator:
+                e.g. identifier = "pipeline_name.some_argument_1.some_argument_2"
+        :param value: The data structure that the logger is logging
+        :param category: The metric category that the log belongs to
+        """
+        msg = (
+            f" Identifier: {identifier} | Category: {category.value} "
+            f"| Logged Data Type: {value}"
```
dbogunowicz, line 50:
```suggestion
            f"| Logged Data: {value}"
```

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,52 @@
...
+__all__ = ["PythonLogger"]
+
+
+class PythonLogger(BaseLogger):
+    """
+    Python logger that writes the collected logs to stdout
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
dbogunowicz, line 33:
```suggestion
```

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,52 @@
...
+
+__all__ = ["PythonLogger"]
+
+
+class PythonLogger(BaseLogger):
+    """
+    Python logger that writes the collected logs to stdout
+    """
+
+    def __init__(self, **kwargs):
```
dbogunowicz, line 32:
```suggestion
    def __init__(self, **kwargs):
```

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,51 @@
...
```
dbogunowicz, line 32:
```suggestion

```

src/deepsparse/pipeline.py
```diff
@@ -142,13 +145,15 @@ def __init__(
...
         executor: Optional[Union[ThreadPoolExecutor, int]] = None,
+        logger: Optional[BaseLogger] = None,
     ):
         self._model_path_orig = model_path
         self._model_path = model_path
         self._engine_type = engine_type
         self._batch_size = batch_size
         self._alias = alias
         self.context = context
+        self.logger = logger
```
corey-nm, line 156: We could add a `class NullLogger(BaseLogger)` to set self.logger to here so we don't have to wrap everything with `if self.logger`:

```python
self.logger = logger if logger is not None else NullLogger()
```

src/deepsparse/pipeline.py
```diff
@@ -666,6 +692,20 @@ def to_config(self) -> "PipelineConfig":
             kwargs=kwargs,
         )
 
+    def log(self, identifier: str, value: Any, category: str):
```
corey-nm, line 695: maybe private method?

src/deepsparse/loggers/python_logger.py
```diff
@@ -0,0 +1,49 @@
...
+from datetime import datetime
+from typing import Any
+
+from deepsparse.loggers import BaseLogger, MetricCategories
+
+
+__all__ = ["PythonLogger"]
+
+
+class PythonLogger(BaseLogger):
```
corey-nm, line 27: I still think `StdoutLogger` would be more clear as to what this class does (PythonLogger also makes it sound like its using the logging package), but I know we have PythonLogger in sparseml, so consistency is probably better? 🤷

corey-nm (approved):
I think this looks really good, nice!

src/deepsparse/pipeline.py
dbogunowicz: I was pondering whether to go down that road. Feels right to make it private, but then felt a bit odd given that so many other functions in `pipeline`, that are practically private, do not start with underscore. So I opted for consistency here.

src/deepsparse/pipeline.py
dbogunowicz: Instead of adding another `Logger`, I'd rather just simply push the `if`-statement the `pipeline.log()` function. Same result, less coding.

src/deepsparse/pipeline.py
bfineran: subclasses should avoid not calling into private methods when possible, especially if outside the parent module scope, don't see any further reason to hide it as well

PR #728, opened 2022-10-31
[Data Logging] Prometheus Logger

dbogunowicz:
Edit re-introduces the `PrometheusLogger` object to our repository and incorporates it into the set of valid leaf loggers of the DeepSparse server.


src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -0,0 +1,116 @@
...
+        Collect information from the pipeline and pipe it them to the stdout
+
+        :param identifier: The name of the thing that is being logged.
+        :param value: The data structure that the logger is logging
+        :param category: The metric category that the log belongs to
+        """
+
+        # needs to adhere to
+        # https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+        formatted_identifier = identifier.replace(".", "__")
```
dbogunowicz, line 79: I am not paying too much focus to the best practices of Prometheus logger, but will need to look into it later: https://prometheus.io/docs/practices/naming/

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -0,0 +1,116 @@
...
+                formatted_identifier, category
+            )
+        prometheus_metric.observe(self._validate(value))
+        self._try_export_metrics_to_textfile()
+
+    def _add_metric_to_registry(self, identifier: str, category: str) -> Summary:
+        description = (
+            """Summary metric for identifier: {identifier} | Category: {category}"""
+        )
+        prometheus_metric = Summary(
```
dbogunowicz, line 92: The second small task to take a look at once we get to the end of the engineering work is maybe using so-called labels/exemplars to make logs a bit nicer and structured:
```python
from prometheus_client import Histogram
h = Histogram('request_latency_seconds', 'Description of histogram')
h.observe(4.7, {'trace_id': 'abc123'})
```
(https://github.com/prometheus/client_python)

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _setup_client(self):
+        # starts the Prometheus client
+        start_http_server(port=self.port)
+        _LOGGER.info(f"Prometheus client: started. Using port: {self.port}.")
+
+    def _validate(self, value: Any) -> Any:
+        # make sure we are passing a value that can be
+        # logged by prometheus client
+        if not any(isinstance(value, datatype) for datatype in SUPPORTED_DATA_TYPES):
```
dbogunowicz, line 114: This is also something I will revisit later

src/deepsparse/loggers/prometheus_logger.py
bfineran: also need to replace any whitespaces. why double underscore?

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        prometheus_metric = Summary(
+            identifier,
+            description.format(identifier=identifier, category=category),
+            registry=REGISTRY,
+        )
+        self._prometheus_metrics[identifier] = prometheus_metric
+        return prometheus_metric
+
+    def _try_export_metrics_to_textfile(self):
```
bfineran, line 100: nit: `try` is a bit misleading - `_update_log_counter` maybe

src/deepsparse/loggers/prometheus_logger.py
bfineran: 👍

bfineran, line 114: ack revising, but fyi 
```suggestion
        if not isinstance(value, SUPPORTED_DATA_TYPES):  # SUPPORTED_DATA_TYPES must be tuple
```

bfineran (changes requested):
lgtm overall, few nits, and needs rebase

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: Our agreed-upon structure of an identifier is:
`<some_string_1>.<some_string_2>.<some_string_3>...`
However, as stated here: https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

```
The metric name specifies the general feature of a system that is measured (e.g. http_requests_total - the total number of HTTP requests received). It may contain ASCII letters and digits, as well as underscores and colons. It must match the regex [a-zA-Z_:][a-zA-Z0-9_:]*.
```
So the non-letter chars that may be used by Prometheus are `:` and `_` right? Well not exactly:
```
The colons are reserved for user-defined recording rules. They should not be used by exporters or direct instrumentation.
```

So we are left only with the underscores.

Usually, a Prometheus metric looks like this: 
`total_time_elapsed_seconds`

But the identifiers can look like `pipeline_input.something1.something2`. If we just replaced all the dots with underscores we end up with `pipeline_inputs_something1_something2` - we lose the tight relationship (e.g. when filtering) between the string `pipeline` and  string `inputs`. So I decided that the least problematic formatting would be `pipeline_inputs__something1__something2`. It's not great, but it's clear enough.

Note: `Label names beginning with __ are reserved for internal use.`

bfineran: yup, makes sense - only concern would be downstream users potentially missing the double

bfineran (approved):

PR #746, opened 2022-11-04
text classification - flatten trivial nested list inputs

bfineran:
the text classification pipeline accepts a union of lists of lists, lists, and strings as input - for models with single items (ie sentiment analysis on sst2) the lists of lists input will cause an error with the huggingface tokenizer
`TypeError: TextEncodeInput must be Union[TextInputSequence, Tuple[InputSequence, InputSequence]]`
which will give an internal server error response in the deepsparse server.

This PR addresses this specific case by flattening text classification inputs that are lists of lists where all inner lists have length 1 as these inputs are not compatible with tokenizers.

`**test_plan:**`
error reproduced from bug report - confirmed the error no longer occurs after this patch


mgoin (approved):
thanks!

corey-nm (approved):

PR #748, opened 2022-11-07
[Data Logging] `build_logger` function

dbogunowicz:
Specifies `build_logger` functions that maps from `server_config: ServerConfig` to `deepsparse_logger: BaseLogger`.
Assumes server configuration in the following form:

```yaml
loggers:
    - some_logger_1
    - some_logger_2
       key1: value1
       key2: value2
endpoints:
    - task: question_answering
      route: /unpruned/predict
      model: zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none
      batch_size: 1
      data_logging:
        - target: pipeline_outputs
          functions:
           - func: builtins:some_function1
             frequency: 2
           - func: .../user_defined_metric_functions.py:some_function2
             frequency: 3
        - target: pipeline_outputs
           functions:
              ...
    - task: question_answering
      route: /unpruned/predict
      model: zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none
      batch_size: 1
      data_logging:
        - target: engine_outputs
          functions:
           - func: np.mean
             frequency: 4
```


src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+        "the context of the DeepSparse server. "
+        "If not None, this configuration stops logging data "
+        "to globally specified loggers, and will only use "
+        "the subset of loggers "
+        "(specified by a list of their names)",
+    )
+
+    def __init__(self, **data):
+        if data.get("function"):
+            raise ValueError()
```
dbogunowicz, line 103: Looking for a smarter way to disable the user from manually specifying the argument (it should be parsed out from the `func` argument).

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,157 @@
...
+    return list_target_logging_cfgs
+
+
+def _get_metric_function_configs(
+    leaf_logger_name: str, metric_function_cfgs: List[MetricFunctionConfig]
+) -> List[MetricFunctionConfig]:
+    list_metric_function_cfgs = []
+    # iterate over all metric function configs
+    for metric_function_cfg in metric_function_cfgs:
+        if metric_function_cfg.logger is not None:
```
dbogunowicz, line 131: Leaving here a possibility for the user to specify .e.g:
```yaml
- target: engine_outputs
  functions:
   - func: np.mean
     frequency: 5
     logger: ["python", "prometheus"]```

File: src/deepsparse/loggers/function_logger.py [hunk 3]
```diff
@@ -31,49 +33,70 @@ class FunctionLogger(BaseLogger):
...
     """
 
     def __init__(
         self,
         logger: BaseLogger,
-        target_identifier: str,
-        function: Callable[[Any], Any],
-        function_name: str = None,
-        frequency: int = 1,
+        target_logging_configs: List["TargetLoggingConfig"],  # noqa F821
```
corey-nm, line 42: Was function logger in previous form not easy to use?

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+    if function_identifier.startswith("builtins:"):
+        func_name = function_identifier.split(":")[1]
+        return getattr(built_ins, func_name), func_name
+
+    # assume a dynamic import function of the form
+    # '<path_to_the_python_script>:<function_name>'
+    path, func_name = function_identifier.split(":")
+    spec = importlib.util.spec_from_file_location("user_defined_metric_functions", path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
```
corey-nm, line 74: This is cool, never seen this before +1

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+        func_name = function_identifier.split(".")[1]
+        return getattr(torch, func_name), func_name
+
+    if function_identifier.startswith("numpy.") or function_identifier.startswith(
+        "np."
+    ):
+        func_name = function_identifier.split(".")[1]
+        return getattr(numpy, func_name), func_name
+
+    if function_identifier.startswith("builtins:"):
```
corey-nm, line 65: thoughts on `builtins.` instead of the colon?

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+import torch
+
+
+__all__ = [
+    "MetricFunctionConfig",
+    "TargetLoggingConfig",
+]
+
+
+def get_function_and_function_name(
```
corey-nm, line 34: I like this function +1 🚀

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+    def non_zero_frequency(cls, frequency):
+        if frequency <= 0:
+            raise ValueError(
+                f"Passed frequency: {frequency}, but "
+                "frequency must be a positive integer greater equal 1"
+            )
+        return frequency
+
+
+class TargetLoggingConfig(BaseModel):
```
bfineran, line 123: why do we need a separate class for this rather than having a target field in the metric function config and storing a list of those?

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+    return getattr(module, func_name), func_name
+
+
+class MetricFunctionConfig(BaseModel):
+    """
+    Holds logging configuration for a metric function
+    """
+
+    func: str = Field(description="Identifier of the metric function")
+    function: Callable[[Any], Any] = Field(description="The metric function callable")
```
corey-nm, line 84: What's the reasoning for having both of these again? Is that what the value error below is for? Will not passing in these configs to FunctionLogger address this?

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+    def __init__(self, **data):
+        if data.get("function"):
+            raise ValueError()
+        if data.get("function_name"):
+            raise ValueError()
+        function_identifier = data.get("func")
+        # automatically extract function and function name
+        # from the function_identifier
+        function, function_name = get_function_and_function_name(function_identifier)
+        data["function"], data["function_name"] = function, function_name
```
corey-nm, line 110: This is parsing code - technically different from configuration, could go into a different place in the code. Just noting here while reviewing

src/deepsparse/loggers/config.py
```diff
@@ -0,0 +1,151 @@
...
+    Holds configuration for a single data logging target
+    """
+
+    target: str = Field(description="Name of the target.")
+    functions: List[MetricFunctionConfig] = Field(
+        description="List of MetricFunctionConfigs pertaining to the target"
+    )
+
+    @root_validator
+    def unique_mapping_names_per_endpoint(cls, data: Dict[str, Any]) -> Dict[str, Any]:
```
corey-nm, line 134: nice!

src/deepsparse/server/config.py
```diff
@@ -144,16 +149,36 @@ class ServerConfig(BaseModel):
 
     endpoints: List[EndpointConfig] = Field(description="The models to serve.")
 
-    loggers: Union[Dict[str, Dict[str, Any]], str, None] = Field(
+    loggers: Optional[List[Union[str, Dict[str, Dict[str, Any]]]]] = Field(
```
corey-nm, line 152:
```suggestion
    loggers: Dict[str, Dict[str, Any]]
```

src/deepsparse/loggers/function_logger.py
```diff
@@ -31,49 +33,70 @@ class FunctionLogger(BaseLogger):
...
         """
+        Collect information from the pipeline and:
+        1) Filter it according to the `self.target_logging_configs`
+        2) Apply metric functions according to the `self.target_logging_configs`
+
         :param identifier: The name of the item that is being logged.
         :param value: The data structure that the logger is logging
         :param category: The metric category that the log belongs to
         """
-        extracted_value = match_and_extract(
```
bfineran, line 67: let's move this to a helper function, `_match_and_log` or something

src/deepsparse/loggers/function_logger.py
```diff
@@ -31,49 +33,70 @@ class FunctionLogger(BaseLogger):
...
+            target_identifier = target_cfg.target
+            functions_cfgs = target_cfg.functions
+
+            for function_cfg in functions_cfgs:
+
+                frequency = function_cfg.frequency
+                function = function_cfg.function
+                function_name = function_cfg.function_name
+
+                extracted_value = match_and_extract(
```
bfineran, line 70: why do we need to match every time for the same identifier?

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,157 @@
...
+
+    loggers_config = server_config.loggers
+    if loggers_config is None:
+        return loggers_config
+
+    leaf_loggers = build_leaf_loggers(loggers_config)
+    loggers = []
+
+    for name, leaf_logger in leaf_loggers.items():
+        target_logging_configs = _get_target_logging_configs(name, server_config)
```
corey-nm, line 56: These two seem inverted to me, thoughts on doing the loop like:

```
for endpoint in server_cfg.endpoints:
    for cfg in endpoint.data_logging:
        for target in ...:
```
?

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,157 @@
...
+
+        list_metric_function_cfgs.append(metric_function_cfg)
+    return list_metric_function_cfgs
+
+
+def _build_single_logger(
+    logger_name: str, logger_arguments: Optional[Dict[str, Any]] = None
+) -> Union[logger_objects.PythonLogger]:
+    if logger_name == "python":
+        if logger_arguments:
```
bfineran, line 145: no need for this, let the python class crash on its own

src/deepsparse/server/config.py
dbogunowicz: Ok, this makes sense, I also like the simplification, but why removing the `Optional`?

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            logger = leaf_logger
+
+        loggers.append(logger)
+
+    return logger_objects.MultiLogger(loggers) if len(loggers) > 1 else loggers[0]
+
+
+def build_leaf_loggers(
+    loggers_config: List[Union[str, Dict[str, Dict[str, Any]]]]
```
bfineran, line 71: I think the typing is a bit off here - `List[Dict[str, ...]]` wouldn't make much sense if the dict keys are logger names, that case instead should be just a dictionary where each key is a logger class and the args are their kwargs

src/deepsparse/loggers/config.py
dbogunowicz: Good call, I like that

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,157 @@
...
+    list_target_logging_cfgs = []
+    # iterate over all endpoints
+    for endpoint in server_config.endpoints:
+        if endpoint.data_logging is None:
+            continue
+        # if endpoint has data logging information,
+        # iterate over endpoint's target logging config
+        for target_logging_cfg in endpoint.data_logging:
+            # new target name is a composition of the endpoint name and target name
+            target_identifier = f"{endpoint.name}.{target_logging_cfg.target}"
```
bfineran, line 111: should move to a helper function so we can also account for regex

src/deepsparse/server/config.py
corey-nm: I was thinking empty dict is equivalent to `None`, and it would make typing a bit easier. Fine with keeping optional though

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,157 @@
...
+            # get metric logging config corresponding to the leaf logger
+            metric_function_configs = _get_metric_function_configs(
+                leaf_logger_name, target_logging_cfg.functions
+            )
+            # append to the result
+            list_target_logging_cfgs.append(
+                TargetLoggingConfig(
+                    target=target_identifier, functions=metric_function_configs
+                )
```
bfineran, line 120: again, functionality here would be simplified and flow a bit clearer with the target moved into the metric schema

src/deepsparse/server/config.py
```diff
@@ -144,16 +149,36 @@ class ServerConfig(BaseModel):
...
             "Optional dictionary of logger integration names to initialization kwargs."
-            " Set to 'default' for default logger based on deployment. Set to None for"
-            " no loggers. Default is `None`. Example: "
-            "{'prometheus': {'port': 8001}}."
+            "Set to None for no loggers. Default is `None`."
         ),
     )
 
+    @validator("endpoints")
+    def set_unique_endpoint_names(
```
bfineran, line 161: why are we adding this here - seems orthogonal to the purpose of this PR, additionally don't we already catch this as a user error?

bfineran (changes requested):
great start - core functionality looks good, let's focus on getting this cleaned up. starting with the metric schema and work through simplifying the helpers. we should be able to avoid a lot of the nesting

src/deepsparse/server/config.py
dbogunowicz: I agree and disagree:
1. Yes, it is orthogonal to this PR. It makes more sense to add it to this PR's successor.
2. Currently are catching the error yes, but only after the loggers have been created. This check should happen before the loggers are created, and IMO the best place for it is on the instantiation of the BaseModel.

I will remove this feature for now, but will revisit in the next PR.

src/deepsparse/loggers/config.py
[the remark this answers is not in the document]
dbogunowicz: Good call.

src/deepsparse/loggers/function_logger.py
```diff
@@ -31,49 +33,70 @@ class FunctionLogger(BaseLogger):
...
-                    value=mapped_value,
-                    category=category,
+
+        for target_cfg in self.target_logging_configs:
+            target_identifier = target_cfg.target
+            functions_cfgs = target_cfg.functions
+
+            for function_cfg in functions_cfgs:
+
+                frequency = function_cfg.frequency
```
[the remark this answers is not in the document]
dbogunowicz: Just curious why not? For instance, if it serves readability?

src/deepsparse/loggers/function_logger.py
dbogunowicz: Please elaborate.

src/deepsparse/server/build_logger.py
dbogunowicz: Could you elaborate?

src/deepsparse/loggers/function_logger.py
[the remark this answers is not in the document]
bfineran: consistency with the values in the object, readability as well so it is clear where the variables are coming from

src/deepsparse/server/config.py
dbogunowicz: Actually, I still like the typing (well, I do not like how complicated it is but it expresses the right idea).
In the worst-case (most complex) scenario, the user could have the following setting:

```yaml
loggers:
    - python
    - some_other_logger
    - prometheus:
        - arg1: value
        - arg2: value
    - some_other_logger_1:
        - arg1: value
        - arg2: value
```

To me, such representation feels quite intuitive, and would like to support it in the `ServerConfig`.
It brakes down to a list that contains two strings (`['python', 'some_other_logger']) and two dictionaries
(['prometheus: {...}, `some_other_logger_1' :{...}']). @bfineran @corey-nm what am I missing here?

dbogunowicz: (this is why I think that typing: `Optional[List[str, Dict[str, Dict[str, Any]]]]` still makes sense.

src/deepsparse/loggers/function_logger.py
```diff
@@ -31,49 +33,76 @@ class FunctionLogger(BaseLogger):
...
     """
 
     def __init__(
         self,
         logger: BaseLogger,
-        target_identifier: str,
-        function: Callable[[Any], Any],
-        function_name: str = None,
-        frequency: int = 1,
+        target_to_metric_function_configs: Dict[
```
corey-nm, line 43: Bumping this: why are we changing FunctionLogger in this PR? Why can't you use the existing FunctionLogger?

src/deepsparse/server/config.py
corey-nm: can that not just be represented by a dict?

```yaml
loggers:
    python: {}
    some_other_logger: {}
    prometheus:
        - arg1: value
        - arg2: value
    some_other_logger_1:
        - arg1: value
        - arg2: value
```

I think dict is better repesentation here because names of loggers are unique. E.g. we can't have
```yaml
loggers:
  - prometheus: { ... }
  - prometheus: { ... }
```

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,125 @@
...
+from deepsparse.loggers.config import MetricFunctionConfig
+from deepsparse.server.config import ServerConfig
+
+
+__all__ = ["build_logger"]
+
+_LOGGER_MAPPING = {"python": logger_objects.PythonLogger}
+
+
+def build_logger(server_config: ServerConfig) -> logger_objects.BaseLogger:
```
corey-nm, line 31: Have you tried using this in server code? I think a more natural formulation will be to create a logger for each individual endpoint using leaf loggers
```python
leaf_loggers = ...

def build_endpoint(endpoint_cfg: EndpointConfig, leaf_loggers: Dict[str, BaseLogger]) -> Pipeline:
    func_loggers = []
    for target_logging_cfg in endpoint_cfg.data_logging:
          target = leaf_loggers[target_logging_cfg.target]
          logger = FunctionLogger(logger=target, function=parse_function(...))
          func_loggers.append(logger)
    return Pipeline.create(..., logger=MultiLogger(func_loggers)
```

corey-nm (changes requested):

src/deepsparse/server/build_logger.py
corey-nm: I think the most important part for this PR is to sketch out the use in server code - TBH the actual implementation of build_logger can be left to another PR. Otherwise when we start trying to integrate into server we may run into things that require changes to build_logger - which would mean we double the effort. Instead we should sketch things out from top down and fill in internal implementation details after the full system is sketched out

src/deepsparse/loggers/function_logger.py
```diff
@@ -31,49 +33,76 @@ class FunctionLogger(BaseLogger):
...
-        extracted_value = match_and_extract(
-            template=self.target_identifier, identifier=identifier, value=value
+        for (
+            target_identifier,
+            metric_function_cfgs,
+        ) in self.target_to_metric_function_configs.items():
+            extracted_value = match_and_extract(
+                template=target_identifier, identifier=identifier, value=value
+            )
+            if extracted_value is not None:
```
bfineran, line 71: let's change this to `if extracted_calue is None: continue`

src/deepsparse/loggers/helpers.py
```diff
@@ -11,15 +11,72 @@
...
+            raise ValueError(f"Path to the python script {path} does not exist")
+        spec = importlib.util.spec_from_file_location(
+            "user_defined_metric_functions", path
+        )
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        return getattr(module, func_name), func_name
+
+    func_name = function_identifier
+    return getattr(built_ins, func_name), func_name
```
bfineran, line 79: let's instead check for `func_name` in `built_ins` first so we can raise a useful error message if we can't match the function name

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,125 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Specifies the mapping from the ServerConfig to the DeepSparse Logger
+"""
+
+from typing import Any, Dict, List, Union
+
+from deepsparse import loggers as logger_objects
```
bfineran, line 21: nit: prefer to explicitly import classes and functions from this module rather than the entire module (logger_objects also could be a misnomer as non logger objects may be imported as well)

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,125 @@
...
+
+    leaf_loggers = build_leaf_loggers(loggers_config)
+
+    loggers = []
+    for name, leaf_logger in leaf_loggers.items():
+        target_to_metric_function_configs = _get_target_to_metric_function_configs(
+            endpoints=server_config.endpoints, logger_name=name
+        )
+        if target_to_metric_function_configs:
+            logger = logger_objects.FunctionLogger(
```
bfineran, line 61: why are we creating a separate function logger for each leaf logger? This would require us running the same function multiple times for separate leaf loggers, no?  Any reason we can't use one function logger here?

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,125 @@
...
+    Instantiate a set of leaf loggers according to the configuration
+
+    :param loggers_config: Data structure, that specifies the leaf loggers to
+        be instantiated
+    :return: A dictionary that contains a mapping from a logger's name to its instance
+    """
+
+    loggers = {}
+    for logger_config in loggers_config:
+        if isinstance(logger_config, str):
```
bfineran, line 86: along the lines of what @corey-nm commented before - why are we supporting the string case here? going with just dict leaves a lot less room for error

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,125 @@
...
+        else:
+            logger = leaf_logger
+
+        loggers.append(logger)
+
+    return logger_objects.MultiLogger(loggers) if len(loggers) > 1 else loggers[0]
+
+
+def build_leaf_loggers(
+    loggers_config: List[Union[str, Dict[str, Dict[str, Any]]]],
```
bfineran, line 74: let's change the type here to expect ` Dict[str, Dict[str, Any]]]`
we should essentially be returning `logger_name: _LOGGER_MAPPING[logger_name](logger_arguments)` where logger_name and logger_arguments are the keys/values respectively

src/deepsparse/server/config.py
bfineran: as mentioned in comments in `build_logger` agreed on just `dict`, we want to enforce simple, consistent behavior

src/deepsparse/loggers/function_logger.py
bfineran: this was discussed in the function logger PR, the goal is to remove the need for an extra multi logger that wraps all of the function loggers, additionally, match and extract only need to be run once under this configuration. note that the old functionality is still in tact if given a single function

src/deepsparse/server/build_logger.py
bfineran: server integration will occur in a follow up PR since that increases the testing scope by quite a bit. 

each endpoint or each function? the flow under the hood can be refactored later if there is a use case.

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,125 @@
...
+        target_to_metric_function_configs = _get_target_to_metric_function_configs(
+            endpoints=server_config.endpoints, logger_name=name
+        )
+        if target_to_metric_function_configs:
+            logger = logger_objects.FunctionLogger(
+                logger=leaf_logger,
+                target_to_metric_function_configs=target_to_metric_function_configs,
+            )
+        else:
+            logger = leaf_logger
```
bfineran, line 66: why are we defaulting to adding a leaf_logger to the top level? this should never be the case - leaf logging should always be under the function logger layer otherwise there will be no control over what is logged (ie all raw data would be passed to the leaf loggers).

let's refactor the function into two parts:
1. build a multi logger of leaf loggers
2. build the dictionary of identifier to metric configs to create a single function logger that wraps the leaf multi logger

bfineran (changes requested):
looking closer, the build_logger function still needs a bit of work, we can sync on this if needed

src/deepsparse/server/build_logger.py
dbogunowicz: This would require us to delegate even more functionalities to a single `FunctionLogger`. Not only "matching" and "applying the function", but also choosing which leaf logger to log (in the scenario where a metric is aiming to override the global loggers, which according to the latest news from Rob is likely to be the case).

Having to decide between scaling the number of `FunctionLoggers` linearly with the number of leaf loggers vs having even more responsibilities on the `FunctionLogger` side, I err for the former, not the latter.

src/deepsparse/loggers/function_logger.py
dbogunowicz: We are still in the process of figuring out the tradeoff between having fewer amount of loggers instantiated overall and less coupling between the FunctionLogger and the server inputs.

src/deepsparse/server/config.py
dbogunowicz: As mentioned in a different comment of mine, `python: {}` feels weird to me.

src/deepsparse/server/build_logger.py
bfineran: yes let's think on this, going back to multiple individual function loggers would make this a bit easier but we'd lose a bit of structure. we absolutely do not want to run the functions multiple times

src/deepsparse/server/build_logger.py
bfineran: these are two separate things - do not want to get away from the point that every leaf call in the server environment should exist behind (a/the) function logger

regarding the responsibilities - we need to map and extract values based on targets and run potentially a series of functions on them. we can do this either in an unstructured way with a series of multi loggers and function loggers, or have a single source of truth in a function logger. 

there's merits to both, concern with the former is understanding the composition and flow of the loggers will be a lot tougher and less clear without the structure (imagine being given the top level server logger as a python objet and needing to map out what is going on), while the second could limit flexibility in the future.

src/deepsparse/loggers/function_logger.py
corey-nm: Yeah I"m good with that - which is exactly what the current FunctionLogger does (and what was merged in on Friday). I'm asking why are there any changes to FunctionLogger in this PR? Why the switch to passing a config object instead of what is there now?

src/deepsparse/server/build_logger.py
dbogunowicz: How about "outsourcing" some responsibilities to MultiLogger? It could potentially be in charge of "turning on/off" the metric functions that override the global loggers. Just a thought. MultiLogger for now does nothing else but serves as a container for multiple loggers (which is its responsibility). But maybe it could do something more. Let's sleep on it.

src/deepsparse/loggers/function_logger.py
```diff
@@ -26,14 +26,14 @@
 
 class FunctionLogger(BaseLogger):
     """
-    DeepSparse logger that applies functions to raw,
-    logged values (collected in the log() method)
-    according to FunctionLogger's attributes
+    DeepSparse logger that applies a metric function to
+    a raw, values (received in the log() method) and then
```
dbogunowicz, line 30:
```suggestion
    a raw value (received in the log() method) and then
```

src/deepsparse/server/__init__.py
```diff
@@ -19,4 +19,4 @@
 the DeepSparse Engine.
 """
 
-from .cli import main
+# from .cli import main
```
dbogunowicz, line 22:
```suggestion
from .cli import main
```

dbogunowicz:
I will update the PR description and the master doc once the PR gets approved.

src/deepsparse/loggers/multi_logger.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+from typing import Any, List
+
+from deepsparse.loggers import BaseLogger, MetricCategories
+
+
+__all__ = ["MultiLogger"]
+
+
+class MultiLogger(BaseLogger):
```
corey-nm, line 27: 🔥

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,140 @@
...
+from deepsparse.loggers.helpers import get_function_and_function_name
+from deepsparse.server.config import ServerConfig
+
+
+__all__ = ["build_logger"]
+
+_LOGGER_MAPPING = {"python": PythonLogger}
+
+
+def build_logger(server_config: ServerConfig) -> BaseLogger:
```
corey-nm, line 31: 🔥

src/deepsparse/server/config.py
```diff
@@ -48,6 +48,16 @@ class ImageSizesConfig(BaseModel):
     )
 
 
+class TargetLoggingConfig(BaseModel):
+    target: str = Field(
+        description="Name of the target to apply the metric functions to."
+    )
+    functions: List[Dict[str, Any]] = Field(
```
corey-nm, line 55: optional nit to do later after this PR (non blocking IMO)
```suggestion
    functions: List[Dict[str, FunctionConfig]] = Field(
```
so we can get validation on FunctionConfig, but FunctionLogger would be unchanged (i.e. the build_logger would pass field from FUnctionConfig to FunctionLogger)

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,140 @@
...
+    loggers_config = server_config.loggers
+    if not loggers_config:
+        return None
+    leaf_loggers = build_leaf_loggers(loggers_config)
+
+    loggers = build_function_loggers(server_config.endpoints, leaf_loggers)
+    if not loggers:
+        loggers = list(leaf_loggers.values())
+
+    return MultiLogger(loggers) if len(loggers) > 1 else loggers[0]
```
corey-nm, line 59: non blocking comment, but right now this multi logger contains loggers for ALL endpoints. so there's a ton of extra filtering going on every log call. not an issue per se, but an easy thing to change if we have different loggers per pipelines

corey-nm (approved):
nice! things we can address in future PRs

src/deepsparse/server/build_logger.py
bfineran: a goal that was brought up was regex support across all endpoints which would require the loggers be shared

src/deepsparse/loggers/function_logger.py
```diff
@@ -50,24 +50,34 @@ def __init__(
...
+
         :param identifier: The name of the item that is being logged.
         :param value: The data structure that the logger is logging
         :param category: The metric category that the log belongs to
         """
         extracted_value = match_and_extract(
             template=self.target_identifier, identifier=identifier, value=value
         )
-        if extracted_value:
+        if extracted_value is not None:
```
bfineran, line 80: `None` should be a valid extracted value

src/deepsparse/loggers/helpers.py
```diff
@@ -11,15 +11,74 @@
...
+
+    if function_identifier.startswith("torch."):
+        func_name = function_identifier.split(".")[1]
+        return getattr(torch, func_name), func_name
+
+    if function_identifier.startswith("numpy.") or function_identifier.startswith(
+        "np."
+    ):
+        func_name = function_identifier.split(".")[1]
+        return getattr(numpy, func_name), func_name
```
bfineran, line 63: accessing only `function_identifier.split(".")[1]` isn't quite enough because there may be nested functions. for example `torch.linalg.l2` would return `torch.linalg`

src/deepsparse/loggers/helpers.py
```diff
@@ -11,15 +11,74 @@
...
+
+    if len(function_identifier.split(":")) == 2:
+        # assume a dynamic import function of the form
+        # '<path_to_the_python_script>:<function_name>'
+        path, func_name = function_identifier.split(":")
+        if not os.path.exists(path):
+            raise ValueError(f"Path to the python script {path} does not exist")
```
bfineran, line 70:
```suggestion
            raise ValueError(f"Path to the metric function file {path} does not exist")
```

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,140 @@
...
+    :return: a DeepSparse logger instance
+    """
+
+    loggers_config = server_config.loggers
+    if not loggers_config:
+        return None
+    leaf_loggers = build_leaf_loggers(loggers_config)
+
+    loggers = build_function_loggers(server_config.endpoints, leaf_loggers)
+    if not loggers:
```
bfineran, line 56: we need to think about this more - this would mean that ALL data gets logged which we don't want in a default case. Instead should just be logging timings here.

We should add some base function loggers to log timings and have those wrap the default loggers which would prevent this case. Can do in follow up if necessary, but let's remove this case before landing

src/deepsparse/server/build_logger.py
bfineran, line 59: Let's always return a multi logger even if length is 1 - this is for consistency so the server knows it can expect a MultiLogger if we ever need to access its properties.

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,140 @@
...
+) -> Dict[str, BaseLogger]:
+    """
+    Instantiate a set of leaf loggers according to the configuration
+
+    :param loggers_config: Config; specifies the leaf loggers to be instantiated
+    :return: A dictionary that contains a mapping from a logger's name to its instance
+    """
+    loggers = {}
+    for logger_name, logger_arguments in loggers_config.items():
+        leaf_logger = _LOGGER_MAPPING[logger_name](**logger_arguments)
```
bfineran, line 73: let's add an error check here

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,140 @@
...
+    return FunctionLogger(
+        logger=target_logger,
+        target_identifier=target_identifier,
+        function=function,
+        function_name=function_name,
+        frequency=function_cfg["frequency"],
+    )
+
+
+def _get_target_identifier(endpoint_name: str, target_name: str) -> str:
```
bfineran, line 139: need to account for regex here

src/deepsparse/server/config.py
bfineran: yes - the `FunctionConfig` is really the piece that needs validation. TargetLoggingConfig is less necessary as it's essentially a key value pair with target: functions. Prefer to change data_logging in server to be type `Dict[str, List[FunctionConfig]]`

so it would be 
```
data_logging:
  identifier_1:
      - function_config...
```

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,140 @@
...
+    if function_cfg.get("target_loggers"):
+        target_loggers = [
+            leaf_logger
+            for name, leaf_logger in loggers.items()
+            if name in function_cfg["target_loggers"]
+        ]
+    else:
+        target_loggers = list(loggers.values())
+
+    # if `target loggers` is a list len > 1, wrap it inside a MultiLogger
```
bfineran, line 124: again let's always just wrap in a multi logger for consistency

bfineran (changes requested):
lgtm pending comments

src/deepsparse/server/__init__.py
dbogunowicz, line 22:
```suggestion
from .cli import main
```

bfineran (approved):
LGTM, two small comments to think about

tests/server/test_build_logger.py
```diff
@@ -0,0 +1,136 @@
...
+endpoints:
+    - task: question_answering
+      route: /unpruned/predict
+      model: zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none
+      batch_size: 1
+      data_logging:
+        - target: re:*_outputs
+          functions:
+           - func: tests/test_data/metric_functions.py:user_defined_identity
+             frequency: 2"""
```
corey-nm, line 111: I interpreted ben's comment from earlier as more akin to:
```suggestion
        - re:*_outputs:
           - func: tests/test_data/metric_functions.py:user_defined_identity
             frequency: 2"""
```

where the key is the target, and the value is the list of functions. So notably, there isn't a target & functions key.

This would change the type of data_logging to
```python
Optional[Dict[str, List[Union[MetricFunctionConfig, str]]]]
```
& would require changes in `build_function_loggers`

@bfineran can you confirm?

bfineran: agreed, yes missed this in review - no need for the outer list

src/deepsparse/server/config.py
```diff
@@ -72,6 +111,10 @@ class EndpointConfig(BaseModel):
         default=1, description="The batch size to compile the model for."
     )
 
+    data_logging: Optional[
+        List[Dict[str, Union[List[MetricFunctionConfig], str]]]
```
bfineran, line 115: what is the purpose of the `str` in the union? is that handled in `build_logger`, let's remove if I'm not missing something

src/deepsparse/server/config.py
```diff
@@ -72,6 +111,10 @@ class EndpointConfig(BaseModel):
...
+    ] = Field(default=None, description="Specifies the rules for the data logging")
```
bfineran, line 116: let's expand the help text here

src/deepsparse/server/config.py
bfineran, line 115: per @corey-nm 's catch, we don't need this outer list, each identifier should be able to map to a list of its functions like so
```yaml
data_logging:
  target_1:
     - function: any
       freq: 1
     - function: any
       freq
  target_2: [{...}, ...]
```

```suggestion
        Dict[str, List[MetricFunctionConfig]]
```

src/deepsparse/server/build_logger.py
```diff
@@ -0,0 +1,146 @@
...
+    function_loggers = []
+    for endpoint in endpoints:
+        if endpoint.data_logging is None:
+            continue
+        for target_logging_dict in endpoint.data_logging:
+            target = target_logging_dict["target"]
+            target_identifier = _get_target_identifier(
+                endpoint.name or endpoint.task, target
+            )
+            for metric_function_cfg in target_logging_dict["functions"]:
```
bfineran, line 104: this is the piece where I'm missing the `Union[..., str]` case.
quick suggestion:
```suggestion
        for target, metric_function_cfgs in endpoint.data_logging.items():
            target_identifier = _get_target_identifier(
                endpoint.name or endpoint.task, target
            )
            for metric_function_cfg in metric_function_cfgs:
```

bfineran (changes requested):
took a second look following @corey-nm 's comments, looks like there were a few small issues with the `data_logging` type and processing. left suggestions to illustrate

PR #753, opened 2022-11-08
Generic Embedding Extraction Pipeline

bfineran:
TODO @rahul-tuli :
- [x] resolve code TODOS
- [x] pydoc
- [x] rename transformers embed extractor to `TransformersEmbeddingExtractor` (this is a temporary fix)
- [x] simple unit tests (mobilenet)
- [x] test with deepsparse.server

rahul-tuli: Added two new TODOs
- [X] Shape inference failing for non transformer models
 
needed for PR:
* testing

extra:
* potentially re-routing to transformers extraction helper for nlp models for attention layer support
* server compatibility (potentially future PR)
* testing with custom pipeline

Test with `deepsparse.server`

`config.yaml`
```yaml
num_cores: 2
num_workers: 2
endpoints:
  - task: embedding_extraction
    route: /predict
    model: default
    kwargs:
      base_task: image_classification
      emb_extraction_layer: -1
    name: ic_em_ex_pipeline_1
```

Test command:
```bash
deepsparse.server --config_file config.yaml
```


src/deepsparse/pipelines/embedding_extraction.py
```diff
@@ -0,0 +1,298 @@
...
+            "engine_type",
+            "batch_size",
+            "num_cores",
+            "scheduler",
+            "input_shapes",
+            "alias",
+            "context",
+            "executor",
+            "_delay_engine_initialize",
+        }
```
bfineran, line 130: let's use `inspect.signature(Pipeline).parameters.keys()` instead - if the kwargs were to change this will break silently

src/deepsparse/utils/onnx.py
```diff
@@ -345,3 +347,78 @@ def truncate_onnx_model(
...
+    """
+
+    """
+    onnx_filepath: str,
+    output_filepath: str,
+    final_node_names: List[str],
+    graph_output_names: List[str],
+    graph_output_shapes: Optional[List[List[int]]] = None,
+    """
+    output_filepath = output_path
```
bfineran, line 396: let's just rename the function arg to `output_filepath`

src/deepsparse/utils/onnx.py
```diff
@@ -345,3 +347,78 @@ def truncate_onnx_model(
...
+        emb_extraction_layer if isinstance(emb_extraction_layer, str) else None
+    )
+
+    if final_node_name is None:
+        model = onnx.load(model_path)
+
+        try:
+            sorted_layer_init_names = get_sorted_layer_init_names(model)
+        except Exception as exception:
+            raise RuntimeError(f"Failed to truncate onnx model: {exception}")
```
bfineran, line 412:
```suggestion
    if isinstance(emb_extraction_layer, str):
        final_node_name = emb_extraction_layer
    else:
        model = onnx.load(model_path)
        final_node_name = model.graph.node[emb_extraction_layer].name
        
        if final_node_name is None:
            raise ValueError(f"Node at index {emb_extraction_layer} does not have a name set")
```

src/deepsparse/pipelines/embedding_extraction.py
```diff
@@ -0,0 +1,298 @@
...
+        """
+        return self.base_pipeline.process_inputs(inputs)
+
+    # TODO @rahul-tuli: this assumes extra information from transformers need to make
+    #  this generic
+    def process_engine_outputs(
+        self,
+        engine_outputs: List[numpy.ndarray],
+        pad_masks: numpy.ndarray,
+        cls_masks: numpy.ndarray,
```
bfineran, line 241: this will need to accept any generic args passed in from base pipeline

```suggestion
        **kwargs,
```

src/deepsparse/pipelines/embedding_extraction.py
```diff
@@ -0,0 +1,224 @@
...
+    ):
+        self._base_task = base_task
+        self._emb_extraction_layer = emb_extraction_layer
+        self._return_numpy = return_numpy
+        self._flatten_outputs = flatten_outputs
+
+        # initialize engine after model truncate
+        self.base_pipeline = Pipeline.create(
+            task=base_task,
+            _delay_engine_initialize=True,
```
corey-nm, line 103: If the main purpose of this is to avoid doing the expensive model compilation, maybe we can explicitly set the engine_type to ORT here, and then revert it back to its default before calling _initialize_engine() below.

I think that should remove the need for this internal delay flag

bfineran: both approaches have small drawbacks, i'm in favor of leaving as is for now and revisiting later to clean up

rahul-tuli (approved):

src/deepsparse/pipelines/embedding_extraction.py
rahul-tuli: @corey-nm I've made a note of this and will come back with a better solution at a later time, for now leaving this as is unless this is a blocker from your side

corey-nm (approved):

PR #755, opened 2022-11-09
[Data Logging] Slicing/Indexing/Accessing Values

dbogunowicz:
Allows for simple indexing and slicing of the data logging targets.

**Before**
In the `ServerConfig` user could only specify the "string identifier"; a value to apply the metric function to:

```yaml
data_logging:
        - target: pipeline_inputs.images
          functions:
           ...
```

**Now**
The user can do simple indexing and slicing. This allows to e.g. apply the metric function to a specific part of the array, or extract information from the lists (e.g. when working with `engine_inputs`).

```yaml
data_logging:
        - target: pipeline_inputs.images[0, 2:-1]
          functions:
           ...
```

  Supported operations:
  - indexing: e.g value[0] or value[-2]
  - slicing: e.g value[0:2] or value[1:-3]
  - a composition of both: e.g value[0:2, 0] or value[1:-3, -1]


tests/deepsparse/loggers/test_helpers.py
```diff
@@ -30,5 +42,22 @@
...
 )
-def test_match(template, identifier, expected_output):
+def test_check_identifier_match(template, identifier, expected_output):
     assert check_identifier_match(template, identifier) == expected_output
+
+
+@pytest.mark.parametrize(
+    "value, remainder, expected_value",
+    [
+        ([0, 1, 2, 3], "key_1.key_2[2]", 2),
```
corey-nm, line 52:
```suggestion
        ([0, 1, 2, 3], "key_1.key_2[2]", 2),
        (..., "key_1[0].key_2[2]", ...),
```

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+    # splits remainder into separate strings.
+    # each string can access the new nesting depth and optionally
+    # do indexing and slicing at this depth
+    for sub_remainder in remainder.split("."):
+        square_brackets = re.search(r"\[(.*?)\]", sub_remainder)
+        if square_brackets:
+            # retrieve the string without the square brackets
+            sub_remainder = sub_remainder.split("[")[0]
+        # access the new nesting depth
+        value = value.__getattribute__(sub_remainder)
```
corey-nm, line 62:
```suggestion
        value = getattr(value, sub_remainder)
```

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
 
-    for sub_remainders in remainder.split("."):
-        value = value.__getattribute__(sub_remainders)
+    # splits remainder into separate strings.
+    # each string can access the new nesting depth and optionally
+    # do indexing and slicing at this depth
+    for sub_remainder in remainder.split("."):
+        square_brackets = re.search(r"\[(.*?)\]", sub_remainder)
+        if square_brackets:
+            # retrieve the string without the square brackets
```
corey-nm, line 59:
```suggestion
            # maps things like "value[0]" to "value"
```

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+            sub_remainder = sub_remainder.split("[")[0]
+        # access the new nesting depth
+        value = value.__getattribute__(sub_remainder)
+
+    if square_brackets:
```
corey-nm, line 64: should be inside the loop i think

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+    :param square_brackets: The string that contains the indexing and/or slicing
+        information inside square brackets
+    :return: The value of interest
+    """
+    for string_operator in square_brackets.split(","):
+        if ":" in string_operator:
+            # slicing
+            i, j = re.findall(r"-?\d+", string_operator)
+            i, j = int(i), int(j)
+            value = value.__getitem__(slice(i, j))
```
corey-nm, line 94:
```suggestion
            value = value.__getitem__(slice(*map(int, re.findall(r"-?\d+", string_operator))))
```

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+        else:
+            # indexing
+            i = int(string_operator)
+            value = value.__getitem__(i)
```
corey-nm, line 98: i think you can move the getitem outside of the loop to make it more accurate to what the index represents
```
indices = []
for op in square_brackets.split(","):
    indices.append(...)
value.__getitem__(indices)
```

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+
+    if square_brackets:
+        return do_slicing_and_indexing(
+            value=value, square_brackets=square_brackets.group(1)
+        )
+
+    return value
+
+
+def do_slicing_and_indexing(
```
corey-nm, line 72: non blocking comment, but this function seems super easily unit testable, so maybe add some tests for this one

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+    - slicing: e.g value[0:2] or value[1:-3]
+    - a composition of both: e.g value[0:2, 0] or value[1:-3, -1]
+
+    :param value: A sequential/tensor/array type variable to be indexed and/or sliced
+    :param square_brackets: The string that contains the indexing and/or slicing
+        information inside square brackets
+    :return: The value of interest
+    """
+    for string_operator in square_brackets.split(","):
+        if ":" in string_operator:
```
corey-nm, line 90: may be able to remove this if statement entirely if you do the map(int, ...) thing below

corey-nm (changes requested):
just some small fixes to add, otherwise flow looks good

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
     if not remainder:
         return value
 
-    for sub_remainders in remainder.split("."):
-        value = value.__getattribute__(sub_remainders)
+    # splits remainder into separate strings.
```
bfineran, line 53: this should all be moved to a helper function

tests/deepsparse/loggers/test_helpers.py
```diff
@@ -30,5 +42,22 @@
         ("re:string_*..*.string.*", "string_3.string_4", (True, None)),
     ],
 )
-def test_match(template, identifier, expected_output):
+def test_check_identifier_match(template, identifier, expected_output):
     assert check_identifier_match(template, identifier) == expected_output
+
+
+@pytest.mark.parametrize(
```
bfineran, line 49: need to test multiple repeated indices, indices between parameters, etc. not sure that these are currently covered
i.e. `key_1[0][0].key2[1].key3`

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+    # each string can access the new nesting depth and optionally
+    # do indexing and slicing at this depth
+    for sub_remainder in remainder.split("."):
```
bfineran, line 56: we can't use split here since a `.` since doesn't necessarily imply a property lookup.

for example:
`my_dict["some.key.like.this"].foo`

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+        square_brackets = re.search(r"\[(.*?)\]", sub_remainder)
+        if square_brackets:
+            # retrieve the string without the square brackets
+            sub_remainder = sub_remainder.split("[")[0]
```
bfineran, line 60: what do we do with the rest here?

src/deepsparse/loggers/helpers.py
```diff
@@ -50,8 +50,52 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+    :param value: A sequential/tensor/array type variable to be indexed and/or sliced
+    :param square_brackets: The string that contains the indexing and/or slicing
+        information inside square brackets
+    :return: The value of interest
+    """
+    for string_operator in square_brackets.split(","):
+        if ":" in string_operator:
+            # slicing
+            i, j = re.findall(r"-?\d+", string_operator)
+            i, j = int(i), int(j)
```
bfineran, line 93: this will fail if they aren't ints, i.e. it's a dictionary lookup with a `:` as a key, should do a check for a dict at the top and just return a get item if so, potentially try checking for numeric keys first

bfineran:
@corey-nm and I had an offline conversation about error handling when a target starts with an identifier but does not contain the property or index

i.e.

```python
# target: phase_1.x

self.logger.log("phase_1", 100)  # should we crash here since 100 does not have property "x"?
# ...
self.logger.log("phase_1.x", 110)  # how the target could be valid currently
```

our proposed solution is to not allow `.` or `[]` values in logged identifiers. pytorch does a similar thing with submodule naming conventions

src/deepsparse/loggers/helpers.py
dbogunowicz: nothing. the actual body of the square brackets gets extracted using regex

src/deepsparse/loggers/python_logger.py
```diff
@@ -29,6 +29,7 @@ class PythonLogger(BaseLogger):
     Python logger that writes the collected logs to stdout
     """
 
+    @validate_identifier
```
dbogunowicz, line 32:
```suggestion
```

src/deepsparse/loggers/python_logger.py
```diff
@@ -18,7 +18,7 @@
 from datetime import datetime
 from typing import Any
 
-from deepsparse.loggers import BaseLogger, MetricCategories
+from deepsparse.loggers import BaseLogger, MetricCategories, validate_identifier
```
dbogunowicz, line 21:
```suggestion
from deepsparse.loggers import BaseLogger, MetricCategories
```

corey-nm (approved):
lgtm, supporting string indexes can be done in a separate PR imo

src/deepsparse/loggers/helpers.py
```diff
@@ -119,9 +123,69 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+        square_brackets, sub_remainder = _check_square_brackets(sub_remainder)
+        if not hasattr(value, sub_remainder):
+            raise ValueError(
+                "Attempting to access an non existing "
+                f"attribute {sub_remainder} of an object {value}"
+            )
+        value = getattr(value, sub_remainder)
+
+        if square_brackets:
+            value = do_slicing_and_indexing(
```
bfineran, line 138: should check if value is a dict first

dbogunowicz: why dict?

src/deepsparse/loggers/helpers.py
```diff
@@ -119,8 +123,85 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
...
+
+    :param value: A sequential/tensor/array type variable to be indexed and/or sliced
+    :param square_brackets: The string that contains the indexing and/or slicing,
+        and/or dictionary access information
+    :return: The value of interest
+    """
+    for string_operator in square_brackets.split(","):
+        # check whether `string_operator` contains at least one character
+        # -> must be a dictionary key
+        if string_operator.upper().isupper():
```
corey-nm, line 187: isascii()?

tests/deepsparse/loggers/test_helpers.py
```diff
@@ -66,5 +71,68 @@ def test_get_function_and_function_name(
...
+        ([[0, 1, 2, 3]], "0, 0:3", [0, 1, 2]),
+        ([[0, 1, 2, 3]], "0, -1", 3),
+        ([[0, 1, 2, 3]], "0, 0:3, 2", 2),
+        ([[0, 1, 2, 3]], "0, 0:3, 2", 2),
+        ([0, 1, 2, 3, 4], "1:-2, 0", 1),
+        ([[0, 1, 2, 3, 4]], "0, 1:-1:1", [1, 2, 3]),
+        (numpy.array([[0, 1, 2, 3, 4]]), "0, 1:-1:1", numpy.array([1, 2, 3])),
+        (torch.tensor([[0, 1, 2, 3, 4]]), "0, 1:-1:1", torch.tensor([1, 2, 3])),
+        ({"key_1": {"key_2": [3]}}, "'key_1','key_2',0", 3),
+        ({"key_1": [0, 1, {"key_2": [3, 4, 5]}]}, "'key_1',2,'key_2',0", 3),
```
corey-nm, line 124: nice tests 🔥

corey-nm (approved):

src/deepsparse/loggers/helpers.py
dbogunowicz: pure digits are also ascii:

```bash
x = '3'
x.isascii()
>> True
```

corey-nm: ohhh yeah duh. okay how about `isalpha()`? should be alphabet characters

dbogunowicz: i tried to avoid the `string_operator.upper().isupper()` as hard as I could - it is sadly (despite expression's ugliness), my best attempt to make it work

src/deepsparse/pipeline.py
```diff
@@ -703,6 +703,7 @@ def log(self, identifier: str, value: Any, category: str):
         :param value: The logged data structure
         :param category: The metric category that the log belongs to
         """
+        validate_identifier(identifier)
         if self.logger:
             self.logger.log(
                 identifier=f"{self.alias or self.task}.{identifier}",
```
bfineran, line 709: Isn't this an immediate violation?

src/deepsparse/loggers/helpers.py
bfineran, line 187: IMO no need to check for things like this, if it's a dict we should pass through to the dict. If not, we should delegate the error handling to the accessed object. Could lead to unwanted failures in the future

bfineran (approved):

src/deepsparse/pipeline.py
bfineran: should change the `.` here before `{identifier}` to some thing else as discussed offline. final identifier should pass validation as well

PR #756, opened 2022-11-09
[Data Logger] Include `build_logger` in the server logic

dbogunowicz:
Changes tested manually


src/deepsparse/server/helpers.py
```diff
@@ -11,86 +11,28 @@
...
-]
 
+__all__ = ["log_system_info", "default_logger"]
 
 _LOGGER = logging.getLogger(__name__)
-_SUPPORTED_LOGGERS = {"prometheus": PrometheusLogger}
 
 
-def default_logger_manager() -> ManagerLogger:
+def default_logger() -> PythonLogger:
```
corey-nm, line 28:
```suggestion
def default_logger() -> BaseLogger:
```

src/deepsparse/server/helpers.py
```diff
@@ -11,86 +11,28 @@
...
-            raise ValueError(
-                f"Unknown logger integration {integration}. Supported integrations: "
-                f"{list(_SUPPORTED_LOGGERS)}"
-            )
-        logger = logger_class(**logger_kwargs)
-        _LOGGER.info(f"Created logger {logger}")
-        loggers.append(logger)
 
-    return ManagerLogger(loggers=loggers)
+def log_system_info(server_logger: BaseLogger):
```
corey-nm, line 37: 👍

src/deepsparse/server/server.py
```diff
@@ -208,42 +206,32 @@ def _add_endpoint(
...
     context: Context,
-    pipeline_logger: ManagerLogger,
+    server_logger: BaseLogger,
 ):
     pipeline_config = endpoint_config.to_pipeline_config()
     pipeline_config.kwargs["executor"] = executor
 
     _LOGGER.info(f"Initializing pipeline for '{endpoint_config.name}'")
-    pipeline = Pipeline.from_config(pipeline_config, context)
+    pipeline = Pipeline.from_config(pipeline_config, context, server_logger)
```
corey-nm, line 215: i think this is fine for now, but still think we should have separate loggers for each pipeline. can do in a separate PR

src/deepsparse/server/server.py
```diff
@@ -208,42 +206,32 @@ def _add_endpoint(
...
-            pipeline_outputs,
-            pipeline_inputs,
-            engine_inputs,
-            inference_timing,
-        ) = pipeline.run_with_monitoring(request)
-        pipeline_logger.log_latency(pipeline_name, inference_timing)
-        pipeline_logger.log_data(
-            pipeline_name, (pipeline_inputs, engine_inputs), pipeline_outputs
-        )
+        pipeline_outputs = pipeline(request)
```
corey-nm, line 231: so much cleaner 🔥

src/deepsparse/server/server.py
```diff
@@ -208,42 +206,32 @@ def _add_endpoint(
...
+        server_logger = pipeline.logger
+        if server_logger:
+            log_system_info(server_logger)
```
corey-nm, line 234: fine for now, i think we may want to move this into a background thread in the future. e.g. if server doesn't receive a request for an hour we still probablky want to be logging system info during that time

corey-nm (approved):
lgtm, can adjust whether we have single logger or multiple per pipeline in another PR

src/deepsparse/server/config.py
```diff
@@ -166,6 +166,28 @@ class ServerConfig(BaseModel):
         ),
     )
 
+    @validator("endpoints")
```
rahul-tuli, line 169: nice!

src/deepsparse/server/build_logger.py
```diff
@@ -93,9 +97,7 @@ def build_function_loggers(
             continue
         for target_logging_cfg in endpoint.data_logging:
             target = target_logging_cfg.target
-            target_identifier = _get_target_identifier(
-                endpoint.name or endpoint.task, target
-            )
+            target_identifier = _get_target_identifier(endpoint.name, target)
```
rahul-tuli, line 100: I'm assuming this is safe now due to the validator added to `EndPointConfig`?

bfineran (changes requested):
LGTM - let's use this PR to update our E2E tests

src/deepsparse/server/server.py
```diff
@@ -140,7 +143,7 @@ def _health():
     def _add_endpoint_endpoint(cfg: EndpointConfig):
         if cfg.name is None:
             cfg.name = f"endpoint-{len(app.routes)}"
-        _add_endpoint(app, server_config, cfg, executor, context, pipeline_logger)
+        _add_endpoint(app, server_config, cfg, executor, context, server_logger)
```
rahul-tuli, line 146: nit: could we now change these to keyword arguments? like the following: 
```
_add_endpoint(app=app, server_config=server_config, ...)
```

My train of thought being that with this, the order of calling arguments would not matter and reduce potential bugs?

I also feel it provides a first level of documentation on how to call a function.

But this is a personal preference and is not a blocker

src/deepsparse/server/server.py
rahul-tuli: +1 that is a great callout

src/deepsparse/server/server.py
rahul-tuli: +1

src/deepsparse/pipeline.py
```diff
@@ -541,6 +542,8 @@ def from_config(
             other runtime information that will be used across instances of the
             MultiModelEngine to provide optimal performance when running
             multiple models concurrently
+        :param logger: An optional DeepSparse Logger object for inference
```
rahul-tuli, line 545: nit: DeepSparse Base Logger?

src/deepsparse/server/build_logger.py
dbogunowicz: yessir

src/deepsparse/pipeline.py
dbogunowicz: I mean, some kind of logger. Every DeepSparse Logger will inherit from the Base Logger.

src/deepsparse/server/config.py
```diff
@@ -191,14 +191,36 @@ class ServerConfig(BaseModel):
 
     endpoints: List[EndpointConfig] = Field(description="The models to serve.")
 
-    loggers: Dict[str, Optional[Dict[str, Any]]] = Field(
+    loggers: Dict[str, Dict[str, Any]] = Field(
```
dbogunowicz, line 194:
```suggestion
    loggers: Dict[str, Optional[Dict[str, Any]]] = Field(
```

dbogunowicz:
All the server tests are locally green except for:
- `test_prometheus` - obviously, since Prometheus "does not exist yet" from this branches POV
- `test_hot_reloading.py::test_task_cli_hot_reload` - we have established with Corey that this may be something that might be worth taking a look once we have GHA running on the branch (so once this lands in the feature branch).

bfineran (approved):

PR #760, opened 2022-11-11
[Data Logging] Initial built-in functions

dbogunowicz:
Introduces the first set of built-in functions:
```
  bounding_box_count,
  fraction_zeros,
  image_shape,
  max_pixels_per_channel,
  mean_pixels_per_channel,
  std_pixels_per_channel
```


corey-nm (approved):
nice!

src/deepsparse/loggers/metric_functions/built_ins.py
```diff
@@ -11,12 +11,192 @@
...
+             tensor/array
+           - the image has 3 or 1 channels
+    :return: A float in range from 0. to 1.
+    """
+    image_numpy = _assert_numpy_image(img)
+    _check_valid_image(image_numpy)
+    return (image_numpy.size - numpy.count_nonzero(image_numpy)) / image_numpy.size
+
+
+def bounding_box_count(bboxes: List[List[Optional[List[float]]]]) -> Dict[int, int]:
```
bfineran, line 144: We should probably support the yolact and yolo outputs directly here. Additionally a generic len or shape function would be useful

bfineran (approved):

src/deepsparse/loggers/metric_functions/built_ins.py
dbogunowicz: Yeah, those are in the end the yolact and yolo outputs (`List[List[Optional[List[float]]]]` for yolo and `List[List[List[float]]]` for yolact). Let's land now, we will be revisiting the built-ins anyway.

PR #764, opened 2022-11-14
[data-logging] support server default logging of all pipeline 'SYSTEM' logs

bfineran:
Currently, for a log to be propagated, the user must add a function logger definition for it in the endpoint config

By default, we would like to enable the logging of all pipeline system logs (right now this includes just the base timings recorded by the pipeline). This PR enables the propagating of these logs with the following changes:

1) support for a new identifier pattern `category:<CATEGORY>` which matches to any log of a given `MetricCategories` value
2) adding a function logger in build loggers that matches to `category:system` and passes these logs through on each call with an `identity` function

future work will enable the user to disable specific or all system logs via the config, logic for this may be added through updates to `_create_system_logger`

**test_plan**:
unit tests for identifier matching and `build_logger` updated

### DIFF
--- a/src/deepsparse/loggers/function_logger.py
+++ b/src/deepsparse/loggers/function_logger.py
@@ -70,7 +70,10 @@ class FunctionLogger(BaseLogger):
         :param category: The metric category that the log belongs to
         """
         extracted_value = match_and_extract(
-            template=self.target_identifier, identifier=identifier, value=value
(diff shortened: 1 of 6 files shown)


dbogunowicz (approved):
Very nice, I like how easily function logger has been extended to handle system logging.

corey-nm (approved):

PR #766, opened 2022-11-14
[data-logging] AsyncLogger implementation

bfineran:
basic implementation for a `AsyncLogger` that submits logging jobs that do not block the main process.

current implementation uses a threadpool, however should further evaluate tradeoffs with process pool, and potential future need for a more sophisticated job worker

open questions:
* how to best propagate errors after jobs have been submitted and user requests have been returned. currently issuing a python `error` log that displays the error message however this is more of a "silent" failure which is not ideal - a separate job worker may be able to better handle this
* exploration of limitations of executors and potential overhead added

**test_plan:**
included tests to verify that calls are non blocking, numpy arrays and functions may be submitted, verification of job run and error propagation

more thorough testing should be added in E2E tests

### DIFF
--- a/src/deepsparse/loggers/__init__.py
+++ b/src/deepsparse/loggers/__init__.py
@@ -12,10 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .base_logger import *
-
 # flake8: noqa
+# isort: skip_file
+
+# base modules
+from .base_logger import *
 from .constants import *
+
+# logger implementations
+from .async_logger import *
 from .function_logger import *
 from .multi_logger import *
 from .prometheus_logger import *
--- a/src/deepsparse/loggers/async_logger.py
+++ b/src/deepsparse/loggers/async_logger.py
@@ -1,0 +1,66 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Logger wrapper to run log calls asynchronously to not block the main process
+"""
+
+import logging
+from concurrent.futures import Executor, ThreadPoolExecutor
+from typing import Any
+
+from deepsparse.loggers import BaseLogger, MetricCategories
+
+
+__all__ = ["AsyncLogger"]
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class AsyncLogger(BaseLogger):
+    """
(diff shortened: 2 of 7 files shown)


src/deepsparse/loggers/__init__.py
```diff
@@ -12,10 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from .base_logger import *
-
 # flake8: noqa
+# isort: skip_file
```
dbogunowicz, line 16: to disable isort from enforcing his own, messy line ordering?

src/deepsparse/loggers/async_logger.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def log(self, identifier: str, value: Any, category: MetricCategories):
+        """
+        Forward log calls to wrapped logger to run asynchronously
+
+        :param identifier: The name of the item that is being logged.
+        :param value: The data structure that the logger is logging
+        :param category: The metric category that the log belongs to
+        """
+        job_future = self._job_pool.submit(
```
dbogunowicz, line 54: Nice, also a very well-documented set of tests.

I think optimizing the `AsyncLogger` at this point would be less important than optimizing the pre/post-processing of the pipelines. I would for now leave this PR as it is. We should probably talk to the runtime team/use the cprofiler with snakeviz to understand the possible overhead we are adding with the data logging. But I do not see this as the prority atm.

dbogunowicz (approved):

src/deepsparse/loggers/async_logger.py
```diff
@@ -0,0 +1,66 @@
...
+    the main process. No call back/returned future currently provided
+
+    :param logger: logger object to wrap
+    :param max_workers: maximum logging tasks to run in the job pool at once.
+        defaults to 1
+    """
+
+    def __init__(self, logger: BaseLogger, max_workers: int = 1):
+        self.logger = logger
+        self._job_pool: Executor = ThreadPoolExecutor(max_workers=max_workers)
```
corey-nm, line 44: Option to pass this in to this object? I think server could pass it the thread pool executor it owns to this so they share one. Or do we explicitly want a separate one I guess? How do 2 thread pool executors interact?

src/deepsparse/server/build_logger.py
```diff
@@ -58,7 +59,10 @@ def build_logger(server_config: ServerConfig) -> BaseLogger:
 
     leaf_loggers = build_leaf_loggers(loggers_config)
     loggers = build_function_loggers(server_config.endpoints, leaf_loggers)
-    return MultiLogger(loggers)
+    return AsyncLogger(
```
corey-nm, line 62: 🔥

tests/deepsparse/loggers/helpers.py
```diff
@@ -0,0 +1,81 @@
...
+
+from deepsparse.loggers import BaseLogger, MetricCategories
+
+
+__all__ = [
+    "ErrorLogger",
+    "FileLogger",
+    "NullLogger",
+    "SleepLogger",
+]
```
corey-nm, line 32: i love em all

corey-nm (approved):
lgtm

src/deepsparse/loggers/async_logger.py
bfineran: yes I like this, let's get it added in later. 2 thread pools should be ok depending on usage, can lead to deadlock in certain scenarios. let's come back to this in testing

PR #767, opened 2022-11-15
Haystack integration test and DeepSparseReader for ExtractiveQA

mgoin:
Add a github action for integrations since the haystack dependencies are so different compared to `deepsparse[dev]`


corey-nm (approved):
Nice changes!

bfineran (approved):

PR #775, opened 2022-11-16
[Data-Logging] add protection to hard prometheus import

bfineran:
without this check any import of loggers will fail if prometheus is not installed


dbogunowicz (approved):
good catch

PR #783, opened 2022-11-17
Refresh examples/benchmark comparison with ORTEngine

mgoin:
Also changing the max diff exception in `verify_outputs` to an error log since it doesn't make sense to crash the process


KSGulin (approved):
LGTM

bfineran (approved):

PR #787, opened 2022-11-18
[Open Pif Paf] STAGE 0 - onnx export and basic pipeline implementation

dbogunowicz:
# OpenPifPaf Stage 0 

## Onnx Export
Without any network training, the OpenPifPaf devs allow the user to export any of the untrained versions of their model (Resnet-/Mobilenet-/ SwinTransformers- /... based) as ONNX file (`https://openpifpaf.github.io/cli_help.html?highlight=onnx#export-onnx`). They also apply utilities to optimize and simplify the resulting onnx file.

image

## Quick Benchmarking
Roughly we have 2x speedup running the sample OpenPifPaf model with the engine vs onnxruntime

image

## Minimal pipeline implementation for Open Pif Paf 
```python
import numpy as np


pipeline = Pipeline.create(task="open_pif_paf", batch_size = 1)
out = pipeline(images=[np.random.rand(1, 3, 97, 129).astype(np.float32)])
```

`out: List[numpy.array]` with two numpy arrays of size `(1,17,5,13,17)` and `(1,19,8,13,17)` (consistent with the expectations, see: PIF and PAF network components in the original OpenPifPaf paper)

### DIFF
--- a/src/deepsparse/open_pif_paf/pipelines.py
+++ b/src/deepsparse/open_pif_paf/pipelines.py
@@ -1,0 +1,52 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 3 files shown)


rahul-tuli (commented):
Minor comments! Great first pass

src/deepsparse/open_pif_paf/pipelines.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+__all__ = [
+    "OpenPifPafPipeline",
+]
+
+
+@Pipeline.register(
+    task="open_pif_paf",
+    default_model_path=("/home/damian/pifpaf/openpifpaf-resnet50.onnx"),
```
mgoin, line 31: do we have a public onnx we can point to?

mgoin: yup okay, getting that for a pretrained model with an annotation example is what i'd like to have eventually

PR #788, opened 2022-11-21
[OpenPifPaf] Productionizing the pipeline

dbogunowicz:


src/deepsparse/open_pif_paf/pipelines.py
```diff
@@ -45,8 +81,50 @@ def output_schema(self) -> Type[OpenPifPafOutput]:
...
+    def process_engine_outputs(self, fields, **kwargs):
+        """
+        Fields (raw network outputs) are a list of numpy arrays of sizes:
+        (B,17,5,13,17) (CIF) and (B,19,8,13,17) (CAF).
+        The processor maps fields into the actual list of pose annotations
+        """
+        for idx, (cif, caf) in enumerate(zip(*fields)):
+            annotations = self.processor._mappable_annotations(
+                [torch.tensor(cif), torch.tensor(caf)], None, None
+            )
```
corey-nm, line 108: Ahh I see, so this is why we'd need to depend on openpifpaf? You said this is a pretty complicated function too right?

dbogunowicz: I will do a more in-depth write up on this, but yeah, self.processor is basically a composition of decoders, each decoder is a fairly complex structure.

corey-nm (dismissed):

src/deepsparse/open_pif_paf/pipelines.py
```diff
@@ -15,38 +15,107 @@
...
+        class properties into an inference ready onnx file to be compiled by the
+        engine of the pipeline
+
+        :return: file path to the ONNX file for the engine to compile
+        """
         return self.model_path
 
-    def process_inputs(self, inp):
-        return inp.images
+    def process_inputs(self, inputs: OpenPifPafInput) -> List[numpy.ndarray]:
```
dbogunowicz, line 82: We could unify this for all the CV tasks.

src/deepsparse/open_pif_paf/annotate.py
```diff
@@ -0,0 +1,233 @@
...
+open_pif_paf_default_stub = None
+
+DEEPSPARSE_ENGINE = "deepsparse"
+ORT_ENGINE = "onnxruntime"
+TORCH_ENGINE = "torch"
+
+_LOGGER = logging.getLogger(__name__)
+
+
+@click.command()
```
corey-nm, line 90: Will allow you to remove the underscore & hyphens options, AND the show_default=True
```suggestion
@click.command(
    context_settings=dict(
        token_normalize_func=lambda x: x.replace("-", "_"), show_default=True
    ),
)
```

src/deepsparse/open_pif_paf/annotate.py
```diff
@@ -0,0 +1,233 @@
...
+    open_pif_paf_pipeline = Pipeline.create(
+        task="open_pif_paf",
+        model_path=model_filepath,
+        engine_type=engine,
+        num_cores=num_cores,
+    )
+
+    for iteration, (input_image, source_image) in enumerate(loader):
+        if iteration % 5 != 0:
+            continue
```
corey-nm, line 206: optional nit: make this a cli arg?

src/deepsparse/open_pif_paf/utils/annotate.py
```diff
@@ -0,0 +1,193 @@
...
+        numpy.divide(input_image_resolution, kwargs["model_resolution"])
+    )
+
+    for annotation in annotations:
+        data = numpy.multiply(annotation.data[:, :2], scale)
+        keypoints = annotation.keypoints
+        image = _draw_skelethon(image, annotation.skeleton, data, keypoints)
+        image = _draw_joints(image, data)
+        image = _draw_confidence(image, data[0], annotation.score)
+        if images_per_sec:
```
corey-nm, line 91: Should this be outside the for loop

src/deepsparse/open_pif_paf/utils/annotate.py
```diff
@@ -0,0 +1,193 @@
...
+                images_per_sec=images_per_sec,
+                x=20,
+                y=30,
+                font_scale=0.9,
+                thickness=2,
+            )
+    return image
+
+
+def _draw_confidence(
```
corey-nm, line 103: are all these private functions modified from openpifpaf repo? if so optional nit is to say so in docstrings

corey-nm (approved):
just some optional nits, looks great to me!

src/deepsparse/open_pif_paf/utils/annotate.py
dbogunowicz: No, I wrote them. Note how some of them reuse our previous functions :)

corey-nm: Oh awesome! 👍

src/deepsparse/open_pif_paf/utils/annotate.py
dbogunowicz: Right

src/deepsparse/open_pif_paf/annotate.py
dbogunowicz: This should not be there. A small code change to create demos for Alex faster :)

src/deepsparse/open_pif_paf/utils/annotate.py
```diff
@@ -0,0 +1,193 @@
...
+    _put_annotation_text,
+)
+from openpifpaf import decoder, network
+
+
+_LOGGER = logging.getLogger(__name__)
+
+# necessary openpifpaf dependencies for now
+model_cpu, _ = network.Factory().factory(head_metas=None)
+processor = decoder.factory(model_cpu.head_metas)
```
bfineran, line 43: initialization of these must be moved to behind a function otherwise this will run as soon as anything from open pif paf directory is imported

bfineran (changes requested):
LGTM pending comment

Questions:
 * since we are returning the encoded outputs directly, what is the intended use for the user?
 * what is the intended server environment flow?
 * test plan?

src/deepsparse/open_pif_paf/utils/annotate.py
```diff
@@ -0,0 +1,193 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+try:
+    import openpifpaf
+
+    openpifpaf_error = None
+except ModuleNotFoundError as openpifpaf_import_error:
+    openpifpaf = None
+    openpifpaf_error = openpifpaf_import_error
```
bfineran, line 21: where is this used or raised?

src/deepsparse/open_pif_paf/utils/annotate.py
dbogunowicz: I do not understand. What is going to happen is:
1.  `src/deepsparse/open_pif_paf/annotate.py` is ran
2.  during the importing step, `src/deepsparse/open_pif_paf/utils/annotate.py` is being imported and lines 42/43 are evoked. `processor` gets created.
3. until termination, only the `annotate_image` function is called. Note: lines 42/43 never get run again.

src/deepsparse/open_pif_paf/utils/annotate.py
dbogunowicz: Yeah, this is not correct here.
What is the convention around making sure that we are setting up properly all the dependencies for the pipeline?
Proper edits in `setup.py` and pipelines `__init__.py`?

dbogunowicz:
@bfineran 

## Question 1
Ideally, I see the OpenPifPaf integration work like this:
1. The pipeline
The pipeline takes as an input two arguments: `image` (delivered by the user) and `config` (that comes together with the onnx model in the `deployment` folder). The `config` would specify the apriori information about the skeletons to be recognized (information from the training pipeline).  Once the network returns the CifCaf fields (engine forward), in the `post_processing` phase of the pipeline we use the config to turn fields into human-understandable detections (what they call decoding in the openpifpaf paper) (e.g. here https://github.com/openpifpaf/openpifpaf/blob/main/src/openpifpaf/csrc/src/cifcaf.cpp one can see that the cifcaf decoding step does use skeleton structure priors). The final output from the pipeline could be e.g. 
annotations: List[annotation]
annotation: BaseModel with properties `data: np.ndarray`, `keypoints: List[str]`, `score: float`, `skeleton: List[int,int]`

2. Visualisation
I could rewrite the visualization pipeline that could be fully configurable through the final output, i.e. it could figure out the structure of the skeleton from the output schema of the pipeline.

Actually, this does seem like a better idea for now - > pipeline returns a human-readable output in a specific structure, that can be automatically recognized by the visualization pipeline, potentially for different types of skeletons. I can quickly rework it after your response. If this is not clear, please lets sync offline.

## Question 2
Yeah, for the server use we probably would like to refrain from passing any numpys and directly returning human-understandable labels. This even strongly calls for the proposed modification.

## Question 3
Right now, it has been tested manually (will fill out PR description); but I did not see this as top priority, as we will be focusing on integration tests of pipelines as a next task anyway right?

dbogunowicz:
@bfineran 

I moved some of the postprocessing from visualization to the pipeline, so the user receives human-readable output from the pipeline and thus the pipeline can be also used efficiently in the server.

src/deepsparse/open_pif_paf/utils/annotate.py
corey-nm: Yeah I think ben's initial concern won't happen since the `__init__.py` files don't import from sub modules. Another option could be have an Annotator class that creates/holds the processor in constructor, and it just has the `annotate_image` as a method. Might be overkill though. A really hacky option is to store the processor on the function itself (yes you can set attributes on functions lmao):

```python
def annotate_image(...):
    if annotate_image.processor is None:
        annotate_image.processor = ...
    processor = annotate_image.processor
```

src/deepsparse/open_pif_paf/utils/annotate.py
corey-nm: My understanding is:
1. have feature in setup.py to install dependencies as you mention
2. in openpifpaf/__init__.py do a check to see if it's installed, and raise error if not
3. (optional) auto install dependencies in openpifpaf/__init__.py if not installed

src/deepsparse/open_pif_paf/utils/annotate.py
dbogunowicz: After consideration, moved the openpifpaf dependencies to my pipeline itself.

src/deepsparse/open_pif_paf/annotate.py
```diff
@@ -0,0 +1,235 @@
...
+_LOGGER = logging.getLogger(__name__)
+
+
+@click.command(
+    context_settings=dict(
+        token_normalize_func=lambda x: x.replace("-", "_"), show_default=True
+    ),
+)
+@click.option(
+    "--model_filepath",
```
corey-nm, line 96: optional nit: you can remove the underscore options now since the token_normalize_func was added. (same comment to all underscore options below)
```suggestion
```

src/deepsparse/open_pif_paf/annotate.py
```diff
@@ -0,0 +1,235 @@
...
+    "--model-filepath",
+    type=str,
+    default=open_pif_paf_default_stub,
+    help="Path/SparseZoo stub to the model file to be used for annotation",
+    show_default=True,
```
corey-nm, line 101: optional nit: you can remove this since it was moved to show_default (and same comment to all show_defaults below)

src/deepsparse/open_pif_paf/annotate.py
```diff
@@ -0,0 +1,235 @@
...
+    engine: str,
+    image_shape: Tuple[int, int],
+    num_cores: Optional[int],
+    save_dir: str,
+    name: Optional[str],
+    target_fps: Optional[float],
+    no_save: bool,
+) -> None:
+    """
+    Annotation Script for OpenPifPaf with DeepSparse
```
corey-nm, line 184: this function is so clean 🔥

src/deepsparse/open_pif_paf/utils/annotate.py
```diff
@@ -0,0 +1,208 @@
...
+    data: numpy.ndarray,
+    joint_thickness: int = 6,
+    trace_thickness: int = 2,
+) -> numpy.ndarray:
+    """
+    Draw the joints of the pose estimation
+    """
+    for joint in data:
+        if all(joint) is False:
+            continue
```
corey-nm, line 147: not sure if this suggestion improves things but figured i'd toss it out
```suggestion
    for joint in filter(all, data):
```

corey-nm (dismissed):
lgtm! really clean implementations, nice! 🚀

corey-nm (approved):
🔥

bfineran (approved):
Great work @dbogunowicz 

following @corey-nm 's comments in vit-pose let's get a simple readme up (can be a quick follow up). also as discussed offline let's do a brief check on postprocessing times using our new logging tools :)

dbogunowicz:
@bfineran , per `PythonLogger`, this is mean and std for 20 consecutive forward passes using a dense openpifpaf pose estimation model that takes 384x384 images as input (this is the resolution with which @anmarques is working with at the moment)

```bash
pre_process_latency[s] 0.0044 +- 0.0015
engine_forward_latency[s] 0.1699 +- 0.0173
post_process_latency[s] 0.0050 +- 0.0010
```

No red flags here seems that on average both pre- and post-processing steps are two OOM smaller than the engine forward.

corey-nm (approved):
Nice!!

bfineran (approved):

PR #791, opened 2022-11-22
[Data Logging] Target identifier bug fix / Making PrometheusLogger more robust

dbogunowicz:


src/deepsparse/loggers/helpers.py
```diff
@@ -142,12 +142,13 @@ def possibly_extract_value(value: Any, remainder: Optional[str] = None) -> Any:
         # check whether sub_remainder contains square brackets
         # and thus needs slicing/indexing e.g. `some_key[0:2, 0]`
         square_brackets, sub_remainder = _check_square_brackets(sub_remainder)
-        if not hasattr(value, sub_remainder):
-            raise ValueError(
-                "Attempting to access an non existing "
-                f"attribute {sub_remainder} of an object {value}"
-            )
-        value = getattr(value, sub_remainder)
+        if sub_remainder:
```
dbogunowicz, line 145: adapting for the use case where `target_identifier` is e.g. `engine_inputs[0]`, that means `sub_remainder = ''`

src/deepsparse/loggers/helpers.py
```diff
@@ -256,7 +257,8 @@ def check_identifier_match(
     if template == identifier:
         return True, None
     if template.startswith(identifier):
-        return True, template.replace(identifier, "")[1:]
+        remainder = template.replace(identifier, "")
+        return True, remainder if remainder.startswith("[") else remainder[1:]
```
dbogunowicz, line 261: adapting for the use case where `target_identifier` is e.g. `engine_inputs[0]`, that means we should not be removing the first char of a remainder (which is sometimes a coma, if the property of the target identifier present), but returning the full remainder

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -91,7 +91,9 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
 
         # needs to adhere to
         # https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
-        formatted_identifier = identifier.replace(".", "__")
+        formatted_identifier = (
```
dbogunowicz, line 94: Making new syntax of the identifiers match the Prometheus metric requirements

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -128,6 +130,8 @@ def _setup_client(self):
     def _validate(self, value: Any) -> Any:
         # make sure we are passing a value that is
         # a valid metric by prometheus client's standards
+        if hasattr(value, "is_integer"):
+            value = int(value)
```
dbogunowicz, line 134: so that we can parse out integers from numpy arrays

src/deepsparse/loggers/helpers.py
corey-nm: nice find, worth adding a test case for this?

corey-nm (dismissed):
lgtm, nice improvements

src/deepsparse/loggers/helpers.py
dbogunowicz: done

bfineran (approved):

PR #797, opened 2022-11-29
Changing license when build is enterprise

corey-nm:
When is_enterprise flag is set, the METADATA file now looks like:
```
Metadata-Version: 2.1
Name: deepsparse-ent
Version: 1.3.0
<snip>
License: Apache
<snip>
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,11 @@ _PACKAGE_NAME = (
 )
 
 if is_enterprise:
+    license = "Apache"
+else:
+    license = "Neural Magic DeepSparse Community License, Apache"
+
+if is_enterprise:
     # do not include the LICENSE-NEURALMAGIC file
     # in the deepsparse-ent installation folder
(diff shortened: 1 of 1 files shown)


dbogunowicz (approved):

setup.py
```diff
@@ -50,6 +50,11 @@
     else "deepsparse-nightly"
 )
 
+if is_enterprise:
+    license = "Apache"
+else:
+    license = "Neural Magic DeepSparse Community License, Apache"
+
 if is_enterprise:
```
dbogunowicz, line 58: nit: maybe merge with this if statement?

corey-nm: yeah and there's also the package name above that also checks this flag. i think in interest of not touching code that already works i won't merge the ifs. `setup.py` sketches me out too much haha

KSGulin (approved):

PR #802, opened 2022-12-05
[Fix] Prometheus Logger logging all zeros for latency metrics

dbogunowicz:
Reported by @rsnm2 : https://app.asana.com/0/1203146234476221/1203483881614817/f

Prometheus collects metrics in the form o of integers/floats and does not digest any complex data types such as lists, tuples, or arrays. 

Before I implemented a quick fix to account for values such as `np.array([3])`, s.t Prometheus can log it as `3`. 
However, since not only NumPy arrays have an attribute `is_integer` but python ints/floats, as well, the logic was rounding up all the values to the full integers.

I decided not to account for edge cases and not add any additional if statements to the `_validate` method. Accounting for different exceptions feels a bit hacky and dangerous. We should reserve `PrometheusLogger` for metrics (floats/ints) only, as this is an implicit assumption when one uses Prometheus. If the user wishes to log more complex data structures, they are free to use `CustomLogger` or one of the future Loggers we will introduce soon.

Ps. Also fixed here is a small unrelated bug (`MultiLogger` expects to receive a list of loggers, not a logger)

image

### DIFF
--- a/src/deepsparse/loggers/prometheus_logger.py
+++ b/src/deepsparse/loggers/prometheus_logger.py
@@ -130,8 +130,6 @@ class PrometheusLogger(BaseLogger):
     def _validate(self, value: Any) -> Any:
         # make sure we are passing a value that is
         # a valid metric by prometheus client's standards
-        if hasattr(value, "is_integer"):
-            value = int(value)
         if not isinstance(value, SUPPORTED_DATA_TYPES):
             raise ValueError(
                 "Prometheus logger expects the incoming values "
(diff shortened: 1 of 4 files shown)


src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -130,8 +130,6 @@ def _setup_client(self):
     def _validate(self, value: Any) -> Any:
         # make sure we are passing a value that is
         # a valid metric by prometheus client's standards
-        if hasattr(value, "is_integer"):
```
bfineran, line 133: any reason not to keep this but add a check for numpy?

dbogunowicz: I explain this in the PR:
**We should reserve PrometheusLogger for metrics (floats/ints) only, as this is an implicit assumption when one uses Prometheus. If the user wishes to log more complex data structures, they are free to use CustomLogger or one of the future Loggers we will introduce soon.**

bfineran (approved):

corey-nm (approved):

PR #807, opened 2022-12-07
[EmbeddingExtraction][fix] clear 'None' shapes from uninfered extracted onnx outputs

bfineran:
fixes an issue where extracted layers who were unable to be issued shape inference (any non trivial node) with the onnx shape inference API would have improper tensor type shapes set causing an unnecessary crash with the onnx checker

**test_plan:**
* unit tests passing
* manually verified fix across multiple domains and models


corey-nm (approved):

bfineran:
gha failures unrelated, merging

PR #809, opened 2022-12-07
[IC.eval] engine_type arg

bfineran:
**test_plan:**
manual


KSGulin (approved):

src/deepsparse/image_classification/validation_script.py
```diff
@@ -68,6 +72,9 @@
 from torch.utils.data import DataLoader
 
 
+DEEPSPARSE_ENGINE = "deepsparse"
+ORT_ENGINE = "onnxruntime"
```
corey-nm, line 76: FYI I think these exist in deepsparse somewhere IIRC

corey-nm (approved):

PR #818, opened 2022-12-15
[System Logging] UI

dbogunowicz:


src/deepsparse/server/system_logging.py
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from os import getpid
+from typing import Tuple
+
+import psutil
+
+
+def log_system_info(
```
bfineran, line 21: let's split the system logging UI and actual system log calls into separate PRs

src/deepsparse/server/build_logger.py
```diff
@@ -137,16 +136,51 @@ def build_function_loggers(
...
+            raise ValueError(
+                f"Unknown system logging group: {target_name}. Expected one of: {available_system_groups}"
+            )
+        target_logging_enabled = system_logging_config.get("enable", True)
+        if not target_logging_enabled:
+            continue
+
+        system_loggers.append(
+            _build_function_logger(
+                metric_function_cfg=MetricFunctionConfig(
```
bfineran, line 173: why do we need a function logger here? what does the logger hierarchy for these system logs look like now?

src/deepsparse/loggers/async_logger.py
```diff
@@ -43,19 +43,28 @@ def __init__(self, logger: BaseLogger, max_workers: int = 1):
         self.logger = logger
         self._job_pool: Executor = ThreadPoolExecutor(max_workers=max_workers)
 
-    def log(self, identifier: str, value: Any, category: MetricCategories):
+    def log(
+        self,
+        identifier: str,
+        value: Any,
+        category: MetricCategories,
+        group: Optional[SystemMetricGroups] = None,
```
bfineran, line 51: why do we need an explicit group? this also hardcodes an optional variable specifically for system logs which seems less than ideal. any reason to not log as a `system` log, with the identifier as something like `system/{GROUP}/name`?

bfineran (changes requested):
good start, let's see what we can do to whittle this down. specifically:

1. split out any actual system log impls from this PR
2. remove use of extra Group kwarg unless absolutely necessary
3. write down system logger hierarchy and see what we can do to whittle down
4. rework `create_system_loggers` so that there is a clear translation between the config and the produced loggers

PR #697, opened 2022-10-06
[Data Logging] Feature Branch

dbogunowicz:
# The Main Feature Branch for the Data Logging Functionality

### Support for two types of logging types

The incoming logs from the endpoint are grouped into two types: `Data` (pertaining to data logging) and `System` (pertaining 

### Added `BaseLogger` objects.

There are multiple logger objects at our disposal now (such as `PythonLogger`, `PrometheusLogger`, `MultiLogger`, `FunctionLogger` etc.) , that the server/user can compose together. The logger can be either passed directly to a pipeline as an argument (`Pipeline.create(..., pipeline = pipeline)`) or created from the `ServerConfig` in the DeepSparse Server context.

### Support `data_logging` property of the `EndpointConfig` in the server:

We can now pass the extended `ServerConfig`.

```yaml
num_cores: 2
num_workers: 2
system_logging: off # optional flag to turn off default logging group
# or to configure individual groups of metrics at system level:
system_logging:
  request_latency: on
  resources: off
loggers:
    - some_logger_1
    - some_logger_2
       key1: value1
       key2: value2
endpoints:
    - task: question_answering
      route: /unpruned/predict
      model: zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none
      batch_size: 1
      data_logging:
        pipeline_outputs.images[0]['key'][2:1,0]:
           - func: some_function_1 # from our built-ins
             frequency: 2
           - func: .../user_defined_metric_functions.py:some_function2
             frequency: 3
        pipeline_inputs:
           - func: torch.linalg.norm
             …
    - task: question_answering
      route: /unpruned/predict
      model: zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none
      batch_size: 1
      name: question_answering_
      data_logging:
        engine_outputs:
           - func: np.mean
             frequency: 4
```
That will enable data and/or system logging and enable the user to have granular control over the logging functionality.

PR #819, opened 2022-12-19
[System Logging] Logic for UI

dbogunowicz:
The main goal of this PR is to implement functionalities that support the digestion of the new element of the `ServerConfig`, the `SystemLoggingConfig`.

## Reading the system logging config

It enables the translation from a system logging config:
```yaml
system_logging:
  enable: true/false
  target_loggers:
    - python
    - prometheus
  prediction_latency:
        enabled: true
  resource_utilization_group:
         enabled: true
         target_loggers:
           - python
```
To a modified `server_logger` structure. This system config would add two additional function loggers to our overall loggers. We would translate all the logging configs into a 
following `server_logger`:
```python
server_logger = AsyncLogger(logger = MultiLogger(
    loggers = [
        FunctionLogger_1(...),
        FunctionLogger_2(...),
        ...
        FunctionLogger_n(...) # FunctionLoggers from 1 to n are about the data logging config
        FunctionLogger_n_+_1(...) # Pertains to prediction_latency logging (frequency = 1; target_identifier = "system/inference_latency"). Acts as a "filter" for all the inference latency logs
        FunctionLogger_n_+_2(...) # Pertains to resource logging (frequency = 1; target_identifier = "system/resource_utilization"). Acts as a "filter" for all the resource logs
]))
```

## Introducing `SystemLoggingManager`

I introduce the `SystemLoggingManager` object that is responsible for managing the system logging.


src/deepsparse/server/server.py
```diff
@@ -121,6 +121,7 @@ def _build_app(server_config: ServerConfig) -> FastAPI:
 
     app = FastAPI()
     server_logger = build_logger(server_config)
+    system_logging_manager = SystemLoggingManager.from_server_config(server_config)
```
dbogunowicz, line 124: This is WIP to make the tests pass. I will focus more on E2E testing in the next PRs.

src/deepsparse/server/build_logger.py
```diff
@@ -190,7 +237,32 @@ def _build_custom_logger(logger_arguments: Dict[str, Any]) -> BaseLogger:
...
+        if len(identifier_strings) != 2:
+            raise ValueError(
+                "Target identifier must be a tuple of two strings. "
+                f"Got {identifier_strings} instead."
+            )
+        return _get_target_identifier_data(*identifier_strings)
+
+
+def _get_target_identifier_system(group_name: str) -> str:
+    # TODO: Let's discuss how the support for regex would
```
dbogunowicz, line 260: @bfineran let's discuss in more detail this feature.

src/deepsparse/loggers/helpers.py
```diff
@@ -250,10 +253,12 @@ def check_identifier_match(
         pattern = template[3:]
         return re.match(pattern, identifier) is not None, None
     if template.startswith("category:") and category is not None:
-        template_category = template[9:]
+        template_identifier = template[9:]
+        template_category, *remainder = template_identifier.split("/")
         template_category = MetricCategories(template_category)  # parse into Enum
         category = MetricCategories(category)  # ensure in Enum form
-        return template_category == category, None
+        remainder = None if not remainder else "/".join(remainder)
```
dbogunowicz, line 260: I feel like the structure of those identifiers is quickly getting out of hand. For 1.4 I feel we are in a good state, but before we start introducing complexity for the next release, I would focus on a concrete vision of how the identifiers should look like, what are the delimiters, what is their meaning etc.

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+    enable: Optional[bool] = Field(
+        default=True, description="Whether to enable system logging"
+    )
+    target_loggers: Optional[List[str]] = Field(
+        default=None,
+        description="The list of loggers to log to. If None, "
+        "logs to all the available loggers.Will "
+        "be overriden by the target_loggers specified "
+        "in the individual system logging groups.",
+    )
```
corey-nm, line 83: What about this?
```suggestion
    # Global System Logging
    system: SystemLoggingGroup = Field(..., default=SystemLoggingGroup(enable=True))
```

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+
+    # System Logging Groups
+    resource_utilization: Optional[SystemLoggingGroup] = Field(
+        default=None,
```
corey-nm, line 87: Could probably make them not optional and make default explicitly enable=False for all these groups:
```suggestion
    resource_utilization: SystemLoggingGroup = Field(
        default=SystemLoggingGroup(enable=False),
```

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+    )
+
+
+class SystemLoggingConfig(BaseModel):
+    """
+    Holds the configuration for the system logging
+    """
+
+    # Global System Logging
+    enable: Optional[bool] = Field(
```
corey-nm, line 74:
```suggestion
    enable: bool = Field(
```

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+        "refer to the DeepSparse server system logging "
+        "documentation. By default this group is disabled.",
+    )
+    prediction_latency: SystemLoggingGroup = Field(
+        default=None,
+        description="The configuration group for the prediction latency "
+        "system logging group. For details "
+        "refer to the DeepSparse server system logging "
+        "documentation. By default this group is disabled.",
+    )
```
corey-nm, line 99: Could this be handled on the pipeline logger config level?

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+
+    enable: bool = Field(
+        default=False, description="Whether to enable the system logging group"
+    )
+
+    target_loggers: Optional[List[str]] = Field(
+        default=None,
+        description="The list of target loggers to log to. "
+        "If None, logs to all the available loggers",
+    )
```
corey-nm, line 65: i guess someone could disable by setting target_loggers to empty list right? think its easier to understand how to configure by removing enable field?
```suggestion
    target_loggers: Optional[List[str]] = Field(
        default=None,
        description="The list of target loggers to log to. "
        "If None, logs to all the available loggers",
    )
```

src/deepsparse/server/system_logging.py
```diff
@@ -0,0 +1,121 @@
...
+    def log_request_details(self, pipeline: Pipeline, **kwargs: Dict[str, Any]):
+        """
+        Scope for 1.4:
+        - Number of Successful Requests
+        (binary events, 1 or 0 per invocation of an endpoint)
+        - Batch size
+        - Number of Inferences (
+        number of successful inferences times the respective batch size)
+        """
+        raise NotImplementedError
```
corey-nm, line 121: i think this set of changes would be a lot simpler if we just explicitly put these functions in the server.py file as standalone functions/methods and don't have this additional object.

e.g.
```python
def _predict(...):
    log_request_details(...)
    log_resource_utilization(...)
   ...
```

src/deepsparse/loggers/helpers.py
bfineran: yes, this sounds good. agreed on getting this simplified

src/deepsparse/server/build_logger.py
```diff
@@ -111,42 +110,90 @@ def build_leaf_loggers(
...
+    # extract exclusively the system logging groups data
+    system_config_groups = {
+        config_group_name: config_group_args
+        for (config_group_name, config_group_args) in system_logging_config
+        if isinstance(config_group_args, SystemLoggingGroup)
+    }
+    # propagate the global target loggers information if applicable
+    for config_group_name, config_group_args in system_config_groups.items():
+        if target_loggers and not config_group_args.target_loggers:
+            config_group_args.target_loggers = target_loggers
```
bfineran, line 194: this code could read a bit better as a single for loop since we are creating a dictionary and then immediately mutating it

src/deepsparse/server/build_logger.py
```diff
@@ -111,42 +110,90 @@ def build_leaf_loggers(
...
+        to its arguments
+    """
+    if not system_logging_config.enable:
+        return {}
+    target_loggers = system_logging_config.target_loggers
+
+    # extract exclusively the system logging groups data
+    system_config_groups = {
+        config_group_name: config_group_args
+        for (config_group_name, config_group_args) in system_logging_config
```
bfineran, line 188: are we missing something in the iteration here?

src/deepsparse/server/build_logger.py
bfineran: why do we need to prepend with `category`? instead of something like `*./group_name/.*`?

src/deepsparse/server/build_logger.py
```diff
@@ -190,7 +237,32 @@ def _build_custom_logger(logger_arguments: Dict[str, Any]) -> BaseLogger:
     return logger
 
 
-def _get_target_identifier(endpoint_name: str, target_name: str) -> str:
+def _get_target_identifier(
+    *identifier_strings, is_system_logging_identifier=False
+) -> str:
+    if is_system_logging_identifier:
```
bfineran, line 243: why bake this all into one function with an optional bool flag that does completely different functionality? seems like we can just point build system loggers directly to `_get_target_identifier_system`

src/deepsparse/server/config.py
bfineran: I'm in favor of having the explicit enable field to remove any ambiguity between defaulting to all loggers and `None` vs empty list

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+
+
+class SystemLoggingConfig(BaseModel):
+    """
+    Holds the configuration for the system logging
+    """
+
+    # Global System Logging
+    enable: Optional[bool] = Field(
+        default=True, description="Whether to enable system logging"
```
bfineran, line 75: nit: add `"defaults to True`"

src/deepsparse/server/config.py
bfineran: I think having target loggers at this level is overkill here. we can just delegate the user to the global target loggers and then can filter in the individual group configs. no need IMO for 3 levels of filtering target loggers. clutters config and may lead to confusion on hierarchy.

src/deepsparse/server/config.py
bfineran: +1, will make handling on our end more straightforward.  Additionally, why do we default system logging to True, but then disable all of the groups?

src/deepsparse/server/config.py
bfineran: latency was agreed to be a "system" level log, so should fall in this config. afair there isn't a way to disable these right now

src/deepsparse/server/config.py
```diff
@@ -49,6 +49,63 @@ class ImageSizesConfig(BaseModel):
...
+    # System Logging Groups
+    resource_utilization: Optional[SystemLoggingGroup] = Field(
+        default=None,
+        description="The configuration group for the resource "
+        "utilization system logging group. For details "
+        "refer to the DeepSparse server system logging "
+        "documentation. By default this group is disabled.",
+    )
+    prediction_latency: SystemLoggingGroup = Field(
+        default=None,
```
bfineran, line 94: let's default these to enabled (matches current functionality as well)

src/deepsparse/server/system_logging.py
bfineran: agreed overall. I like the structure here @dbogunowicz, but looking through it looks like the class isn't keeping much state so could be restructured to a helper function

tests/server/test_build_logger.py
```diff
@@ -167,14 +168,110 @@ def test_build_logger(yaml_config, raises_error, default_logger, num_function_lo
...
+        assert isinstance(logger.logger.loggers[0].logger.loggers[0], PythonLogger)
         return
     assert len(logger.logger.loggers) == num_function_loggers + 1
 
-    # check for system logger
+    # check for default system logger behaviour
     system_logger = logger.logger.loggers[-1]
     assert system_logger.target_identifier == (
-        f"category:{MetricCategories.SYSTEM.value}"
+        f"category:{MetricCategories.SYSTEM.value}/prediction_latency"
```
bfineran, line 178: do the pipeline logs all have `category:` prepended to them? where does this happen again?

bfineran (changes requested):

src/deepsparse/server/config.py
dbogunowicz: Fair point, I agree.

src/deepsparse/server/config.py
dbogunowicz: To enforce some defaults (I imagined prediction_latency would be a default). For the sake of this PR I would refrain from setting any defaults, they can come later in a form of a very short PR.

src/deepsparse/server/config.py
dbogunowicz: Well, depends what you mean by disable. 
We are unable to disable prediction latency generation (this happens on the pipeline level), but we are able to disable it on the logger level (just refrain from creating the FunctionLogger with appropriate identifier)

src/deepsparse/server/config.py
dbogunowicz: I like it

src/deepsparse/server/system_logging.py
dbogunowicz: Yes, this is true for now (1.4 release). To obey the KISS principle for this release, I would employ helper function. If the scope of system logging potentially increases in the future and we need some more complexity, we can always default to a class object.

src/deepsparse/server/build_logger.py
dbogunowicz: I think this was initially introduced by you here: https://github.com/neuralmagic/deepsparse/pull/697/commits/780b38aa685a33d934a82a36dc641d8f2ab91db1

My suggestion is, let's land this PR ignoring the identifier convention, this will be something I will clean up in the scope of 1.4. in a separate PR

tests/server/test_build_logger.py
dbogunowicz: See the comment above, introduced in https://github.com/neuralmagic/deepsparse/pull/697/commits/780b38aa685a33d934a82a36dc641d8f2ab91db1

corey-nm (approved):
lgtm, nice work!

src/deepsparse/server/build_logger.py
```diff
@@ -16,24 +16,30 @@
...
-from deepsparse.server.helpers import custom_logger_from_identifier, default_logger
+from deepsparse.server.config import (
+    MetricFunctionConfig,
+    ServerConfig,
+    SystemLoggingConfig,
+    SystemLoggingGroup,
+)
 
 
 __all__ = ["build_logger"]
```
bfineran, line 40: this list needs updating

src/deepsparse/server/build_logger.py
```diff
@@ -111,42 +111,93 @@ def build_leaf_loggers(
...
+            continue
+
+        system_loggers.append(
+            _build_function_logger(
+                metric_function_cfg=MetricFunctionConfig(
+                    func="identity",
+                    frequency=1,
+                    target_loggers=config_group_args.target_loggers,
+                ),
+                target_identifier=_get_target_identifier(config_group_name),
```
bfineran, line 171: what does `_get_target_identifier` do here?
and don't we need to update the identifier to match to all logged values containing the config group name?

bfineran (changes requested):
great improvements to this PR - lgtm pending comments

src/deepsparse/pipeline.py
```diff
@@ -700,15 +704,24 @@ def to_config(self) -> "PipelineConfig":
             kwargs=kwargs,
         )
 
-    def log(self, identifier: str, value: Any, category: str):
+    def log(
+        self,
+        identifier: str,
+        value: Any,
+        category: str,
+        identifier_pre_fix: Optional[str] = None,
```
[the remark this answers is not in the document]
dbogunowicz: This would probably not solve our problem.
Let me illustrate why, by giving a short example.

Let's say that we are logging the prediction latency ( of the `pre_process` step) of a pipeline called `sample_pipeline`.

Right now, with the simplified design of identifiers and "identifier pre-fix", the pipeline would emit an identifier: `prediction_latency/sample_pipeline/pre_process`. The function logger that is "looking for" this identifier would use the target identifier `prediction_latency` and simply check for the match between the first components to the target identifier and identifier strings.

Without the identifier pre-fix, we are losing this neat matching functionality. If the pipeline were to emit an identifier `sample_pipeline/prediction_latency/pre_process` (this would happen if I followed your advice), matching would get quite complex. Especially in the situation where our pipeline becomes complex/nested. 

We can try to make it work, but it would significantly complicate the matching logic. And I think the matching logic is something that we want to keep extremely simple, otherwise, it may get unnecessarily complex quickly.

But I am open to discussing this further.

src/deepsparse/server/build_logger.py
dbogunowicz: In this case, it acts as an "identity" function. But I wanted to keep it here in case we were to add functionalities in the future (e.g. this function handles making sure that regex identifiers will not be modified).
There was no need to update the identifier while the `identifier_pre_fix` was present in the `pipeline`. 

Let's sync offline and discuss, I think this is an important point we need to align on.

bfineran (approved):

PR #824, opened 2022-12-20
[System Logging] Log resource utilization

dbogunowicz:
Following the PRD:
image
This PR implements all four aforementioned logging information.


tests/server/test_system_logging.py
```diff
@@ -0,0 +1,56 @@
...
+def _test_total_memory_available(calls, num_iterations):
+    relevant_calls = [
+        call
+        for call in calls
+        if call.startswith("identifier:total_memory_available_[MB]")
+    ]
+    values = [float(call.split("value:")[1].split(",")[0]) for call in relevant_calls]
+    assert len(relevant_calls) == num_iterations
+    # assert all values are the same (total memory available is constant)
+    assert all(value == values[0] for value in values)
```
rahul-tuli, line 45: nit: maybe use `len(set(values)) < 2`, would protect from condition when values is empty, however cause this is a test, I'm good with this idiom

rahul-tuli (approved):
lgtm!

src/deepsparse/server/system_logging.py
```diff
@@ -12,24 +12,47 @@
...
+    Logs the resource utilization of the server process.
+    This includes:
+    - CPU utilization
+    - Memory utilization
+    - Total memory available
+
+    :param server_logger: the logger to log the metrics to
     """
-    pass
+    process = psutil.Process(getpid())
```
corey-nm, line 36: is this already in our setup.py dependencies?

corey-nm (approved):
nice!!

src/deepsparse/server/system_logging.py
dbogunowicz: haha, no. good catch.

src/deepsparse/server/system_logging.py
```diff
@@ -12,24 +12,60 @@
...
+
+    identifier_to_value = {
+        "cpu_utilization_[%]": cpu_percent,
+        "memory_utilization_[%]": memory_percent,
+        "total_memory_available_[MB]": total_memory_megabytes,
+    }
+    if kwargs:
+        identifier_to_value.update(kwargs)
+
+    for identifier, value in identifier_to_value.items():
```
bfineran, line 63: looks like we can share the helper method in #831 here

src/deepsparse/server/server.py
```diff
@@ -242,8 +245,9 @@ def _add_pipeline_endpoint(
 
     def _predict(request: pipeline.input_schema):
         pipeline_outputs = pipeline(request)
-        if pipeline.logger:
-            log_resource_utilization(pipeline)
+        server_logger = pipeline.logger
+        if server_logger:
```
bfineran, line 249: how would a user disable this? shouldn't this be controlled by the config?

src/deepsparse/server/system_logging.py
```diff
@@ -12,24 +12,60 @@
...
 from typing import Any
 
+import psutil
 from deepsparse import Pipeline
+from deepsparse.loggers import BaseLogger, MetricCategories
 
 
 __all__ = ["log_resource_utilization", "log_request_details"]
 
+RESOURCE_UTILIZATION_IDENTIFIER_PREFIX = "resource_utilization/"
```
bfineran, line 25: same as #831, let's move these out to a `logging/contants.py` file

src/deepsparse/server/system_logging.py
```diff
@@ -12,24 +12,60 @@
...
+    # A float representing process memory utilization as a percentage
+    memory_percent = process.memory_percent()
+    # Total physical memory
+    total_memory_bytes = psutil.virtual_memory().total
+    total_memory_megabytes = total_memory_bytes / 1024 / 1024
+
+    identifier_to_value = {
+        "cpu_utilization_[%]": cpu_percent,
+        "memory_utilization_[%]": memory_percent,
+        "total_memory_available_[MB]": total_memory_megabytes,
```
bfineran, line 58: let's avoid using symbols as words and also reserve brackets in identifiers for slicing
```suggestion
        "cpu_utilization_percent": cpu_percent,
        "memory_utilization_percent": memory_percent,
        "total_memory_available_MB": total_memory_megabytes,
```

bfineran (changes requested):
log implementation looks great - main question is how this fits in with the config/UX

src/deepsparse/server/system_logging.py
dbogunowicz: I am handling this edge case in the code, just fyi.

src/deepsparse/server/server.py
dbogunowicz: Good point. After removing the SystemLoggingManager object I did neglect this detail.

dbogunowicz: I probably have made an erroneous implicit assumption. When we decided early on to move from an object that will control all sorts of system logging activities (I called it `SystemLoggingManager`) in favor of separate, simple functions, I explicitly assumed that this would also mean that for now the only instance that enables/disables the logging is the presence or absence of the appropriate function logger. Not the most scalable solution, but sufficient for now. Let's sync on that as well.

src/deepsparse/server/server.py
```diff
@@ -242,8 +245,9 @@ def _add_pipeline_endpoint(
 
     def _predict(request: pipeline.input_schema):
         pipeline_outputs = pipeline(request)
-        if pipeline.logger:
-            log_resource_utilization(pipeline)
+        server_logger = pipeline.logger
```
dbogunowicz, line 248: callback -> attach property to server_logger, lambda function on the side in the callback we know what to log.

src/deepsparse/server/system_logging.py
```diff
@@ -42,3 +83,25 @@ def log_request_details(pipeline: Pipeline, **kwargs: Any):
     number of successful inferences times the respective batch size)
     """
     pass
+
+
+def _logging_enabled(server_logger: BaseLogger, group_name: str) -> bool:
```
dbogunowicz, line 88: @bfineran what do you think about this private function for enabling and disabling system metric group logging? Maybe we could move this helper function to "higher up" to the server level?

bfineran: In general, we should avoid functions that take the generic logger expecting server structure. Greatly prefer to read the config originally and keep that info in memory so the config is single source of truth

bfineran (approved):
LGTM. Let's investigate performance implications of adding the middleware/log calls

PR #826, opened 2022-12-20
[System Logging] Clean up identifier design

dbogunowicz:
Initially, I am proposing a new design. Let's sign off on this before I start engineering work (should be very minimal once we agree on the design). 

In the nutshell:
"/" - symbol to join all the components of the identifiers (except for the ones mentioned below):
"." - symbol to join the properties in the DATA category
"__" - symbol to join the function_name in the DATA category.

Simplify the matching of the SYSTEM category logs - remove the "category:" prefix.


src/deepsparse/loggers/PROPOSAL.md
```diff
@@ -0,0 +1,105 @@
...
+## For Data Logging
+In the `build_logger` function, we create the target_identifier by concatenating the pipeline_identifier and the main_identifier.
+```python
+target_identifier = f"{endpoint_name}/{target_name}" # if not regex
+target_identifier = target # if regex
+```
+Since target_identifier creation follows the same logic as the identifier creation for the pipeline, the matching is straightforward.
+
+## For System Logging
+I think we should follow a similar as the data logging, and aim for 1:1 matching between the identifier and the target_identifier.
```
dbogunowicz, line 50:
```suggestion
I think we should follow a similar logic as in the data logging, and aim for 1:1 matching between the identifier and the target_identifier.
```

src/deepsparse/loggers/PROPOSAL.md
```diff
@@ -0,0 +1,107 @@
...
+    return True, remainder if remainder.startswith("[") else remainder[1:]
+
+return False, None
+```
+
+# On the output from the FunctionLogger
+- if logging belongs to the DATA category and the remainder has slicing/indexing/access information, 
+  remove it e.g. "some_value[0, 1:2]" -> "some_value". Then concatenate it with the identifier using ".": "identifier.some_value"
+- when the function is applied to the value, concatenate it with the identifier using "_": "identifier.some_value_function_name". 
+  If the identifier belongs to the SYSTEM category (function is identity function), skip this step.
```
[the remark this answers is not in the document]
dbogunowicz: errmm, so that this is logged:
`image_classification/pipeline_inputs.images__mean_pixel_value
but not this:
`image_classification/pre_processing__identity` 
but this:
`image_classification/pre_processing`

Basically we assume that all system data has identity being applied to it, so we do not include the func name in the identifier.

corey-nm (dismissed):
sounds good to me, looks like it'll simplify things a lot so i'm on board! nice write up

tests/server/test_build_logger.py
```diff
@@ -218,22 +216,22 @@ def test_build_logger(yaml_config, raises_error, default_logger, num_function_lo
...
+                "prediction_latency",
+                "resource_utilization",
             },
             [2, 2],
         ),
         (yaml_config_4, set(), []),
         (
             yaml_config_5,
             {
-                "category:system/prediction_latency",
```
dbogunowicz, line 235: Mental note: add regex test case!

corey-nm (approved):
lgtm, definitely a good simplification. left one suggestion for simplifying pipeline log call

src/deepsparse/pipeline.py
```diff
@@ -277,11 +279,13 @@ def __call__(self, *args, monitoring: bool = False, **kwargs) -> BaseModel:
             identifier=InferencePhases.POST_PROCESS,
             value=timer.time_delta(InferencePhases.POST_PROCESS),
             category=MetricCategories.SYSTEM,
+            identifier_pre_fix="prediction_latency",
```
dbogunowicz, line 282: IMO the most controversial thing in this PR. Willing to defend this design, but am aware of the ugliness.

dbogunowicz: This enables us to match system logs in the following way:
e.g.
identifier = `prediction_latency/pipeline_name/pre_processing`
vs
target_identifier = `prediction_latency` (from the system logging config)

src/deepsparse/loggers/helpers.py
```diff
@@ -31,11 +31,42 @@
...
 
 NO_MATCH = "NO_MATCH"
 
 
+def finalize_identifier(
+    identifier: str,
+    category: MetricCategories,
+    function_name: str,
+    remainder: Optional[str] = None,
+):
```
dbogunowicz, line 45:
```suggestion
) -> str:
```

bfineran (approved):

Issue #825, opened 2022-12-20
Broken Transformers QA Inference Pipeline


dbogunowicz:
**Describe the bug**

Transformers QA pipeline fails on a simple inference task.

**Expected behavior**
The inference pipeline for Question Answering should work without raising any errors.

**Environment**
Python version: 3.8
DeepSparse version: current main

**To Reproduce**

```python
from deepsparse import Pipeline

task = "question-answering"
dense_qa_pipeline = Pipeline.create(
        task=task,
        model_path="zoo:nlp/question_answering/distilbert-none/pytorch/huggingface/squad/base-none",
        # or model_path = "zoo:nlp/question_answering/bert-base/pytorch/huggingface/squad/base-none",
        # was checking whether the problem is not model-dependent
    )

question = "DeepSparse is sparsity-aware inference runtime offering GPU-class performance on CPUs and APIs to integrate ML into your application"
q_context = "What is DeepSparse?"

dense_qa_pipeline(question=question, context=q_context)
```

**Errors**
```bash
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
DeepSparse, Copyright 2021-present / Neuralmagic, Inc. version: 1.3.0.20221217 COMMUNITY | (d5bf112b) (release) (optimized) (system=avx2, binary=avx2)
Traceback (most recent call last):
  File "/usr/lib/python3.8/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/home/ubuntu/.pycharm_helpers/pydev/_pydev_bundle/pydev_umd.py", line 198, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "/home/ubuntu/.pycharm_helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/ubuntu/damian/deepsparse_copy/hehe.py", line 14, in <module>
    dense_output = dense_qa_pipeline(question=question, context=q_context)
  File "/home/ubuntu/damian/deepsparse_copy/src/deepsparse/pipeline.py", line 217, in __call__
    engine_inputs: List[numpy.ndarray] = self.process_inputs(pipeline_inputs)
[...]

PR #827, opened 2022-12-20
[Fix] Update the code for handling ragged numpy arrays in numpy >= 1.24.0

dbogunowicz:
Response to: https://github.com/neuralmagic/deepsparse/issues/825

Since NumPy version 19.0, one must specify dtype=object when creating an array from "ragged" sequences, otherwise, one receives a warning:

```bash
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
```
Starting NumpyPy version 1.24.0, this warning turns into an explicit error:

```bash
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (128,) + inhomogeneous part.
```

This PR adds the code change necessary to remove the error. Afaik, only transformers QA code requires the update.

This fix should also be backward compatible at least dating back to 1.19 (in our requirements we honor numpy>=1.16.3, maybe it is time to bump this version up?)


dbogunowicz:
also updated numpy version in the setup.py, tested with the 1.17 and it was too old (gave me a bunch of errors that indicated that it was too old for our requirements)

setup.py
```diff
@@ -85,7 +85,7 @@ def _parse_requirements_file(file_path):
 
 
 _deps = [
-    "numpy>=1.16.3",
+    "numpy>=1.19.0",
```
dbogunowicz, line 88:
```suggestion
    "numpy>=1.16.3",
```

corey-nm (dismissed):
such a beautiful diff 🥇

src/deepsparse/transformers/pipelines/token_classification.py
```diff
@@ -367,7 +367,7 @@ def route_input_to_bucket(
             input_schema.inputs,
             add_special_tokens=True,
             return_tensors="np",
-            padding=False,
+            padding=True,
```
bfineran, line 370: why are we adding padding here? if there's padding we cant extract the true sequence length, no?

bfineran (changes requested):
is this change backwards compatible?

src/deepsparse/transformers/pipelines/token_classification.py
dbogunowicz: True. Let me further investigate the optimal solution to the problem, I have to admit that I was just quickly taking a look at the problem as a "side-project" and found out that more work is needed to understand it and to get the right solution.

dbogunowicz:
Simple code to reproduce the main issue that blocks us from landing this patch:
```python
from transformers import AutoTokenizer

sequences = ['a ' * 10, 'a ' * 35]
tokenizer = AutoTokenizer.from_pretrained('/home/ubuntu/.cache/sparsezoo/59081a43-86c9-441c-980d-83c23dcd8de5/deployment')
tokens = tokenizer(
    sequences,
    add_special_tokens=True,
    return_tensors="np",
    padding=False,
    truncation=False,
)
```
```bash
ValueError: Unable to create tensor, you should probably activate truncation and/or padding with 'padding=True' 'truncation=True' to have batched tensors with the same length. Perhaps your features (`input_ids` in this case) have excessive nesting (inputs type `list` where type `int` is expected).
```

PR #841, opened 2023-01-03
[System Logging] Few fixes post testing

dbogunowicz:


src/deepsparse/pipeline.py
```diff
@@ -192,6 +193,12 @@ def __init__(
 
         self._batch_size = self._batch_size or 1
 
+        self.log(
```
bfineran, line 196: Hm now that I see this here, not sure if this is necessary since cores is static relative to a config change

src/deepsparse/server/system_logging.py
```diff
@@ -131,7 +136,7 @@ def log_resource_utilization(
 
 def log_request_details(
     server_logger: BaseLogger,
-    prefix: str,
+    prefix: str = REQUEST_DETAILS_IDENTIFIER_PREFIX,
```
bfineran, line 139: Do we want prefix only or should this be "group" and match if the value is *anywhere* in the identifier

bfineran (approved):

PR #842, opened 2023-01-03
Dynamic Engine Load

rahul-tuli:
This PR will add support to load a Engine Dynamically from a user specified python script

In the `benchmark_sweep` and `deepsparse.benchmark` utility now users can specify a custom engine class, that will be dynamically loaded during runtime. A possible use case could be to benchmark different Inference backends.

Note: If the graph is not in `onnx` format then the engine classes must specify a function `generate_random_inputs` that takes in a batch size and returns inputs for the model in an ingestable format. A better solution could be to force this contract by creating a base class. But have left this as is for now

### DIFF
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -320,11 +320,11 @@ def load_custom_engine(custom_engine_identifier: str):
            '<path_to_the_python_script>:<custom_engine_class_name>
     :return: custom engine class object
     """
-    path, logger_object_name = custom_engine_identifier.split(":")
+    path, engine_object_name = custom_engine_identifier.split(":")
     spec = importlib.util.spec_from_file_location("user_defined_custom_engine", path)
     module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(module)
-    return getattr(module, logger_object_name)
+    return getattr(module, engine_object_name)
 
 
 def benchmark_model(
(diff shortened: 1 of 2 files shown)


corey-nm (approved):
lgtm, may be worth just adding tfline engine to deepsparse

src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -308,6 +311,22 @@ def parse_num_streams(num_streams: int, num_cores: int, scenario: str):
...
+
+    :param custom_engine_identifier: string in the form of
+           '<path_to_the_python_script>:<custom_engine_class_name>
+    :return: custom engine class object
+    """
+    path, logger_object_name = custom_engine_identifier.split(":")
+    spec = importlib.util.spec_from_file_location("user_defined_custom_engine", path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return getattr(module, logger_object_name)
```
bfineran, line 327:
```suggestion
    path, engine_object_name = custom_engine_identifier.split(":")
    spec = importlib.util.spec_from_file_location("user_defined_custom_engine", path)
    module = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(module)
    return getattr(module, engine_object_name)
```

KSGulin (approved):
LGTM, pending incorporation of Ben's suggestion

PR #849, opened 2023-01-05
Use scikit-learn instead of sklearn

corey-nm:
Name of package was changed. Other places in the repo have already been updated


rahul-tuli (approved):

dbogunowicz (approved):
easy

PR #856, opened 2023-01-09
[MLOps 1.5] Pipeline Logging

dbogunowicz:
# Pipeline Logging Feature

## Feature preview

```python
import yaml
from deepsparse import logger_from_config
from deepsparse.loggers.config import PipelineLoggingConfig

config = """
loggers:
    python:
system_logging:
    resource_usage:
        enable: true
data_logging:
    pipeline_outputs:
       - func: identity
         frequency: 2
    engine_outputs:
       - func: np.mean
         frequency: 3"""

pipeline = Pipeline.create("token_classification", logger = config
pipeline.logger
>> <deepsparse.loggers.async_logger.AsyncLogger object at 0x7f2629ae96d0>
```

## What's new:
- Moved the logic for building composed loggers from `src/deepsparse/server/build_logger.py` to `src/deepsparse/loggers/build_logger.py`. The motivation is to uncouple as much of the logging code from the server code. This is also a motivation for moving relevant Pydantic models from `/server/config.py` to `/loggers/config.py`

- Added `logger_from_config` method, which takes a path to the config on disk/ yaml config string and returns a pipeline logger.

- Added the `server_logger_from_config` method, which takes a `ServerLoggingConfig` and returns the server logger. It substitutes the function known previously as `build_loggers`. `build_loggers` now is a "lower level" function, that is being called by both `logger_from_config` and `server_logger_from_config`.

- `Pipeline` objects now can take logger argument s.t. `logger: Union[BaseLogger, str (path), str (yaml config)]`

### Assumptions on System Logging for Pipelines/Server:
The pipeline will log system metric groups: `prediction_latency` and `resource_utilization`.
The Server will additionally log `request_details`.
I think this is consistent with the user's expectations.

### Testing 
Adjusted the `loggers` and server `unittests` to reflect the changes. Added one end-to-end test: `tests/deepsparse/loggers/test_end_to_end.py`. Also, rudimentary manual testing. All tests are green.


src/deepsparse/pipeline.py
```diff
@@ -257,7 +257,7 @@ def __call__(self, *args, **kwargs) -> BaseModel:
         # join together the batches of size `self._batch_size`
         engine_outputs = self.join_engine_outputs(batch_outputs)
         timer.stop(InferencePhases.ENGINE_FORWARD)
-
+        # let's put this into a separate metric group!
         self.log(
             identifier=f"{REQUEST_DETAILS_IDENTIFIER_PREFIX}/input_batch_size",
```
dbogunowicz, line 262: Let's not log this value under "request details" -> let's uncouple it from the server context. Maybe we can have something like `pipeline properties`?

tests/deepsparse/loggers/test_function_logger.py
```diff
@@ -46,7 +47,9 @@ def _return_number(x):
         ),
     ],
 )
+@mock_engine(rng_seed=0)
```
dbogunowicz, line 50: Just to speed up those tests.

src/deepsparse/loggers/helpers.py
```diff
@@ -285,13 +280,39 @@ def check_identifier_match(
...
-    if template in identifier:
-        return True, None
+
+    match = SequenceMatcher(None, identifier, template).find_longest_match(
+        0, len(identifier), 0, len(template)
+    )
+    if not match:
+        return False, None
+
+    if match.b == 0:
```
dbogunowicz, line 290: Needed to refactor here to account for the fact, that when we create a logger the pipeline is not yet existing. This means that we need to able to match identifiers against templates that do not have pipeline name e.g:
template: `something.sequence[0]`
identifier: `pipeline_name/something`
should result in match and reminder `sequence[0]`

bfineran (changes requested):
fantastic job @dbogunowicz :)
comments mostly on documentation and then suggestion on how to clean up flow / fix issue regarding pipeline name/construction order.

In the future - I strongly recommend splitting this PR up into 2. First one would be to migrate logger creation utils to the `loggers` submodule. second would be to update the flow to work with pipeline objects

src/deepsparse/loggers/helpers.py
dbogunowicz: Why not keep this solution? It is not cluttering the code or its readability. On the other hand, the solutions that you are proposing are quite elegant, but make the process more complicated for the user.

src/deepsparse/loggers/build_logger.py
```diff
@@ -67,32 +76,56 @@ def default_logger() -> Dict[str, BaseLogger]:
...
+
+    :param config: The configuration of the pipeline logger.
+        Can be:
+         - a string (path to the .yaml file)
+         - a string (string representation of the .yaml file)
+         - a PipelineLoggingConfig instance
+    :return: The pipeline with the logger added
+    """
+    if not isinstance(config, str):
+        pass
```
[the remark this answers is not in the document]
corey-nm: This is just converting the str possibility into a config, it could just be:
```python
if isinstance(config, str):
    ...
    config = PipelineLoggingConfig(...)
```

src/deepsparse/server/config.py
```diff
@@ -250,8 +179,8 @@ class ServerConfig(BaseModel):
         ),
     )
 
-    system_logging: SystemLoggingConfig = Field(
-        default=SystemLoggingConfig(),
+    system_logging: ServerSystemLoggingConfig = Field(
+        default=ServerSystemLoggingConfig(),
```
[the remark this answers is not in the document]
corey-nm: I don't think this is true for pydantic models, it seems like they copy defaults:
```python
>>> class Mine(pydantic.BaseModel):
...     a: dict = pydantic.Field(default={"a": 1})
... 
>>> q = Mine()
>>> q
Mine(a={'a': 1})
>>> q.a["a"] = 2
>>> q
Mine(a={'a': 2})
>>> r = Mine()
>>> r
Mine(a={'a': 1})
```

corey-nm (approved):
lgtm, nice improvements!

src/deepsparse/loggers/build_logger.py
```diff
@@ -67,32 +76,56 @@ def default_logger() -> Dict[str, BaseLogger]:
...
-    The process follows the following hierarchy:
-
-    First: if global logger config is provided,
-        the "leaf" loggers are built.
-
-    Second: if data logging config is specified, a set of
-        function loggers wraps around the appropriate "leaf" loggers.
+     Builds a pipeline logger from the ServerConfig
+
+    :param config: The configuration of the pipeline logger.
```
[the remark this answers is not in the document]
dbogunowicz: Did it in the `build_logger` function, since users that are both coming from server and pipeline logger build functions will come across this pydoc

corey-nm (approved):
lgtm, nice improvements

src/deepsparse/loggers/build_logger.py
```diff
@@ -204,6 +293,36 @@ def build_system_loggers(
...
+
+    for target_identifier, metric_functions in data_logging_config.copy().items():
+        if not target_identifier.startswith(pipeline_identifier):
+            # if the target identifier does not already start
+            # with the pipeline identifier, call get_target_identifier
+            # to prepend it
+            target_identifier = get_target_identifier(
+                target_identifier, pipeline_identifier
+            )
+        data_logging_config[target_identifier] = metric_functions
```
dbogunowicz, line 322: This is a buggy implementation (will log both `some_target` and `pipeline/some_target`. Will fix it asap.

bfineran (approved):

bfineran:
lgtm pending issues

corey-nm (approved):

bfineran (approved):

PR #857, opened 2023-01-09
[MLOps 1.5] Beautify the prints of the loggers

dbogunowicz:
## Feature preview

```python
from deepsparse import PythonLogger, MultiLogger, PrometheusLogger, FunctionLogger, AsyncLogger
import numpy as np
logger = AsyncLogger(MultiLogger([FunctionLogger(logger=MultiLogger([PythonLogger()] * 10),
                                                target_identifier="some_identifier",
                                                function=np.mean),
                                 FunctionLogger(logger=MultiLogger([PythonLogger(), PrometheusLogger(port=8000)]),
                                                target_identifier="some_identifier_2",
                                                function=np.mean),
                                 FunctionLogger(logger=PythonLogger(), target_identifier="some_identifier_3",
                                                function=np.mean, frequency=2)]))
print(logger)
```

```bash
AsyncLogger:
  MultiLogger:
    FunctionLogger:
      target identifier: some_identifier
      function name: mean
      frequency: 1
      target_logger:
        MultiLogger:
          PythonLogger
          PythonLogger
          PythonLogger
          PythonLogger
          PythonLogger
      [...]
    FunctionLogger:
      target identifier: some_identifier_2
      function name: mean
      frequency: 1
      target_logger:
        MultiLogger:
          PythonLogger
          PrometheusLogger:
            port: 8000
    FunctionLogger:
      target identifier: some_identifier_3
      function name: mean
      frequency: 2
      target_logger:
        PythonLogger
```
## Testing 
Added a unit test, that additionally visualizes the structure of the "`__str__` tree".


bfineran (changes requested):
Looks great @dbogunowicz strongly suggest using textwrap.indent to wrap the child loggers rather than propagating level and manually indenting.

KSGulin (dismissed):
LGTM pending @bfineran's comment. This is shaping up to be a truly elegant system

src/deepsparse/loggers/async_logger.py
```diff
@@ -59,6 +59,12 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
         )
         job_future.add_done_callback(_log_async_job_exception)
 
+    def __str__(self, level=0):
+        whitespace = "\t" * level
+        text = "\n" + whitespace + f"{self.__class__.__name__}:"
```
bfineran, line 64: quick nit as well, this whole line can be an f-string to avoid multiple concatenations (same to other instances of this)

bfineran:
also since we do nest quite a bit, to avoid running out of space on say smaller terminal widths let's do two spaces for indentation length

src/deepsparse/loggers/async_logger.py
```diff
@@ -59,6 +60,12 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
         )
         job_future.add_done_callback(_log_async_job_exception)
 
+    def __str__(self, level=0):
+        wrapper = textwrap.TextWrapper(initial_indent="\n" + "  " * level)
+        text = wrapper.fill(f"{self.__class__.__name__}:")
```
rahul-tuli, line 65: Could we use `__qualname__` instead?
More info here: https://www.codingem.com/qualname-in-python/

src/deepsparse/loggers/async_logger.py
```diff
@@ -59,6 +60,12 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
...
+        text += self.logger.__str__(level=level + 1)
+        return text
```
rahul-tuli, line 67: Could we ever go down a MaxRecursionDepth with this code?

bfineran, line 67: thoughts on this instead?
```suggestion
    def __str__(self):
        child_str = textwrap.indent(str(self.logger), prefix="  ")
        return f"{self.__class__.__name__}:\n{child_str}"
```

bfineran: since we call `indent` over the `str` of the entire child, level propagation happens automatically, as if the current "level" is > 1, then this entire constructed string will be indented by its parent(s)

bfineran (changes requested):
let's use `textwrap.indent` instead. point of previous review was to remove the need for propagating level, which requires shared logic between the classes
see: https://docs.python.org/3/library/textwrap.html#textwrap.indent

src/deepsparse/loggers/multi_logger.py
```diff
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 """
-Implementation of the Multi Logger that serves as a
+Implementation of the MultiLogger that serves as a
```
dbogunowicz, line 16:
```suggestion
Implementation of the Multi Logger that serves as a
```

src/deepsparse/loggers/function_logger.py
```diff
@@ -89,3 +90,18 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
...
+        )
+        text += "\n" + textwrap.indent(
+            f"function name: {self.function.__name__}", prefix=prefix
+        )
+        text += "\n" + textwrap.indent(
+            f"frequency: {str(self.frequency)}", prefix=prefix
+        )
+        text += "\n" + textwrap.indent("target_logger: ", prefix=prefix)
+        text += "\n" + textwrap.indent(str(self.logger), prefix=2 * prefix)
+        return f"{self.__class__.__name__}:{text}"
```
bfineran, line 107: can be cleaned up to avoid repeated `indent` calls as well as string concatenation:
```suggestion
    def _indent(value):
      return textwrap.indent(str(value), prefix="  ")

    function_info = (
        f"target identifier: {self.target_identifier}\n"
        f"function name: {self.function.__name__}\n"
        f"frequency: {self.frequency}\n"
        f"target_logger:\n{_indent(self.logger)}"
    )
    return f"{self.__class__.__name__}:\n{_indent(function_info)}"
```

src/deepsparse/loggers/function_logger.py
```diff
@@ -89,3 +90,18 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
...
+            f"target identifier: {self.target_identifier}", prefix=prefix
+        )
+        text += "\n" + textwrap.indent(
+            f"function name: {self.function.__name__}", prefix=prefix
+        )
+        text += "\n" + textwrap.indent(
+            f"frequency: {str(self.frequency)}", prefix=prefix
+        )
+        text += "\n" + textwrap.indent("target_logger: ", prefix=prefix)
+        text += "\n" + textwrap.indent(str(self.logger), prefix=2 * prefix)
```
bfineran, line 106: also wonder if there's a way we can avoid repeatedly printing the leaf loggers as this could get pretty long

src/deepsparse/loggers/multi_logger.py
```diff
@@ -41,3 +42,9 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
         """
         for logger in self.loggers:
             logger.log(identifier, value, category)
+
+    def __str__(self):
+        text = ""
+        for logger in self.loggers:
+            text += "\n" + textwrap.indent(str(logger), prefix="  ")
```
bfineran, line 49: note - can do a single indent after the entire text is constructed

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -103,6 +104,10 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
         prometheus_metric.observe(self._validate(value))
         self._export_metrics_to_textfile()
 
+    def __str__(self):
+        text = textwrap.indent(f"port: {self.port}", prefix="  ")
```
bfineran, line 108: `indent` isn't really needed here since we're guaranteed to only have 1 line

bfineran (dismissed):
LGTM, remaining comments are on code, please read, not mandatory

src/deepsparse/loggers/async_logger.py
bfineran: only class name is relevant to what we're looking for for now, users can dig in for more if needed

src/deepsparse/loggers/function_logger.py
```diff
@@ -89,3 +90,23 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
                 )
                 self._function_call_counter = 0
             self._function_call_counter += 1
+
+    def __str__(self):
+        def _indent(value):
+            return textwrap.indent(str(value), prefix="  ")
+
+        def _shorten(text, max_num_lines=10, placeholder="[...]"):
```
dbogunowicz, line 98: All the proposed solutions involving text wrap (https://stackoverflow.com/questions/1166317/python-textwrap-library-how-to-preserve-line-breaks) sounded more complicated and less readable then this small function.

corey-nm (approved):
nice tests!

bfineran (approved):

PR #858, opened 2023-01-09
Using system_logging_config directly instead of using Depends

corey-nm:
To resolve issue of
```python
  File "/home/ubuntu/workspace/Github/TestDeepsparse/runtests/lib/python3.8/site-packages/deepsparse/server/server.py", line 266, in _predict
    system_logging_config=params.get("system_logging_config"),
AttributeError: 'Depends' object has no attribute 'get'
```

# Testing plan

Ran commands in yolo README.md

### DIFF
--- a/src/deepsparse/server/server.py
+++ b/src/deepsparse/server/server.py
@@ -39,7 +39,7 @@ from deepsparse.server.system_logging import (
     SystemLoggingMiddleware,
     log_system_information,
 )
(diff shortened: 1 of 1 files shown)


bfineran (approved):
tentatively approving. @corey-nm do you know if there was a specific reason why Depends was needed? Otherwise would be good to check in with @dbogunowicz

corey-nm:
Not sure, will wait on info from @dbogunowicz

PR #863, opened 2023-01-11
[MLOps 1.5] Template for loading predefined data logging configs

dbogunowicz:
# Loading Predefined Data Logging Configs
Enables the user to specify a list of `MetricFunctionConfig` in the pipeline/server config, that will load the predefined configuration of identifiers and built-in functions.
## Feature Preview
config:
```yaml
loggers:
    python:
add_predefined:
    - func: image_classification
      frequency: 3
      target_loggers: ...
data_logging:
    pipeline_inputs.images:
    - func: ...
      frequency: 2"""
```
This will map the "function group" name `image_classification` (those names are arbitrary and are being set in the newly implemented registry of the built-in functions) to an appropriate set of identifiers and built-in functions, so that:
```python
pipeline = Pipeline.create("image_classification", logger = config)
print(pipeline.logger)
```
```bash
AsyncLogger:
  MultiLogger:
    FunctionLogger:
      target identifier: image_classification/pipeline_inputs.images
      function name: image_shape
      frequency: 2
      target_logger:
        MultiLogger:
          PythonLogger
    FunctionLogger:
      target identifier: image_classification/pipeline_inputs.images
      function name: mean_pixels_per_channel
      frequency: 2
      target_logger:
        MultiLogger:
          PythonLogger
    FunctionLogger:
      target identifier: image_classification/pipeline_inputs.images
      function name: fraction_zeros
      frequency: 2
      target_logger:
        MultiLogger:
          PythonLogger
    FunctionLogger:
      target identifier: prediction_latency
      function name: identity
      frequency: 1
      target_logger:
        MultiLogger:
          PythonLogger
```
## What's new:
Also added the registry to hold information about the target identifiers and built-in functions per task

### Testing 
Added unit tests and a sample end-to-end test.


src/deepsparse/loggers/build_logger.py
```diff
@@ -316,13 +339,44 @@ def possibly_modify_target_identifiers(
             # if the target identifier does not already start
             # with the pipeline identifier, call get_target_identifier
             # to prepend it
-            target_identifier = get_target_identifier(
+            new_target_identifier = get_target_identifier(
```
dbogunowicz, line 342: I am fixing a bug here, not relevant for the core of the PR

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +332,56 @@ def possibly_modify_target_identifiers(
     return data_logging_config
 
 
+def data_logging_config_from_predefined(
+    data_logging_config: List[MetricFunctionConfig],
+) -> Dict[str, List[MetricFunctionConfig]]:
+    """
+    Given a single MetricFunctionConfig object, parse out
+    the information about the pre-defined data_logging configuration.
```
dbogunowicz, line 340:
```suggestion
    the information about the pre-defined data logging configuration.
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +332,56 @@ def possibly_modify_target_identifiers(
...
+
+    The MetricFunctionConfig maps to a set of built-in functions
```
dbogunowicz, line 342:
```suggestion
    The MetricFunctionConfig.func maps to a set of built-in functions
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +332,56 @@ def possibly_modify_target_identifiers(
...
+    and identifiers that will be retrieved from the registry to
+    build a new data_logging_config.
```
dbogunowicz, line 344:
```suggestion
    build a new data logging config.
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +332,56 @@ def possibly_modify_target_identifiers(
...
+
+    :param data_logging_config: A list containing a single
+        MetricFunctionConfig object that specifies the
+        data logging configuration.
```
dbogunowicz, line 348:
```suggestion
        predefined data logging configuration.
```

src/deepsparse/loggers/metric_functions/registry.py
```diff
@@ -0,0 +1,59 @@
...
+from collections import defaultdict
+
+
+__all__ = ["DATA_LOGGING_REGISTRY", "register"]
+
+
+DATA_LOGGING_REGISTRY = defaultdict(lambda: defaultdict(str))
+
+
+def register(task: str, identifier: str):
```
bfineran, line 27: might want to rename task to "group" so we can eventually register multiple "groups" for a task

src/deepsparse/loggers/metric_functions/computer_vision/built_ins.py
```diff
@@ -19,6 +19,8 @@
 
 import numpy
 
+from deepsparse.loggers.metric_functions import register
```
bfineran, line 22: nice - like how easy it is to construct a set of functions for a task. need to now think about how to manage when registry happens - especially due to dependencies (ie will this file always be imported with the `loggers` submodule so registration fires?

src/deepsparse/loggers/metric_functions/registry.py
```diff
@@ -0,0 +1,59 @@
...
+    task `image_classification` and the data_logging_identifier
+    `pipeline_inputs.image`
+
+    :param task: The name of the task, that uses the function in question.
+        Same functions may be registered under multiple tasks.
+    :param identifier: The name of the identifier, that the function in question
+        acts on. A single identifier may be related to multiple functions.
+    """
+
+    def decorator(f):
```
bfineran, line 47: nit: single char name - rename to `function`

bfineran (changes requested):

src/deepsparse/loggers/metric_functions/computer_vision/built_ins.py
dbogunowicz: Good point. In the incoming commit, I will be probably postponing the actual registration until the logic of logger building encounters a data logging config target that will be associated with loading the predefined configuration. Stay tuned.

src/deepsparse/loggers/build_logger.py
```diff
@@ -245,7 +246,19 @@ def build_data_loggers(
     data_loggers = []
     if not data_logging_config:
         return data_loggers
+
     for target_identifier, metric_functions in data_logging_config.items():
+        if target_identifier.endswith(PREDEFINED_IDENTIFIER_ALIAS):
+            """
+            If the target identifier ends with the predefined alias,
+            overwrite the target identifier and metric functions with the one
```
dbogunowicz, line 254:
```suggestion
            overwrite the current target identifier and metric functions with the one
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +338,52 @@ def possibly_modify_target_identifiers(
...
+        if not registered_function_group:
+            raise ValueError(
+                f"Unknown function group name: {function_group_name}. "
+                f"Supported function group names: {list(DATA_LOGGING_REGISTRY.keys())}"
+            )
+        for (
+            registered_identifier,
+            registered_functions,
+        ) in registered_function_group.items():
+            target_identifier = identifier_prefix + registered_identifier
```
dbogunowicz, line 375:
```suggestion
            target_identifier = "f{identifier_prefix}{registered_identifier}"
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -245,7 +246,19 @@ def build_data_loggers(
...
+        if target_identifier.endswith(PREDEFINED_IDENTIFIER_ALIAS):
+            """
+            If the target identifier ends with the predefined alias,
+            overwrite the current target identifier and metric functions with the one
+            retrieved from the predefined data logging confuguration
+            """
+            for target_identifier, metric_functions in metric_functions_from_predefined(
+                metric_functions, target_identifier[: -len(PREDEFINED_IDENTIFIER_ALIAS)]
+            ):
+                pass
```
dbogunowicz, line 260: This may be a bit ugly but does the job.

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +338,52 @@ def possibly_modify_target_identifiers(
...
+    pairs of (target_identifier, List[MetricFunctionConfig]), that
+    are then used to build the data loggers.
+
+    :param metric_functions: A list containing MetricFunctionConfig
+        objects that specify the predefined data logging configuration.
+    :return: A generator object that generates a sequence of
+        (target_identifier, List[MetricFunctionConfig]) tuples.
+    """
+    # get the registry with the pre-defined data logging configurations
+    from deepsparse.loggers.metric_functions.registry import DATA_LOGGING_REGISTRY
```
dbogunowicz, line 360: Postponing the process of registration of all the built-ins until this function is called.

src/deepsparse/loggers/helpers.py
```diff
@@ -91,6 +139,7 @@ def get_function_and_function_name(
 
     :return: A tuple (function, function name)
     """
+    import deepsparse.loggers.metric_functions as built_ins
```
dbogunowicz, line 142: Alternatively, postponing the process of registration of the built-ins until this function is called (we want to fetch all the built-ins (CV, NLP, etc) here on import, so our decorators will be evoked.

src/deepsparse/loggers/metric_functions/computer_vision/built_ins.py
```diff
@@ -32,13 +34,15 @@
 ]
 
 
+@register(group="image_classification", identifier="pipeline_inputs.images")
```
dbogunowicz, line 37: Will be adding register to the rest of the functions in the next PRs

src/deepsparse/loggers/constants.py
```diff
@@ -23,11 +23,13 @@
     "validate_identifier",
     "REQUEST_DETAILS_IDENTIFIER_PREFIX",
     "RESOURCE_UTILIZATION_IDENTIFIER_PREFIX",
+    "PREDEFINED_IDENTIFIER_ALIAS",
 ]
 
 UNSUPPORTED_IDENTIFIER_CHARS = {".", "[", "]"}
 REQUEST_DETAILS_IDENTIFIER_PREFIX = "request_details"
 RESOURCE_UTILIZATION_IDENTIFIER_PREFIX = "resource_utilization"
+PREDEFINED_IDENTIFIER_ALIAS = "predefined"
```
bfineran, line 32: based on usage, looks like this might be cleaner unless we add more helper fns
```suggestion
PREDEFINED_IDENTIFIER_ALIAS = "_predefined"
```

src/deepsparse/loggers/build_logger.py
bfineran: why is this necessary? does mutation not properly work with the registry. if so, this is a bug, any time a new function is registered, the global registry should be up to date

src/deepsparse/loggers/build_logger.py
bfineran: why is this necessary? if we only care about the last `metric_functions`, then why require the generator at all instead of just returning what we actually want?

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +338,52 @@ def possibly_modify_target_identifiers(
...
+        if not registered_function_group:
+            raise ValueError(
+                f"Unknown function group name: {function_group_name}. "
+                f"Supported function group names: {list(DATA_LOGGING_REGISTRY.keys())}"
+            )
+        for (
+            registered_identifier,
+            registered_functions,
+        ) in registered_function_group.items():
+            target_identifier = "f{identifier_prefix}{registered_identifier}"
```
bfineran, line 375: can we go through the get identifier helper here instead?

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +338,52 @@ def possibly_modify_target_identifiers(
     return data_logging_config
 
 
+def metric_functions_from_predefined(
+    metric_functions: List[MetricFunctionConfig], identifier_prefix: str
+) -> Generator[Tuple[str, List[MetricFunctionConfig]], None, None]:
```
bfineran, line 343: confused why we would want a generator here, especially since we skip over the first n-1 results. why not just return an overall dict of identifier name mapped to all of the metric functions for it across the built ins specified

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +338,52 @@ def possibly_modify_target_identifiers(
     return data_logging_config
 
 
+def metric_functions_from_predefined(
+    metric_functions: List[MetricFunctionConfig], identifier_prefix: str
```
bfineran, line 342: let's rename metric functions to `predefined_metric_functions` or something to specify that this function on ly works for those

bfineran (changes requested):

src/deepsparse/loggers/constants.py
dbogunowicz: Are you sure?
`"image_classification_pipeline/predefined".endswith("predefined") == True`
`"image_classification_pipeline/predefined".endswith("predefined") == False`

dbogunowicz: Given that we keep the convention of the config file:
```yaml
loggers:
    list_logger:
        path: tests/deepsparse/loggers/helpers.py:ListLogger
data_logging:
    predefined: # `_predefined` is ugly imo
    - func: image_classification
      frequency: 1"""
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -39,6 +40,7 @@
     SystemLoggingGroup,
 )
 from deepsparse.loggers.helpers import get_function_and_function_name
+from deepsparse.loggers.metric_functions.registry import DATA_LOGGING_REGISTRY
```
bfineran, line 43: will the registry be populated here if importing directly from ` deepsparse.loggers.metric_functions.registry` want to make sure it is by the time it is called (rather than depending on going through `get_function_and_function_name`

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +332,83 @@ def possibly_modify_target_identifiers(
...
+) -> Dict[str, List[MetricFunctionConfig]]:
+    """
+    Parse out the predefined metric functions from the
+    data logging config and update the data_logging_config
+    accordingly
+
+    :param data_logging_config: The configuration of the data loggers
+    :return: The updated configuration of the data loggers
+    """
+    for target_identifier, metric_functions in data_logging_config.copy().items():
```
bfineran, line 346: hm, not sure about this copy-them-modify flow in terms of maintainability, especially since we are modifying a user input in-place. was this added in to address a dict mutation during iteration issue?
i'd consider either first parsing out any target identifiers for predefined, then iterating through those:
```python
predefined_identifiers = [identifier for identifier in data_logging_config if idenfier.startswith(...)]
for predefined_identifier in predefined_identifiers:
   # add and delete
```

or build a new data_logging_config and update/build on every loop iteration
```python
for target_identifier, metric_functions in data_logging_config.items():
    if target_identifier.endswith(PREDEFINED_IDENTIFIER_ALIAS):
        ...
    else:
       updated_data_logging_config[target_identifier] = metric_functions
```

src/deepsparse/loggers/build_logger.py
```diff
@@ -325,6 +332,83 @@ def possibly_modify_target_identifiers(
...
+                target_name=registered_identifier, pipeline_identifier=identifier_prefix
+            )
+            new_metric_functions = [
+                MetricFunctionConfig(
+                    func=func,
+                    frequency=metric_function.frequency,
+                    target_loggers=metric_function.target_loggers,
+                )
+                for func in registered_functions
+            ]
```
bfineran, line 404: why allocate a new list instead of append to the config? (could use defaultdict)

src/deepsparse/loggers/helpers.py
bfineran: why do we delay this?

bfineran (dismissed):
great refactor @dbogunowicz! comments are just food for thought

src/deepsparse/loggers/build_logger.py
dbogunowicz: yes, it will. on the import, the `DATA_LOGGING_REGISTRY` "goes through" all its decorators and registers all the appropriate functions

tests/deepsparse/loggers/metric_functions/template_functions/test_template_functions.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+yaml_config = """
+loggers:
+    list_logger:
+        path: tests/deepsparse/loggers/helpers.py:ListLogger
+data_logging:
+    predefined:
+    - func: image_classification
+      frequency: 1
```
corey-nm, line 27: It's not clear from the current predefined block what is being passed into the functions (is it pipeline inputs, outputs, everything?), and it's also a different paradigm than we are using for other functions. This means when documenting this feature we have to have a special section for specifying predefined things that work differently from the rest.

Additionally, since this is an inconsistency with the rest of the UX, as a user I might wonder if func/frequency mean the same thing here as they do for other functions? This will increase mental burden when writing yaml configs, as its adding a second set of rules a user has to know.

That being said I suggest the following UX:
```suggestion
    pipeline_inputs:
    - func: builtins.image_classification
      frequency: 1
```
(alternatively can name it `predefined.image_classification`)

I think this formulation would be more clear to user, and also more consistent with our existing setup. Further it would allow more fine grained control on what exactly is being logged in a way that is clear just from the yaml file.

dbogunowicz: Thanks for that @corey-nm .I agree that your proposal adds a mental burden for users, but I think it's manageable with proper documentation. The alternative "convention" gives a different meaning to a template identifier, which can be confusing but eventually may be understandable. However, what I don't like about the proposal is that users still need to think about different target identifiers, such as pipeline_inputs, pipeline_outputs, etc. My main goal was to have one pathway that loads all functions for all identifiers at once. To summarize, your proposal reduces some of the mental burden for users, but it also requires more explicit configs. I see there is a balance to strike here, I would like to hear Ben's opinion on this as well.

bfineran: Predefined was selected to as a catch all since the built ins cover multiple pipeline values. Could potentially choose a different name than predefined, but will need some catch all value as @dbogunowicz mentioned

corey-nm: @dbogunowicz @bfineran Can this be handled in a block separate from data_logging then? For example:

```yaml
inherits_from:
  - builtins.image_classification
data_logging:
  whatever: {...}
  can_be_anything: {...}
```

The nice thing about this is it is explicitly different from the rest of data_logging, and users would know it has different rules. It becomes very easy to document this additional block when writing the builtins, and since its just a list of strings, its easy to use.

The other thing is this allows us to inherit more things than just logging in the future if we wanted to (e.g. hyperparameters).

dbogunowicz: In that case, we need to maintain and process additional "config", I remain sceptical and not convinced :)

corey-nm: Ask chatGPT to explain the existing config and the one I posted and see which one it explains better. If it explains both equally well then let's go with the existing implementation, but I'm guessing it won't know what it does 😀

corey-nm: And also, I'd be willing to pay some slight configuration parsing complexity for better UX 🤷

dbogunowicz: image
I think you are collaborating

corey-nm: 😉 i knew i liked chatGPT

dbogunowicz: the roman orators call it argument ad robotum

corey-nm: 🤣

bfineran: this LGTM, want to be cautious of adding too many args in the future, but prefer this over the more "hidden" functionality we currently have.

"inherits_from" sounds a bit vague for the actual purpose. thoughts on `add_buildins` or `function_groups` as a key instead?

corey-nm (approved):
lgtm! 🚀

bfineran (approved):

PR #865, opened 2023-01-16
[MLOps 1.5] Expand the built-ins: NLP

dbogunowicz:
## Feature description

- Added built-in functions:  'mean_score, percent_zero_labels (token classification) answer_length, answer_score, answer_found (question answering), sequence_length (general nlp function)`.
- Added exhaustive unit tests.
- Cleaned up the importing structure to avoid circular dependencies

## Coverage:

As specified in the PRD:
```yaml
nlp_sequence_length -->  covered by "sequence_length "
nlp_percent_unk_tokens --> not covered; see comments below
nlp_answer_found_qa - 1 if yes, 0 if no --> covered by "answer_found"
nlp_answer_length_qa - integer length --> covered by "answer_length"
nlp_score_qa - 0-1 float representing confidence in answer --> covered by "answer_score"
nlp_percent_label_0_token_classification - 0-1 float --> covered by percent_zero_labels
nlp_mean_score_token_classification - 0-1 float --> covered by  "mean score"
```

## Few Comments:
1. We still need to think about a good naming framework, to avoid ambiguity and possible dupes of the functions naming across integrations. Let's discuss it.
2. For computing the percentage of the unknown tokens, I am afraid that currently we are heavily constrained. This information is present during the tokenization process but is not available inside the `engine_inputs`. Hence, we are currently not able to quickly compute this value in a simple, monadic function.


src/deepsparse/loggers/metric_functions/natural_language_processing/built_ins.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+Set of functions for logging metrics from the natural language processing pipelines
+"""
+from typing import Dict, List, Union
+
+
+__all__ = ["sequence_length", "percent_unknown_tokens"]
+
+
+def sequence_length(sequence: Union[List[str], str]) -> Union[Dict[str, int], int]:
```
bfineran, line 23: `sequence_length` in NLP has the specific connotation of the number of tokens in a sequence rather than the string length. let's rename this to `string_length`

src/deepsparse/loggers/metric_functions/natural_language_processing/question_answering/built_ins.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def answer_length(qa_output: "QuestionAnsweringOutput") -> int:  # noqa: F821
+    """
+    Returns the length of the answer given the QuestionAnsweringOutput
+
+    :param qa_output: The output schema of the question answering pipeline
+    :return: The length of the answer
+    """
+    return sequence_length(qa_output.answer)
```
bfineran, line 42: shouldn't we account for `"empty"` here?

src/deepsparse/loggers/metric_functions/natural_language_processing/token_classification/built_ins.py
```diff
@@ -0,0 +1,75 @@
...
+) -> Dict[str, float]:
+    """
+    Returns the percentage of zero labels in the token classification output
+
+    :param token_classification_output: the TokenClassificationOutput object
+    :return: A dictionary where the key is the token sequence index and the
+        value is the percentage of zero labels in the sequence of tokens
+    """
+    result = {}
+    for prediction_idx, prediction in enumerate(
```
bfineran, line 36: how would the enumerated values be logged currently? they would look like `percent_zero_labels__{idx}` right? thinking we'd prefer for each item in the batch to be logged separately to the same identifier, no? (this goes for all the fns we handle this way)

bfineran (changes requested):

src/deepsparse/loggers/metric_functions/natural_language_processing/token_classification/built_ins.py
dbogunowicz: Good point...

The user probably doesn't care about the specific ordering of their predictions in the dataset, so this information does seem to be adding clutter, not value. 
Ideally, in this case, we would like to have a dict like `{"": 10, "":11}` (so we get `percent_zero_labels_0.6` etc.) which obviously is not feasible.

Let's agree to land this PR for now, I will create a ticket for myself and handle the logging of items in a batch in a different PR ok?

bfineran: we need to think more here on how to handle batchwise VS item-wise logs. we might even want a wrapper object such so we can do something like:

```python
# wrapper class
class BatchLogResult:
    results: List[Any]

# example function
def item_wise_function(batch) -> BatchLogResult:
    return BatchLogResult([my_function(item) for item in batch])

# example leaf logger handling
class SomeLeafLogger:
    def log(item):
        if isinstance(item, BatchLogResult):
            # recurse for batch item in BatchLogResult.results
        else:
            self.logger.log(item)
```

bfineran (approved):
LGTM for now, but absolutely need to get something in place to handle item-wise functions

src/deepsparse/loggers/metric_functions/natural_language_processing/token_classification/built_ins.py
dbogunowicz: Agree

tests/deepsparse/loggers/metric_functions/natural_language_processing/built_ins.py
```diff
@@ -0,0 +1,29 @@
...
+import pytest
+from deepsparse.loggers.metric_functions.natural_language_processing import (
+    string_length,
+)
+
+
+@pytest.mark.parametrize(
+    "string, expected_len",
+    [
+        ("His palms are sweaty", 20),
```
KSGulin, line 24: This is a high quality test

KSGulin (approved):
LGTM!

PR #866, opened 2023-01-18
[MLOps 1.5] Enable `PrometheusLogger` to unwrap the logged value if its a dict.

dbogunowicz:
# Loading Predefined Data Logging Configs
Since Prometheus is only able to log primitive data types such as floats or ints, we would like to break down any complicated logged values (i.e. the ones represented as dicts) as a sequence of flattened dict representations

## Feature Preview

```python
value = {"foo": {"alice": 1, "bob": 2}, "bar": 5}
identifier = "identifier"
prometheus_logger.log(identifier, value, category = ....)
```
will split the act of logging `identifier` &`value` pair into three separate logging activities:
identifier = `identifier__foo__alice` value = 1
identifier = `identifier__foo__bob` value = 2
identifier = `identifier__bar` value = 5

### Testing 
Added unit-tests


src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -90,18 +91,19 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
...
             )
-        prometheus_metric.observe(self._validate(value))
-        self._export_metrics_to_textfile()
+            prometheus_metric = self._prometheus_metrics.get(formatted_identifier)
+            if prometheus_metric is None:
+                prometheus_metric = self._add_metric_to_registry(
+                    formatted_identifier, category
+                )
+            prometheus_metric.observe(self._validate(value))
+            self._export_metrics_to_textfile()
```
bfineran, line 106: might want to break this section out into multiple helper functions, especially as we are getting into using more prometheus functionality

src/deepsparse/loggers/helpers.py
```diff
@@ -34,11 +34,60 @@
...
     "access_nested_value",
     "finalize_identifier",
+    "unwrap_logs_dictionary",
 ]
 
 NO_MATCH = "NO_MATCH"
 
 
+def unwrap_logs_dictionary(
+    value: Any, parent_identifier="", seperator="__"
```
bfineran, line 44: let's default separator to `/` since this is what we use in actual pipeline logging. the `__` is prometheus specific and will be injected later, correct?

bfineran, line 44: nit: missing typing for identifier and separator

src/deepsparse/loggers/helpers.py
```diff
@@ -34,11 +34,60 @@
...
+            yields a sequence of tuples (identifier, value) where:
+                identifier is composed by connecting the keys over the multiple levels
+                    of nesting with the seperator
+                value is extracted from the nested dictionary (corresponding to the
+                    appropriate composed identifier)
+    """
+    if not isinstance(value, dict):
+        yield parent_identifier, value
+    else:
+        for child_identifier, child_value in value.items():
```
bfineran, line 77: good job with the variable names, very easy to track

bfineran (changes requested):
super clean 👌 , few comments to resolve then good to go

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: I am doing it atm while working on the POC for matching the logging targets to appropriate prometheus metrics. Let's turn the blind eye in the context of this PR.

src/deepsparse/loggers/helpers.py
dbogunowicz: Not really.
To reiterate:

Identifiers in our logging framework have three kinds of separators: `/`, `.`, and `__`.
The convention is:
`whatever/we/want/to/have/here.remainder__metric_function_name__subname__subsubname`
e.g.
`image_classification/pipeline_inputs.images__mean_pixels_per_channel__0`

So this identifier is not particular to Prometheus formatting. It continues to append nested subnames to the output of the metric function.

src/deepsparse/loggers/helpers.py
dbogunowicz: Thanks!

bfineran (approved):

src/deepsparse/loggers/prometheus_logger.py
corey-nm, line 106: Should this line be de-indented?

corey-nm (commented):
lgtm, left one question then will approve

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: Really depends on how we understand the frequency of exporting to textfile.
If we de-ident, we increment the counter every time the log method is called.
Otherwise, we increment the counter every time we actually send a valid value to PrometheusLogger.
I think that the current implementation is "more correct"

corey-nm (approved):
🚀

PR #867, opened 2023-01-19
[MLOps 1.5] Autogenerate data logging configs from the registry

dbogunowicz:
## Feature Preview:
```python
string_result = data_logging_config_from_predefined(
        group_names="image_classification",
        frequency=2
        loggers={"some_logger": {"arg_1": 10}},
    )
```
outputs:
```bash
loggers:
  python: null

data_logging:
  pipeline_outputs.labels:
  - func: predicted_classes
    frequency: 1
  - func: predicted_top_score
    frequency: 1
  pipeline_inputs.images:
  - func: image_shape
    frequency: 1
  - func: mean_pixels_per_channel
    frequency: 1
  - func: std_pixels_per_channel
    frequency: 1
  - func: fraction_zeros
    frequency: 1
```

## Description
Introduces a new helper function, `data_logging_config_from_predefined`, that generates data logging configurations using metric group names and optional arguments. The function enables developers to quickly generate examples of data logging configurations, as demonstrated in the examples/data-logging-configs directory.

This PR includes comprehensive testing and features a test that, when run with the appropriate environment variable, can regenerate a fresh set of examples. This functionality is useful for maintaining up-to-date configurations as the log structure evolves over time. :
```bash
GENERATE_CONFIGS=1 pytest tests/deepsparse/loggers/metric_functions/helpers/test_data_config_from_predefined.py 
```

PR #868, opened 2023-01-19
[MLOps 1.5] Autogenerate data logging configs from the registry (helper functions)

dbogunowicz:
A collection of all the necessary (in my opinion) helper functions from which we can quickly build the actual `data_logging_config_from_predefined` function.

This PR is very much TDD-driven, so the unit-tests serve in my opinion as a solid documentation on each of the helper functions.

### DIFF
--- a/src/deepsparse/loggers/metric_functions/helpers/config_generation.py
+++ b/src/deepsparse/loggers/metric_functions/helpers/config_generation.py
@@ -28,8 +28,8 @@ _WHITESPACE = "  "
 
 def data_logging_config_from_predefined(
     group_names: Union[str, List[str]],
-    loggers: Dict[str, Optional[Dict[str, Any]]] = {"python": {}},
     frequency: int = 1,
+    loggers: Optional[Dict[str, Optional[Dict[str, Any]]]] = None,
     save_dir: Optional[str] = None,
     save_name: str = "data_logging_config.yaml",
 ) -> str:
@@ -58,27 +58,34 @@ def data_logging_config_from_predefined(
 def _loggers_to_config_string(
     loggers: Dict[str, Optional[Union[str, List[str]]]]
 ) -> str:
-    return ("\n").join(_nested_dict_to_lines(loggers, yaml_str_lines=[]))
+    return ("\n").join(_nested_dict_to_lines(loggers))
 
 
 def _nested_dict_to_lines(
-    value: Any, key=None, yaml_str_lines=[], level=0
+    value: Any,
+    key: Optional[str] = None,
+    yaml_str_lines: Optional[List[str]] = None,
+    _level: int = 0,
 ) -> List[str]:
     # converts a nested dictionary to a list of yaml string lines
+    if yaml_str_lines is None:
+        yaml_str_lines = []
+
+    identation = _WHITESPACE * _level
+
     for new_key, new_value in value.items():
         if isinstance(new_value, dict):
-            yaml_str_lines.append(f"{_WHITESPACE*level}{new_key}:")
+            yaml_str_lines.append(f"{identation}{new_key}:")
             yaml_str_lines = _nested_dict_to_lines(
-                new_value, new_key, yaml_str_lines, level + 1
+                new_value, new_key, yaml_str_lines, _level + 1
+            )
+        elif isinstance(new_value, list):
+            list_as_str = _metric_functions_configs_to_string(new_value)
+            yaml_str_lines.append(
+                f"{new_key}:\n{textwrap.indent(list_as_str, prefix=_WHITESPACE)}"
             )
         else:
-            if isinstance(new_value, list):
-                # we can safely return here, since we are not expecting nested lists
-                yaml_str_lines.append(
-                    f"{new_key}:\n{textwrap.indent(_metric_functions_configs_to_string(new_value), prefix=_WHITESPACE)}"  # noqa E501
-                )
-            else:
-                yaml_str_lines.append(f"{_WHITESPACE*level}{new_key}: {new_value}")
+            yaml_str_lines.append(f"{identation}{new_key}: {new_value}")
 
     return yaml_str_lines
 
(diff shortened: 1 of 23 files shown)


src/deepsparse/loggers/metric_functions/helpers/config_generation.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _loggers_to_config_string(
+    loggers: Dict[str, Optional[Union[str, List[str]]]]
+) -> str:
+    return ("\n").join(_nested_dict_to_lines(loggers, yaml_str_lines=[]))
+
+
+def _nested_dict_to_lines(
+    value: Any, key=None, yaml_str_lines=[], level=0
```
dbogunowicz, line 65: I know @bfineran , using levels once again :)
Was easier for me to lay out this helper function this way. I will refactor this function to remove this arg before landing ok?

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
```diff
@@ -0,0 +1,123 @@
...
+
+from deepsparse.loggers.config import MetricFunctionConfig
+
+
+_WHITESPACE = "  "
+
+
+def data_logging_config_from_predefined(
+    group_names: Union[str, List[str]],
+    loggers: Dict[str, Optional[Dict[str, Any]]] = {"python": {}},
```
bfineran, line 31: been leaving this comment a lot - we cannot use mutable types as function defaults in python as they can have unintended side effects. read more here: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
bfineran, line 65: same here on mutable default

bfineran: level here is fine - the difference is the scope is limited to a single function with self contained logic in its recursion, can be more specific if you want and rename to `_level` to indicate top level users shouldn't specify

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
```diff
@@ -0,0 +1,123 @@
...
+    raise NotImplementedError()
+
+
+def _loggers_to_config_string(
+    loggers: Dict[str, Optional[Union[str, List[str]]]]
+) -> str:
+    return ("\n").join(_nested_dict_to_lines(loggers, yaml_str_lines=[]))
+
+
+def _nested_dict_to_lines(
```
bfineran, line 64: `_nested_dict_to_yaml_str`

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
```diff
@@ -0,0 +1,123 @@
...
+                new_value, new_key, yaml_str_lines, level + 1
+            )
+        else:
+            if isinstance(new_value, list):
+                # we can safely return here, since we are not expecting nested lists
+                yaml_str_lines.append(
+                    f"{new_key}:\n{textwrap.indent(_metric_functions_configs_to_string(new_value), prefix=_WHITESPACE)}"  # noqa E501
+                )
+            else:
+                yaml_str_lines.append(f"{_WHITESPACE*level}{new_key}: {new_value}")
```
bfineran, line 81: should use elif to save a level of indentation

```python
elif isinstance(...):
    ...
else:
    yaml_str_lines...

File: src/deepsparse/loggers/metric_functions/helpers/config_generation.py [hunk 5]
```diff
@@ -0,0 +1,123 @@
...
+) -> List[str]:
+    # converts a nested dictionary to a list of yaml string lines
+    for new_key, new_value in value.items():
+        if isinstance(new_value, dict):
+            yaml_str_lines.append(f"{_WHITESPACE*level}{new_key}:")
+            yaml_str_lines = _nested_dict_to_lines(
+                new_value, new_key, yaml_str_lines, level + 1
+            )
+        else:
+            if isinstance(new_value, list):
```
bfineran, line 75: why not use `str_list_to_yaml`?

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
bfineran: (moving discussion to #867 since this will go away after rebase)

bfineran (changes requested):
LGTM overall, let's move the base branch from `main` to `[feature/damian/config_generation](https://github.com/neuralmagic/deepsparse/tree/feature/damian/config_generation)`

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
dbogunowicz: Interesting!

dbogunowicz: I actually now see that this does mess up my code.

dbogunowicz: I actually do experience it right now but with a different function (inside the same module):
```
def _nested_dict_to_lines(
    value: Any, key=None, yaml_str_lines=[], _level=0
) -> List[str]:```

such a counterintuitive thing....

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
dbogunowicz: This is a bit misleading, this function outputs a list of string lines.

src/deepsparse/loggers/metric_functions/helpers/config_generation.py
dbogunowicz: Well we do. `_metric_functions_configs_to_string()` uses that function.

bfineran (approved):

PR #869, opened 2023-01-19
Hack up onnx.py to handle models with external data

bnellnm:
I hacked this up so deepsparse.benchmark would run on mbart.  Probably not ready for prime time.


mgoin (commented):
looks good as a first pass but should also add a wrapper for `truncate_onnx_model` as it ends with
```
    # save and check model
    onnx.save(extracted_model, output_filepath)
    onnx.checker.check_model(output_filepath)
```

src/deepsparse/utils/onnx.py
```diff
@@ -68,6 +68,48 @@
...
+    try:
+        yield shaped_model.name
+    finally:
+        os.unlink(shaped_model.name)
+        shaped_model.close()
+        if has_external_data:
+            external_data_path = os.path.join(
+                os.path.dirname(shaped_model.name), external_data
+            )
+            os.unlink(external_data_path)
```
mgoin, line 110: i'm not sure how this will interact with the `@contextlib.contextmanager` decorator in the consumer function.. maybe this function needs the decorator as well and the consumers should also `yield` the return value?

bnellnm: for the record if i add @contextlib.contextmanager to save_onnx then i get an error.  if i remove it from the functions that are now calling it then the error goes away.   so i'm guessing the best thing is to just move this annotation to save_onnx?

src/deepsparse/utils/onnx.py
```diff
@@ -204,17 +246,9 @@ def override_onnx_batch_size(onnx_filepath: str, batch_size: int) -> str:
...
-    try:
-        yield shaped_model.name
-    finally:
-        os.unlink(shaped_model.name)
-        shaped_model.close()
+    return save_onnx(model)
 
 
-@contextlib.contextmanager
 def override_onnx_input_shapes(
```
corey-nm, line 252: hmm do all usages of this function that use it as a context manager need to be changed to normal functions? (same question for override_onnx_batch_size)

src/deepsparse/utils/onnx.py
corey-nm: Also unsure, it's definitely hard to reason about lol. It might be easier to refactor all these functions to act directly on onnx models and handle the saving logic directly in the pipelines where these are called. But that's a wider change than the scope of this PR at least.

> if i remove it from the functions that are now calling it then the error goes away.

Was this with contextmanager + yielding the return value was crashing?

bnellnm: yeah, if i tagged all the functions with that, it would error out.  things work if only save_onnx is tagged or if the immediate "override..." callers are tagged.

src/deepsparse/utils/onnx.py
```diff
@@ -68,6 +68,49 @@
...
+        onnx.checker.check_model(model)
+        onnx.save(model, model_path)
+        return False
+    except ValueError:
+        # Presumably the checker has failed because of external data
+        onnx.save_model(
+            model,
+            model_path,
+            save_as_external_data=True,
+            all_tensors_to_one_file=True,
```
bfineran, line 87: any reason we prefer `all_tensors_to_one_file`?

src/deepsparse/utils/onnx.py
```diff
@@ -68,6 +68,49 @@
...
+    Save model to the given path.  If the model has external data, store the
+    external data in 'external_data_file'.
+    Returns False if the model had no external data, True otherwise.
+    """
+    try:
+        onnx.checker.check_model(model)
+        onnx.save(model, model_path)
+        return False
+    except ValueError:
+        # Presumably the checker has failed because of external data
```
bfineran, line 82: I think this is fine for now, we could also try checking the total size of the proto in memory to see if it's at or around 2G

src/deepsparse/utils/onnx.py
```diff
@@ -68,6 +68,49 @@
...
+            model_path,
+            save_as_external_data=True,
+            all_tensors_to_one_file=True,
+            location=external_data_file,
+        )
+        return True
+
+
+@contextlib.contextmanager
+def save_onnx(model: Model) -> str:
```
bfineran, line 94: would rename to specify that the functionality added here is to save to temporary files. would also rename `save_onnx_to_path` as `save_onnx` so that can be our source of truth for saving in general

src/deepsparse/utils/onnx.py
```diff
@@ -68,6 +68,49 @@
...
+    external data in 'external_data_file'.
+    Returns False if the model had no external data, True otherwise.
+    """
+    try:
+        onnx.checker.check_model(model)
+        onnx.save(model, model_path)
+        return False
+    except ValueError:
+        # Presumably the checker has failed because of external data
+        onnx.save_model(
```
bfineran, line 83: do we not also need to call `convert_model_to_external_data` 

(`from onnx.external_data_helper import convert_model_to_external_data`)

bfineran (commented):
lgtm for an initial pass pending comments

src/deepsparse/utils/onnx.py
bnellnm: if we don't do this then we need to figure out all the files that need to be deleted later.

src/deepsparse/utils/onnx.py
bnellnm: afaict that doesn't seem to be necessary.

src/deepsparse/utils/onnx.py
bnellnm: I found a check like this in the onnx source code so I've applied similar logic here.

mgoin (approved):

bfineran (approved):

PR #871, opened 2023-01-20
[MLOps 1.5] [PrometheusLogger] Support for different Prometheus metrics (depending on the identifier type)

dbogunowicz:
The following functionality is implemented:

We maintain a dictionary that stores the relationship between:

  - A `system_logging_group` and Prometheus metric type (if all targets in this group use the same Prometheus metric type)
  - A `system_metric_group/target `and Prometheus metric type (if targets within the group use different Prometheus metric types)

As a result:

1. If a log pertains to the category `DATA`, it will always use the 'Summary' metric.
2. If a log pertains to the category `SYSTEM`, it will:
   - Be matched with an appropriate metric
   - Be matched with `None` (if an identifier does not map to any metric). In this case, we will warn the user once and then add the `None` to the dictionary of metrics `self._prometheus_metrics`)
  
We will discuss this PR further online. One aspect I am not particularly fond of is maintaining the `_IDENTIFIER_TO_METRIC_TYPE` mapping, but it seems to be an inevitable task.


src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -170,6 +198,33 @@ def get_prometheus_metric(
...
+            e.g. "prediction_latency",
+            the match requires the system_group_name to be
+            contained within the identifier
+            e.g. prediction_latency/pipeline_inputs
+        """
+        if ("/" in system_group_name and identifier.endswith(system_group_name)) or (
+            system_group_name in identifier
+        ):
+            return metric_type
+    warnings.warn(
```
bfineran, line 222: can we not do any inference or default on what kind of metric to use here? or is this already handled?

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -98,18 +116,22 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
...
         prometheus_metric = self._get_prometheus_metric(identifier, category)
+        if prometheus_metric is None:
+            return
         prometheus_metric.observe(self._validate(value))
         self._export_metrics_to_textfile()
 
     def _get_prometheus_metric(
         self, identifier: str, category: MetricCategories
-    ) -> Summary:
+    ) -> Union[None, Histogram, Gauge, Summary, Counter]:
```
bfineran, line 126: if we're going to be using this a lot for our supported prometheus metrics, I'd suggest maybe adding an alias for this ie
```
PrometheusMetric = Union[Histogram, Gauge, Summary, Counter]
```
and update to `Optional[PROMETHEUS_METRIC]

bfineran (approved):
looks good - at this point we 100% need a documentation page on this to keep track of these kinds of "hidden" hardcoded logic

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: Let's sync on that offline, during the same discussion when we will be discussing adding `model_name` label.

corey-nm (approved):
lgtm, nice diff!

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -170,6 +199,33 @@ def get_prometheus_metric(
...
+            e.g. "prediction_latency",
+            the match requires the system_group_name to be
+            contained within the identifier
+            e.g. prediction_latency/pipeline_inputs
+        """
+        if ("/" in system_group_name and identifier.endswith(system_group_name)) or (
+            system_group_name in identifier
+        ):
+            return metric_type
+    warnings.warn(
```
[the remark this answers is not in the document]
dbogunowicz: I think that if we do that, the warning will "pop up" every time we are encountering a log that cannot be processed by Prometheus. I chose to warn here, so that the warning happens once at the attempt to "register" the mapping from identifier to Prometheus metric, and never again (not to clutter the monitoring service)

bfineran (approved):

corey-nm (approved):

PR #872, opened 2023-01-20
pin opencv-python==4.6.0.66

corey-nm:
There are issues with the latest opencv-python as you can see in https://github.com/opencv/opencv-python/issues/772.

This PR pins opencv-python to the last working version, as suggested in the above PR.

# Testing plan

Ran pip install before this PR to verify that opencv `4.7.0.68` is installed, and after this PR `4.6.0.66` is installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -130,11 +130,11 @@ _onnxruntime_deps = [
 ]
 _yolo_integration_deps = [
     "torchvision>=0.3.0,<=0.13",
-    "opencv-python",
+    "opencv-python==4.6.0.66",
 ]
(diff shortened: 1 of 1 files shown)


dbogunowicz:
@corey-nm @derekk-nm  have we experienced any problems already because of this?

corey-nm:
> @corey-nm @derekk-nm have we experienced any problems already because of this?

@dbogunowicz pinged you in the asana ticket

dbogunowicz (approved):

bfineran (approved):

PR #873, opened 2023-01-20
[MLOps 1.5] Add `BatchResult` object to hold list-like values for logging

dbogunowicz:
For the sake of clarity, we want to move away from collecting the information derived from batches of data in a dictionary. Instead, let's collect them inside the `BatchResult` object (which we can further refer to inside the leaf loggers).

So instead of e.g. keeping label of each image in the batch in a dictionary like:
`{"0": "duck, "1": "rabbit", "2": "platypus", ... }`,
we use:
`BatchResult(["duck", "rabbit", "platypus", ...])`

Handing the `BatchResult` object within the leaf loggers will be handled in a separate PR.

### DIFF
--- a/src/deepsparse/loggers/metric_functions/utils.py
+++ b/src/deepsparse/loggers/metric_functions/utils.py
@@ -12,18 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import pytest
-from deepsparse.loggers.metric_functions.natural_language_processing import (
-    string_length,
-)
+__all__ = ["BatchResult"]
 
 
-@pytest.mark.parametrize(
-    "string, expected_len",
-    [
-        ("His palms are sweaty", 20),
-        (["knees weak", "arms are heavy"], [10, 14]),
-    ],
-)
-def test_string_length(string, expected_len):
-    assert string_length(string) == expected_len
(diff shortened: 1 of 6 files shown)


src/deepsparse/loggers/metric_functions/utils.py
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+__all__ = ["BatchResult"]
+
+
+class BatchResult(list):
```
bfineran, line 18: sick 👌

bfineran: especially like this because it gives loggers a reasonable fallback (logging raw batch list) if an implementation for `BatchResult` isn't added

bfineran (approved):

corey-nm (approved):
lgtm, nice idea of subclassing list. Out of curiosity why are we moving away from dictionaries? Is it easier to log lists?

PR #874, opened 2023-01-23
[MLOps 1.5] [PrometheusLogger] Refactor naming convention

dbogunowicz:
This pull request updates the metric naming scheme in Prometheus to conform to best practices by adding labels, a prefix, and suffixes. 

The following changes are made:

- A "deepsparse_" prefix is added to every Prometheus metric to represent the namespace.
- Prediction latency metrics have a "_seconds" suffix.
- Memory available - "_bytes"
- Cores used - "_total"
- Other identifiers are also adjusted accordingly
- Adding `model_name` as a label is a bit tricky. It has been implemented separately here: #884 

## Examples of new naming convention in Prometheus
(please disregard the erroneous attribution of `model_name`, to be fixed)
### `config.yaml`
```bash
loggers:
    prometheus:
      port: 8001
system_logging:
    prediction_latency:
        enable: true
    resource_utilization:
        enable: true
    request_details:
        enable: true
num_cores: 1
num_workers: 1
endpoints:
    - task: question_answering
      model: zoo:nlp/question_answering/distilbert-none/pytorch/huggingface/squad/pruned80_quant-none-vnni
      data_logging:
        pipeline_outputs.score:
          - func: identity
```
### logging `num_cores`
```bash
# HELP deepsparse_question_answering__0__resource_utilization__num_cores_total summary metric for identifier: question_answering-0/resource_utilization/num_cores_total | Category: MetricCategories.SYSTEM
# TYPE deepsparse_question_answering__0__resource_utilization__num_cores_total summary
deepsparse_question_answering__0__resource_utilization__num_cores_total_count{model_name="num_cores_total"} 1.0
deepsparse_question_answering__0__resource_utilization__num_cores_total_sum{model_name="num_cores_total"} 1.0
# HELP deepsparse_question_answering__0__resource_utilization__num_cores_total_created summary metric for identifier: question_answering-0/resource_utilization/num_cores_total | Category: MetricCategories.SYSTEM
# TYPE deepsparse_question_answering__0__resource_utilization__num_cores_total_created gauge
deepsparse_question_answering__0__resource_utilization__num_cores_total_created{model_name="num_cores_total"} 1.6745612586558185e+09
```

### logging `prediction_latency`
```bash
# HELP deepsparse_question_answering__0__prediction_latency__pre_process_seconds summary metric for identifier: question_answering-0/prediction_latency/pre_process_seconds | Category: MetricCategories.SYSTEM
[...]

PR #875, opened 2023-01-23
[MLOps 1.5] [PrometheusLogger] Refactor `PrometheusLogger` (without changing any logic) to be better suited for the incoming PRs

dbogunowicz:
Cleans up the structure of `PrometheusLogger` to integrate easily with the upcoming PRs:
- #871 
- #874

### DIFF
--- a/src/deepsparse/loggers/prometheus_logger.py
+++ b/src/deepsparse/loggers/prometheus_logger.py
@@ -22,7 +22,7 @@ from collections import defaultdict
 from typing import Any, Optional
 
 from deepsparse.loggers import BaseLogger, MetricCategories
-from deepsparse.loggers.helpers import unwrap_logs_dictionary
+from deepsparse.loggers.helpers import unwrap_logged_value
 
 
 try:
--- a/src/deepsparse/loggers/prometheus_logger.py
+++ b/src/deepsparse/loggers/prometheus_logger.py
@@ -22,6 +22,7 @@ from collections import defaultdict
 from typing import Any, Optional
 
 from deepsparse.loggers import BaseLogger, MetricCategories
+from deepsparse.loggers.helpers import unwrap_logs_dictionary
 
 
 try:
@@ -97,8 +98,9 @@ class PrometheusLogger(BaseLogger):
         :param value: The data structure that the logger is logging
         :param category: The metric category that the log belongs to
         """
-        prometheus_metric = self._get_prometheus_metric(identifier, category)
-        prometheus_metric.observe(self._validate(value))
+        for identifier, value in unwrap_logs_dictionary(value, identifier):
+            prometheus_metric = self._get_prometheus_metric(identifier, category)
+            prometheus_metric.observe(self._validate(value))
         self._export_metrics_to_textfile()
 
     def _get_prometheus_metric(
--- a/src/deepsparse/loggers/prometheus_logger.py
+++ b/src/deepsparse/loggers/prometheus_logger.py
@@ -18,7 +18,7 @@ Implementation of the Prometheus Logger
 import logging
 import os
 from collections import defaultdict
-from typing import Any, Optional, Dict
+from typing import Any, Optional, Dict, Type
 
 from deepsparse.loggers import (
     REQUEST_DETAILS_PREFIX,
@@ -37,7 +37,7 @@ try:
         Summary,
         start_http_server,
         write_to_textfile,
-        CollectorRegistry
+        CollectorRegistry,
     )
 
     prometheus_import_error = None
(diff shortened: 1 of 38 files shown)


src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -136,14 +136,51 @@ def _setup_client(self):
...
+    :param category: The metric category that the log belongs to
+    :param registry: The Prometheus registry to which the metric should be added
+    :param description_template: The template for the description of the metric
+    :return: The Prometheus metric object or None if the identifier not supported
+    """
+    metric = Summary
+
+    return metric(
+        format_identifier(identifier),
+        description_template.format(
```
bfineran, line 166: do we expect this value to be overwritten? if so, then it would need to be overwritten with the exact same template variables or this will break

bfineran (commented):
LGTM pending comment

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: No, this value should not be overwritten and the logic protects it from getting overwritten.

In general, whenever the `PrometheusLogger` receives a log (identifier) it will:
a) Either not find it in the "registry" (self._prometheus_metrics) that maps identifiers to Prometheus metrics and go through the appropriate composition of functions to create a metric and save it to the "registry". The composition of functions in that case would be: `self.log` -> `self._get_prometheus_metric` -> `self._add_metric_to_registry` -> `get_prometheus_metric`
b) Find it in the "registry", if it has been previously seen. Then the logic only goes through `self.log` -> `self._get_prometheus_metric`

bfineran: if it should not be overwritten, then why do we have a variable for it in the function rather than reference the constant `_DESCRIPTION` directly? also let's rename to `_METRIC_DESCRIPTION_TEMPLATE`

dbogunowicz: Good point

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -27,24 +28,31 @@
...
     from prometheus_client import (
         REGISTRY,
+        CollectorRegistry,
         Summary,
         start_http_server,
         write_to_textfile,
     )
 
     prometheus_import_error = None
 except Exception as prometheus_import_err:
```
corey-nm, line 38: Unrelated to this PR, but we should move this try/except to import in loggers/__init__.py instead. That way we just can set PrometheusLogger to None, and this file can safely assume prometheus is installed

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -90,20 +98,24 @@ def log(self, identifier: str, value: Any, category: MetricCategories):
...
-            )
-            prometheus_metric = self._prometheus_metrics.get(formatted_identifier)
-            if prometheus_metric is None:
-                prometheus_metric = self._add_metric_to_registry(
-                    formatted_identifier, category
-                )
+            prometheus_metric = self._get_prometheus_metric(identifier, category)
             prometheus_metric.observe(self._validate(value))
-            self._export_metrics_to_textfile()
+        self._export_metrics_to_textfile()
```
corey-nm, line 104: 👀  🤣

corey-nm (dismissed):
lgtm!

bfineran (approved):

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: will address in a separate PR

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -27,24 +28,31 @@
...
-    Summary = None
-    start_http_server = None
-    write_to_textfile = None
+    (
+        REGISTRY,
+        Summary,
+        CollectorRegistry,
+        start_http_server,
+        write_to_textfile,
+    ) = None
```
rahul-tuli, line 45: I don't this is valid and will raise an error mostly:
```python
>>> (a, b) = None
  File "<stdin>", line 1, in <module>
TypeError: cannot unpack non-iterable NoneType object
```

src/deepsparse/loggers/prometheus_logger.py
```diff
@@ -138,14 +138,49 @@ def _setup_client(self):
...
+        registry=registry,
+    )
+
+
+def format_identifier(identifier: str) -> str:
+    """
+    Replace forbidden characters with `__` so that the identifier
+    digested by prometheus adheres to
+    https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+    :param identifier: The identifier to be formatted
```
rahul-tuli, line 178: nit: newline missing

rahul-tuli (dismissed):

src/deepsparse/loggers/prometheus_logger.py
dbogunowicz: Very tru!

corey-nm (approved):

bfineran (approved):

PR #876, opened 2023-01-23
[YOLOV8] Core implementation

dbogunowicz:
### Install dependencies
`pip install "deepsparse[yolov8]`

### Get a sample, non-sparsified model (need to have `ultralytics` installed
`yolo task=detect mode=export model=yolov8n.pt format=onnx opset=13`

### Run pipeline (annotation function)
`python annotate.py --source basilica.jpg  --model_filepath yolov8n.onnx`
OR 
`deepsparse.yolov8.annotate --source /home/ubuntu/damian/deepsparse/src/deepsparse/yolo/sample_images/basilica.jpg --model_filepath yolov8n.onnx`


bfineran (commented):
did we look into any way we could potentially share the pipeline and infer on the spot which NMS function to route to? additionally what are the changes in the annotate script? will get get the scaling introduced in #881?

bfineran (commented):
LGTM overall though, main concern is if we need a new "integration" and pipeline or can reuse more especially since so little changes. comes down to what we want out of the user experience

PR #881, opened 2023-01-24
Update: YOLO pipeline to return scaled Boxes

rahul-tuli:
This PR updates YOLO pipeline to return bounding boxes scaled with respect to the original image rather than preprocessed image.

The solution included two changes:

 - Maintaining and propagating information on the original image sizes, this was done by adding a return value to the `_preprocess` function(the original image shape is returned) and propagating it to `postprocessing` function via the `postprocessing_kwargs`
 
 - Actual rescaling of the bounding boxes based on model input shape and original input shape
 
 
 code to visualize bounding boxes:
 
 ```python
from deepsparse.utils.annotate import load_image
from deepsparse.yolo import annotate_image
from matplotlib import pyplot as plt


image_path = "./tests/deepsparse/pipelines/sample_images/basilica.jpg"
pipeline = Pipeline.create(task="yolo")
_, original_img = load_image(image_path, image_size=(640, 640))

pipeline_output = pipeline(images=[original_img])
annotated_img = annotate_image(image=original_img, prediction=pipeline_output)
plt.imshow(annotated_img)
plt.show()
 ```

- Image Before this PR: 
before

- Image after this PR:
 after

Note: The failing test, seems to be from the [following issue](https://bugs.launchpad.net/ubuntu/+source/distro-info/+bug/2003583), but I haven't taken a deeper look yet

CONTAINS:
 - [X] #882 
 - [X] #883 

---
  - https://app.asana.com/0/0/1203759380764633


corey-nm (dismissed):
Nice fixes and write up!

src/deepsparse/yolo/pipelines.py
```diff
@@ -197,9 +203,25 @@ def process_inputs(
             iou_thres=inputs.iou_thres,
             conf_thres=inputs.conf_thres,
             multi_label=inputs.multi_label,
+            original_image_shapes=original_image_shapes,
         )
         return [image_batch], postprocessing_kwargs
 
+    def _scale_boxes(
+        self, boxes: numpy.ndarray, original_image_shape: Optional[Tuple[int, ...]]
+    ):
```
dbogunowicz, line 212:
```suggestion
    ) -> numpy.ndarray:
```

src/deepsparse/yolo/pipelines.py
```diff
@@ -245,8 +268,15 @@ def process_engine_outputs(
...
-            batch_boxes.append(image_output[:, 0:4].tolist())
+        original_image_shapes = kwargs.get("original_image_shapes")
+        for idx, image_output in enumerate(batch_output):
+            boxes = image_output[:, 0:4]
+            if original_image_shapes:
+                boxes = self._scale_boxes(
+                    boxes=boxes,
+                    original_image_shape=original_image_shapes[idx],
+                )
+            batch_boxes.append(boxes.tolist())
```
dbogunowicz, line 279: I'd rather not clutter the "main" logic here and just change one single line to:
```suggestion
            batch_boxes.append(self.scale_boxes(image_output[:, 0:4], original_image_shapes = ....).tolist())
```

src/deepsparse/yolo/pipelines.py
```diff
@@ -197,9 +203,25 @@ def process_inputs(
             iou_thres=inputs.iou_thres,
             conf_thres=inputs.conf_thres,
             multi_label=inputs.multi_label,
+            original_image_shapes=original_image_shapes,
         )
         return [image_batch], postprocessing_kwargs
 
+    def _scale_boxes(
```
dbogunowicz, line 210: Since we are planning to have this function in multiple pipelines, let's make it public and put docstring on it

src/deepsparse/yolo/pipelines.py
```diff
@@ -197,9 +203,25 @@ def process_inputs(
...
+    ):
+        if not original_image_shape:
+            return boxes
+
+        scale = numpy.flipud(
+            numpy.divide(
+                numpy.asarray(original_image_shape), numpy.asarray(self.image_size)
+            )
+        )
+        scale = numpy.concatenate([scale, scale])
```
dbogunowicz, line 221: I think it may not be obvious for the arbitrary reader why we did this :)

src/deepsparse/yolo/pipelines.py
rahul-tuli: Let's start by keeping this here, eventually when we move this out into a parent pipeline I think it would make sense to make it public, also a pipeline importing a function from another pipeline (at the same level) seems off. Are you suggesting to move this out of the `YOLOPipeline` class altogether and into something like utils?

src/deepsparse/yolo/pipelines.py
rahul-tuli: I'll add in a comment

src/deepsparse/yolo/pipelines.py
```diff
@@ -185,6 +185,12 @@ def process_inputs(
 
         image_batch = list(self.executor.map(self._preprocess_image, inputs.images))
 
+        original_image_shapes = None
+        if image_batch and isinstance(image_batch[0], tuple):
```
dbogunowicz, line 189: When would `image_batch` be a tuple?

rahul-tuli: Each item in image batch list would now be a tuple since `_preprocess_image` is now also returning the original input shape

src/deepsparse/yolo/pipelines.py
```diff
@@ -197,9 +203,28 @@ def process_inputs(
...
+            )
+        )
+
+        # broadcast scales to rescale bounding boxes in one numpy operation
+        scale = numpy.concatenate([scale, scale])
+
+        boxes = numpy.multiply(boxes, scale)
+        return boxes
+
     def _preprocess_image(self, image) -> numpy.ndarray:
```
dbogunowicz, line 228: okay I missed this :)

```suggestion
    def _preprocess_image(self, image) -> Tuple[numpy.ndarray, Tuple[int, int]]:
```

src/deepsparse/yolo/pipelines.py
```diff
@@ -209,11 +234,12 @@ def _preprocess_image(self, image) -> numpy.ndarray:
             image = cv2.imread(image)
 
         image = self._make_channels_last(image)
+        original_image_shape = image.shape[:2]
```
dbogunowicz, line 237: At this point, image can be 4D and thus you would get `original_image_shape` (B, H, W) not (H,W). Maybe let's move it a bit lower in the logic?

rahul-tuli: I think you meant we'll get (B, H) That's a great callout, infact moving it lower also won't fix the problem, I'll add in a if condition for this

src/deepsparse/yolo/pipelines.py
rahul-tuli: Updated

src/deepsparse/yolo/pipelines.py
```diff
@@ -197,10 +203,29 @@ def process_inputs(
...
+        if not original_image_shape:
+            return boxes
+
+        scale = numpy.flipud(
+            numpy.divide(
+                numpy.asarray(original_image_shape), numpy.asarray(self.image_size)
+            )
+        )
+
+        # broadcast scales to rescale bounding boxes in one numpy operation
```
dbogunowicz, line 222:
```suggestion
        # scale is originally np.array([x_scale, y_scale]), needs to be np.array([x_scale, y_scale, x_scale, y_scale])
        # to properly scale bbox with shape (batch, 4)
```

src/deepsparse/yolo/pipelines.py
```diff
@@ -245,8 +275,17 @@ def process_engine_outputs(
...
-            batch_boxes.append(image_output[:, 0:4].tolist())
+        original_image_shapes = kwargs.get("original_image_shapes")
+        for idx, image_output in enumerate(batch_output):
+            original_image_shape = (
+                original_image_shapes[idx] if idx < len(original_image_shapes) else None
+            )
+            batch_boxes.append(
+                self._scale_boxes(
+                    boxes=image_output[:, 0:4],
+                    original_image_shape=original_image_shape,
```
dbogunowicz, line 286: I would still remove as much of boilerplate from this logic here and let _scale_boxes do the logic from lines L278 to 282

dbogunowicz (approved):
LGTM, left some nits

src/deepsparse/yolo/pipelines.py
rahul-tuli: I don't think that would improve the readability of the code, will leave this as is for now if not a blocker?

dbogunowicz (approved):

dbogunowicz:
@rahul-tuli shall we land it?

PR #882, opened 2023-01-24
Update: YOLO Annotation flow to work with rescaled bounding boxes

rahul-tuli:
(Initially YOLO pipeline was returning bounding boxes based off the preprocessed input image, now it has been updated to return bounding boxes w.r.t the original input image)

This PR updates yolo annotation flow to work with rescaled bounding boxes from yolo-pipeline putputs.

The changes needed were three fold:
- Update: All loaders to only resize an input image if a truthy `image_size` is passed, the reason being that the updated pipeline now does it's own book-keeping on the original input sizes, thus we don't need to explicitly resize each image before inputting it to the pipeline
 
- Update: `annotate_image` signature to not use `model_input_size` as yolo-pipeline can implicitly take care of scaling now

- Remove: `--image-shape` argument as it's actually the expected model input shape which can be inferred from the graph; Also this argument was not actually being used to override the onnx graph but just provided redundant information


src/deepsparse/yolo/annotate.py
```diff
@@ -114,15 +111,6 @@
...
 )
-@click.option(
-    "--image_shape",
-    "--image-shape",
-    type=int,
-    nargs=2,
-    default=(640, 640),
-    help="Image shape to use for inference, must be two integers",
-    show_default=True,
-)
```
mgoin, line 125: i personally would still like to change the image shape of the model/pipeline as this can affect accuracy and performance

rahul-tuli: This argument really wasn't doing anything [as in the model shape wasn't being overidden]; it was only being used to calculate scales which the pipeline does implicitly now; 

if we need support for model shape overiding I think that should be a separate PR. Does that make sense?

rahul-tuli: @markurtz

bfineran: @rahul-tuli this is a bug then - the `image-shape` arg should be passed to the pipeline as the shape we run inference at. let's get this added back in, with image shape propagated correctly

rahul-tuli: Support Added Here: https://github.com/neuralmagic/deepsparse/pull/883

rahul-tuli: Support merged!

mgoin (approved):

dbogunowicz (commented):
nice!

PR #883, opened 2023-01-24
Update: Yolo Annotation to allow shape overrides

rahul-tuli:
This PR adds in support to override onnx graph shape for yolo annotation flow:

The update needed two minor changes:
 - Accepting a 2 integer tuple representing image shape to override model input shapes with; added `--model_input_shape` cli argument for that

- Propagating the new model input shape to the yolo pipeline constructor via the `image_size` argument; the actual code to override the shapes is already in place in the YOLO pipeline.

Command:
```bash
deepsaprse.object_detection.annotate \
    --source ./tests/deepsparse/pipelines/sample_images/basilica.jpg \
    --save-dir /home/rahul/projects/deepsparse/results \
    --model_input_shape 1280 1280
```

A breakpoint was set after pipeline creation and the model shape was checked to conform with the passed in `--model_input_shape`

### DIFF
--- a/src/deepsparse/yolo/annotate.py
+++ b/src/deepsparse/yolo/annotate.py
@@ -23,13 +23,16 @@ Options:
                                   used for annotation  [default: zoo:cv/detect
                                   ion/yolov5-s/pytorch/ultralytics/coco/pruned
                                   -aggressive_96]
-  --source TEXT                   File path to an image or directory of image
+  --source TEXT                   File path to image or directory of .jpg
                                   files, a .mp4 video, or an integer (i.e. 0)
                                   for webcam  [required]
   --engine [deepsparse|onnxruntime|torch]
                                   Inference engine backend to run on. Choices
                                   are 'deepsparse', 'onnxruntime', and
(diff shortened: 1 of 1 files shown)


src/deepsparse/yolo/annotate.py
```diff
@@ -111,6 +113,15 @@
     help="Inference engine backend to run on. Choices are 'deepsparse', "
     "'onnxruntime', and 'torch'. Default is 'deepsparse'",
 )
+@click.option(
+    "--model_input_shape",
+    "--model-input-shape",
```
bfineran, line 118: model input shape is a bit misleading here because the model input shape would be the tensor shape (ie (1,3,640,640)) but we're looking for an input shape

corey-nm (approved):
lgtm!

src/deepsparse/yolo/annotate.py
rahul-tuli: Updated name to `--model_input_image_shape`

bfineran (approved):

PR #884, opened 2023-01-25
[MLOps 1.5] [PrometheusLogger] Propagate **kwargs and loggers (and thus enable `pipeline_name` label in `PrometheusLogger`)

dbogunowicz:
Making sure that we can propagate **kwargs through loggers


src/deepsparse/pipeline.py
```diff
@@ -760,6 +760,7 @@ def log(
             identifier=identifier,
             value=value,
             category=category,
+            pipeline_name=self._identifier(),
```
corey-nm, line 763: Can this just be added as an explicit arg instead of adding kwargs? Are all loggers expected to handle this argument?

dbogunowicz: Yeah, this is the crux. Right now, this is only done to improve UX on the `PrometheusLogger` side. If more loggers were depending on this variable, this could probably become an explicit arg; this is not the case right now tho

corey-nm (approved):
lgtm!

bfineran (approved):
LGTM - would be good to have some sort of test to pass a dummy kwarg though our top level logger (built from config) and ensure it's propagated all the way a leaf

dbogunowicz:
@bfineran good call. the test has revealed that there were indeed some missing pieces in the logic

tests/deepsparse/loggers/test_prometheus_logger.py
```diff
@@ -80,8 +80,8 @@ def test_get_prometheus_metric(identifier, category, registry, expected_metric):
     "identifier, no_iterations, value, text_log_save_frequency, should_fail",
     [
         ("dummy_pipeline/dummy.identifier_1", 20, 1.0, 1, False),
-        ("dummy_pipeline/dummy.identifier_2", 20, 1, 5, False),
-        ("dummy_pipeline/dummy.identifier_3", 20, [1.0], 10, True),
+        # ("dummy_pipeline/dummy.identifier_2", 20, 1, 5, False),
```
dbogunowicz, line 83:
```suggestion
        ("dummy_pipeline/dummy.identifier_2", 20, 1, 5, False),
```

tests/deepsparse/loggers/test_prometheus_logger.py
```diff
@@ -81,7 +81,7 @@ def test_get_prometheus_metric(identifier, category, registry, expected_metric):
     [
         ("dummy_pipeline/dummy.identifier_1", 20, 1.0, 1, False),
         ("dummy_pipeline/dummy.identifier_2", 20, 1, 5, False),
-        ("dummy_pipeline/dummy.identifier_3", 20, [1.0], 10, True),
+        # ("dummy_pipeline/dummy.identifier_3", 20, [1.0], 10, True),
```
dbogunowicz, line 84:
```suggestion
        ("dummy_pipeline/dummy.identifier_3", 20, [1.0], 10, True),
```

corey-nm (approved):

bfineran (approved):

PR #885, opened 2023-01-25
[MLOps 1.5] [PrometheusLogger] Account for the presence of `BatchList` implementation

dbogunowicz:
Make sure that the changes introduced in #873 are properly digested by the logic in `PrometheusLogger`


src/deepsparse/loggers/helpers.py
```diff
@@ -62,14 +62,16 @@ def unwrap_logs_dictionary(
     :param seperator: The seperator to use when composing the parent and child
         identifiers
     :return: A generator that:
-        - if `value` is not a dictionary:
+        - if `value` is a dictionary:
```
bfineran, line 65: might want to rebrand this function in general to `unwrap_logged_value` since we also do special list logic.

src/deepsparse/loggers/helpers.py
```diff
@@ -80,7 +82,10 @@ def unwrap_logs_dictionary(
                 if parent_identifier
                 else child_identifier
             )
-            if isinstance(child_value, dict):
+            if isinstance(child_value, list):
```
bfineran, line 85: do we always generically want to unwrap lists? Should only be if it's a `BatchResult`, no?

bfineran (changes requested):

corey-nm (approved):
👍

bfineran (approved):

PR #894, opened 2023-01-27
[MLOps 1.5] Register all the group names

dbogunowicz:
Add decorators to all the remaining built-ins
Remove two end2end tests, they will be replaced by a set of more thorough ones (in the next PR)


src/deepsparse/loggers/metric_functions/registry.py
```diff
@@ -46,14 +47,15 @@ def some_function(image):
 
     def decorator(function):
         identifier_registry = None
-        task_registry = DATA_LOGGING_REGISTRY.get(group)
```
dbogunowicz, line 49: Small refactor of the original `register`. Enables us to use a single decorator to map same identifier to a list of groups, to save lines of code.

so instead of 
```python
@register(group=object_detection", identifier="pipeline_outputs.scores")
@register(group="segmentation", identifier="pipeline_outputs.scores")
def func(...)
```

```python
@register(group=["object_detection", "segmentation"], identifier="pipeline_outputs.scores")
def func(...)
```

bfineran (approved):
LGTM. do we have a setup for automatically testing the application of the predefined functions against the registered pipelines?

dbogunowicz:
Will post a PR for it today.

KSGulin (approved):
Nice! Glad to see this redundancy getting reduced

PR #902, opened 2023-02-06
[MLOps 1.5] Refactor the constants that contain a set of `SystemGroups`

dbogunowicz:
At the current state, manipulating and moving around the system group names is a bit tedious. This is because they exist as separate values. This PR wraps them inside a `dataclass` object.

### DIFF
--- a/src/deepsparse/loggers/constants.py
+++ b/src/deepsparse/loggers/constants.py
@@ -15,19 +15,17 @@ Holds logging-related objects with constant values
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from dataclasses import dataclass
 from enum import Enum
 
 
 __all__ = [
     "MetricCategories",
     "validate_identifier",
-    "REQUEST_DETAILS_IDENTIFIER_PREFIX",
-    "RESOURCE_UTILIZATION_IDENTIFIER_PREFIX",
+    "SystemGroups",
 ]
 
 UNSUPPORTED_IDENTIFIER_CHARS = {".", "[", "]"}
-REQUEST_DETAILS_IDENTIFIER_PREFIX = "request_details"
-RESOURCE_UTILIZATION_IDENTIFIER_PREFIX = "resource_utilization"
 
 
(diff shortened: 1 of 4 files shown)


src/deepsparse/loggers/constants.py
```diff
@@ -43,6 +41,15 @@ class MetricCategories(Enum):
     DATA = "data"
 
 
+@dataclass(frozen=True)
+class SystemGroups:
+    # Pipeline System Groups
+    PREDICTION_LATENCY: str = "prediction_latency"
+    # Server System Groups
+    REQUEST_DETAILS: str = "request_details"
+    RESOURCE_UTILIZATION: str = "resource_utilization"
```
rahul-tuli, line 50: Thought, maybe instead of adding a comment that also needs to be maintained, would renaming to `PIPELINE_PREDICTION_LATENCY`, AND `SERVER_REQUEST_DETAILS` be more meaningful?

dbogunowicz: I do not like to make those variables long tbh. We will be composing quite complex strings using those names, my focus is to keep them short.

bfineran: caution against too short, but this looks like a good balance

bfineran (approved):

src/deepsparse/loggers/constants.py
rahul-tuli: Approving for now, but my main concern is readability and context over short variable names, if someone wants to check which system group belongs to Pipeline they would have to come to this class and read a comment which could quickly go out of date.

rahul-tuli: Also if these are constants, wouldn't an Enum be better?

rahul-tuli (approved):

src/deepsparse/loggers/constants.py
dbogunowicz: I hate using Enums for the awkward process of accessing them ;)

PR #903, opened 2023-02-06
[transformers] do not ignore any labels in conll eval pipeline

bfineran:
fixes a bug where outputs ignored by the default `ignore_labels` value would cause a size mismatch with the expected labels length during metrics computation

**test_plan:**
verified manually. research team to QA


dbogunowicz (approved):
So by default zeros were being omitted and this caused the issue?

KSGulin (approved):

PR #909, opened 2023-02-08
Update: Open Pif Paf callables

rahul-tuli:
The Open Pif Paf callables in `setup.py` were pointing to incorrect location this PR fixes that, and also adds `--model_filepath` in example commands, (it is required)


dbogunowicz (approved):
Sweet, thanks for that!

corey-nm (approved):

PR #916, opened 2023-02-21
[OpenPifPaf] Allow for returning CifCaf fields

dbogunowicz:
## Feature Preview 
```python
input = "src/deepsparse/yolo/sample_images/basilica.jpg"
pipeline = Pipeline.create(task="open_pif_paf", model_path="openpifpaf-resnet50.onnx", return_cifcaf_fields=True)
output = pipeline(images=input)
```
image

## Feature Description

OpenPifPaf is comprised of two algorithms.
- prediction of CifCaf fields (using a parametrized neural network)
- matching CifCaf fields to actual poses (using a deterministic, algorithm)

I would like to give users an option not only to receive decoded poses (final representation) but also to fetch CifCaf fields (intermediate representation).
The motivation is for easy validation (my validation use case) or the ability to use the pipeline as a component of a larger system (that requires the CifCaf fields and not necessarily the decoded poses).

### DIFF
--- a/src/deepsparse/open_pif_paf/schemas.py
+++ b/src/deepsparse/open_pif_paf/schemas.py
@@ -12,9 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Tuple
+from typing import Iterable, List, TextIO, Tuple
 
 import numpy
+from PIL import Image
 from pydantic import BaseModel, Field
 
 from deepsparse.pipelines.computer_vision import ComputerVisionSchema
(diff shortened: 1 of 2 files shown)


src/deepsparse/open_pif_paf/schemas.py
```diff
@@ -33,6 +35,26 @@ class OpenPifPafInput(ComputerVisionSchema):
...
+    """
+    Open Pif Paf is composed of two stages:
+     - Computing Cif/Caf fields using a parametrized model
+     - Applying a matching algorithm to obtain the final pose
+        predictions
+    In some cases (e.g. for validation), it may be useful to
+    obtain the Cif/Caf fields as output.
+    """
+
+    fields: List[List[numpy.ndarray]] = Field(
```
rahul-tuli, line 48: Oooo, our old friend `numpy.ndarray`; will this cause issues with serialization needed for deepsparse server? Do we need to add `Any`?

dbogunowicz: Yeah, was thinking about that.  
We can now assume that this pathway is not indented to be used broadly by the users, so we do not need the support for it in the server. Weak assumption, I know, but we can still change this in the future.

Actually, @corey-nm didn't you find some sweet hack back in the days to enable numpy serialization?

bfineran: let's see if we can get this resolved, if not, let's add an inline todo

dbogunowicz: Resolved

bfineran: What was the resolution?

dbogunowicz: If we call the openpifpaf pipeline that is set to output cifcaf in the server, ValueError will be thrown.

rahul-tuli (approved):

PR #917, opened 2023-02-21
 Fix how `preprocess_array` (openpifpaf/yolact) handles float32/64 input

dbogunowicz:
## Feature Description

`preprocess_array` is an input preprocessing function that is used by both yolact and openpifpaf (yeah, it's pretty messy, we should standardize the preprocessing for all the CV pipelines at some point). 

Back when I wrote it, I only considered that preprocessing function in the context of the annotation i.e. the input is a path to the model -> we will be effectively passing the uint8 array. Because of this explicit assumption `preprocess_array` e.g. always divides the input by 255, which is clearly incorrect if we pass float32 arrays.

This PR rectifies this error i.e. makes the function robust to both float32 and uint8 image inputs.

## Manual Testing
```python
import cv2
import numpy
input_uint8 = "src/deepsparse/open_pif_paf/sample_images/dancers.jpg"
input_float32 = cv2.imread(input_uint8).astype(numpy.float32) / 255

## Manual test for openpifpaf pipeline
pipeline = Pipeline.create(task="open_pif_paf", model_path="openpifpaf-resnet50.onnx")
output_uint8 = pipeline(images=input_uint8)
output_float32 = pipeline(images=input_float32)
# scores are close but not the same
# (1st case we resize the uint8 and then convert to float32)
# (2nd case we convert the uint8 to float32 and then resize)
assert numpy.allclose(output_float32.scores, output_uint8.scores, atol=1e-2)

## Manual test for yolact pipeline
pipeline = Pipeline.create(task="yolact")
output_uint8 = pipeline(images=input_uint8)
output_float32 = pipeline(images=input_float32)
assert numpy.allclose(output_float32[0].scores, output_uint8[0].scores, atol=1e-1)
```

### DIFF
--- a/src/deepsparse/yolov8/utils/validation/__init__.py
+++ b/src/deepsparse/yolov8/utils/validation/__init__.py
@@ -14,3 +14,5 @@
 
 # flake8: noqa
 from .detection_validator import *
+from .helpers import *
+from .segmentation_validator import *
--- a/src/deepsparse/yolov8/utils/validation/deepsparse_validator.py
+++ b/src/deepsparse/yolov8/utils/validation/deepsparse_validator.py
@@ -1,32 +1,13 @@
-# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#    http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing,
-# software distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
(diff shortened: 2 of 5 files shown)


src/deepsparse/yolact/utils/utils.py
```diff
@@ -48,21 +48,25 @@ def preprocess_array(
     :return: preprocessed numpy array (B, C, D, D); where (D,D) is image size expected
         by the network. It is a contiguous array with RGB channel order.
     """
-    image = image.astype(numpy.float32)
+
+    uint8_image = image.dtype == numpy.uint8
```
rahul-tuli, line 52: nit: rename to `is_uint8_image`?

rahul-tuli (dismissed):
Super clean!

src/deepsparse/yolact/utils/utils.py
dbogunowicz: good idea!

bfineran (dismissed):

rahul-tuli (dismissed):

bfineran (approved):
Lgtm update pr title to remove openpifpaf paf

rahul-tuli (approved):

PR #922, opened 2023-02-24
[YOLOv8-Seg] Pipeline & Annotation Script

dbogunowicz:
Annotation for YOLO:
result-0
Annotation for YOLOv8-Det
image
Annotation for YOLOv8-Seg
```bash
python src/deepsparse/yolact/annotate.py  --subtask segmentation --model_filepath yolov8n-seg.onnx --source src/deepsparse/yolo/sample_images/basilica.jpg
```
image

### DIFF
--- a/src/deepsparse/yolo/schemas.py
+++ b/src/deepsparse/yolo/schemas.py
@@ -18,8 +18,10 @@ Input/Output Schemas for Object Detection with YOLO
 """
 
 from collections import namedtuple
-from typing import List
+from typing import Iterable, List, TextIO
 
+import numpy
+from PIL import Image
 from pydantic import BaseModel, Field
 
 from deepsparse.pipelines.computer_vision import ComputerVisionSchema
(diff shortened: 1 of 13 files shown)


src/deepsparse/yolo/schemas.py
```diff
@@ -54,6 +56,39 @@ class YOLOInput(ComputerVisionSchema):
             "pathway by default run with multi_label on"
         ),
     )
+    return_masks: bool = Field(
+        default=True,
+        description="Controls whether the pipeline should additionally "
+        "return segmentation masks (if running a segmentation model",
```
dbogunowicz, line 62:
```suggestion
        "return segmentation masks (if running a segmentation model)",
```

KSGulin (approved):
Nice! Making it look easy 😎

corey-nm (approved):
lgtm, open question whether these should be separate pipelines or not, but that is not blocking imo!

PR #923, opened 2023-02-24
[YOLOv8-Seg] Refactor YOLACT annotation and pipeline

dbogunowicz:
This PR refactors the YOLACT pipeline. Noteworthy changes:
  - endow the pipeline with the ability to resize the output masks and bounding boxes so that they correspond to the original image size (reimplementation of the functionality introduced by @rahul-tuli to the YOLO pipeline)
  - removing the `_resize_to_fit_img` function from the yolact annotation function. The motivation here is to make `src/deepsparse/yolact/utils/annotate.py` annotation function easily handle both YOLACT and YOLOv8Seg outputs. In the future, when the long-awaited refactor of CV pipelines happens, we could hopefully use this function to annotate any object/instance segmentation models.

Testing:
```bash
deepsparse.instance_segmentation.annotate  --source src/deepsparse/yolact/sample_images/thailand.jpg
```
imageResult in an image that corresponds to my expectations:

### DIFF
--- a/src/deepsparse/yolact/annotate.py
+++ b/src/deepsparse/yolact/annotate.py
@@ -29,9 +29,9 @@ Options:
                                   Inference engine backend to run on. Choices
                                   are 'deepsparse', 'onnxruntime', and
                                   'torch'. Default is 'deepsparse'
-  --image_shape, --image_shape INTEGER
-                                  Image shape to use for inference, must be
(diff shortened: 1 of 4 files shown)


bfineran (approved):

corey-nm (approved):
lgtm, easy refactoring!

PR #931, opened 2023-03-02
Update README.md

Beth-Kosis:
Line 6:  Is this a heading? a link?
Line 77:  Is there info to be added for the Expected Output?
Lines 103 and 131:  Should CIF and CAF be PIF and PAF for consistency?
Line 105:  Can we eliminate the word "probably" because it implies ambiguity?


dbogunowicz:
Oh wow, I am a bit embarrassed.
this README was initially my brain dump to communicate some of the important things I found out about openpifpaf. Then I managed to beautify it to some degree, but I am suspicious about whether this is in a good state. I would personally redo it so that it looks more like our IC or transformers READMEs.

jeanniefinks:
@dbogunowicz may you reply with the URLs to the IC or transformers READMEs in question so we're looking at the ones you are citing, thanks!

dbogunowicz:
I am sorry for being a slowpoke:
https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/image_classification/README.md
https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/transformers/README.md

PR #934, opened 2023-03-03
Update Docker Infra

rahul-tuli:
This push will revamp our docker infrastructure to
make it easier to start with deepsparse

The identified goals for now are as follows:

- [x] Reduce number of build args
- [x] Create multistage Dockerfile for the following:
  - [x] All Dependencies
  - [x] Server Dependencies
  - [x] Task Specific Dependencies
  - [x] All Scenarios should have a Prod Stage (with exact pinned versions) and a Dev Stage
  - [x] Trigger auto-installs per integration 

Overarching goals:
- [ ] The docker images must be as small as possible and follow industry standards, needs more investigation
- [x] Including a HealthCheck
- [X] CI/CD to auto build and push images to GitHub Container Registry

TODO:
Once Approved, remove current branch from event triggers, maybe also remove nightly builds
- [x] Check if git checkout is needed

We can also use the specified dockerfile to build a dev container like the following:

```bash
docker build --target container_branch_dev --build-arg DEPS=dev -t deepsparse-dev-docker . && docker container run -it deepsparse-dev-docker bash
```
A successful GHA push on a dummy `release/99.99.96` branch: https://github.com/neuralmagic/deepsparse/actions/runs/4415921132/jobs/7739522073
~NOTE: GHA red due to #939~

### DIFF
--- a/.github/workflows/build-and-publish-tagged-images.yaml
+++ b/.github/workflows/build-and-publish-tagged-images.yaml
@@ -6,9 +6,6 @@ on:
       - '*'
     branches:
       - 'release/*'
-  pull_request:
-    branches:
-      - 'release/*'
 
 jobs:
   build-and-push-docker-image:
--- a/.github/workflows/build-and-publish-tagged-images.yaml
+++ b/.github/workflows/build-and-publish-tagged-images.yaml
@@ -55,7 +55,7 @@ jobs:
             VERSION=${{ steps.extract_tag.outputs.tag }}
           push: true
           tags: |
-            ghcr.io/neuralmagic/deepsparse-all:$${{ steps.extract_tag.outputs.tag }}
+            ghcr.io/neuralmagic/deepsparse-all:${{ steps.extract_tag.outputs.tag }}
 
       - name: Build and push deepsparse-server
         if: ${{ contains(github.ref, 'refs/heads/release/') }}
--- a/.github/workflows/build-and-publish-tagged-images.yaml
+++ b/.github/workflows/build-and-publish-tagged-images.yaml
@@ -1,110 +1,124 @@
 name: Build and Publish
 
 on:
-  push:
-    tags:
-      - '*'
-    branches:
-      - 'release/*'
+  create
 
 jobs:
   build-and-push-docker-image:
     name: Build and Push Version Tagged Docker Images to GitHub Container Registry
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
       - name: Set up Docker Buildx
+        if: ${{ contains(github.ref, 'refs/heads/release/') }}
         id: buildx
         uses: docker/setup-buildx-action@v2
+        with:
+          buildkitd-flags: --debug
 
       - name: Login to Github Packages
+        if: ${{ contains(github.ref, 'refs/heads/release/') }}
         uses: docker/login-action@v2
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Get branch names
-        id: branch-name
-        uses: tj-actions/branch-names@v6
+      - name: Checkout code
+        if: ${{ contains(github.ref, 'refs/heads/release/') }}
+        uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,26 +1,83 @@
...
+ARG VENV
+ENV PATH="${VENV}/bin:$PATH"
+ENV PIP_DEFAULT_TIMEOUT=200
+ARG DEEPSPARSE_COMMAND=./deepsparse[server,yolo,onnxruntime,yolov8]
+RUN $VENV/bin/pip install install --no-cache-dir "${DEEPSPARSE_COMMAND}"
 
-# if $GIT_CHECKOUT is specified - clone, checkout $GIT_CHECKOUT, and install with -e
-RUN if [ -n "${GIT_CHECKOUT}" ] ; then git clone https://github.com/neuralmagic/deepsparse.git --depth 1 -b $GIT_CHECKOUT; fi
-RUN if [ -n "${GIT_CHECKOUT}" ] ; then pip3 install --no-cache-dir --upgrade -e "./deepsparse[server]" ; fi
+RUN deepsparse.transformers.run_inference --help \
```
KSGulin, line 36: Is this intentional? It seems there's already a conditional below to run one or the other help command

.github/workflows/build-and-publish-tagged-images.yaml
```diff
@@ -0,0 +1,110 @@
...
+        if: steps.branch-name.outputs.is_tag == 'true'
+        uses: docker/build-push-action@v2
+        with:
+          context: ./docker
+          build-args: |
+            DEPS=all
+            VERSION=${{ steps.branch-name.outputs.tag }}
+          tags: |
+            ghcr.io/neuralmagic/deepsparse-all:${{ steps.branch-name.outputs.tag }}
+            ghcr.io/neuralmagic/deepsparse:${{ steps.branch-name.outputs.tag }}
```
KSGulin, line 49: What name paths do these get saved under? `deepsparse-all` makes sense to me, `deepsparse` standalone less so. And how does it tie into the user pointer to deepsparse/deepsparse-all?

KSGulin (commented):
Nice and really clean implementation. Have a few questions, but otherwise looks ready to go for me. Also, could you add the sizes for each of the images to the PR description

.github/workflows/build-and-publish-tagged-images.yaml
rahul-tuli: Both the tags point to the same image, it was decided that a vanilla deepsparse image should install all dependecies; I'll update this comment with name paths

docker/Dockerfile
rahul-tuli: Yes, this is intentional Konstantin, if DEPS=all, then only the base, container_branch_all and prod stage will be run,  

container_branch_all stage should be responsible here to trigger autoinstalls

rahul-tuli:
> Nice and really clean implementation. Have a few questions, but otherwise looks ready to go for me. Also, could you add the sizes for each of the images to the PR description

Great call out will add that info in

docker/Dockerfile
```diff
@@ -1,26 +1,84 @@
...
+ENV PATH="${VENV}/bin:$PATH"
+ENV PIP_DEFAULT_TIMEOUT=200
+ARG DEEPSPARSE_COMMAND=./deepsparse
+RUN $VENV/bin/pip install install --no-cache-dir "${DEEPSPARSE_COMMAND}"
+
+FROM base AS container_branch_dev
+ARG VENV
+ENV PATH="${VENV}/bin:$PATH"
+ENV PIP_DEFAULT_TIMEOUT=200
+ARG DEEPSPARSE_COMMAND=./deepsparse[dev]
```
bfineran, line 74: in developer mode we should be installing against a local copy of the repo, no? so either copying it from the users base machine or cloning it to allow for an editable install

rahul-tuli: Great callout, I added the -e flag, however we do not need to do anything special for a dev container, we can get that by supplying a `--target` flag like the following:

`docker build --target container_branch_dev --build-arg DEPS=dev -t deepsparse-dev-docker . && docker container run -it deepsparse-dev-docker bash`

KSGulin (approved):

bfineran (approved):
🚀

PR #939, opened 2023-03-07
[BugFix] Add  opencv to IC AutoInstalls

rahul-tuli:
Right now deepsparse ic auto installs are broken:
Steps to reproduce

```bash
python3.8 -m venv test-venv
source test-venv/bin/activate
pip install deepsparse
deepsparse.image_classification.annotate --help
```

Relevant Traceback:
```bash
  File "/home/rahul/test-venv/bin/deepsparse.image_classification.annotate", line 5, in <module>
    from deepsparse.image_classification.annotate import main
  File "/home/rahul/test-venv/lib/python3.8/site-packages/deepsparse/image_classification/annotate.py", line 72, in <module>
    import cv2
ModuleNotFoundError: No module named 'cv2'
```

This PR fixes the error by auto-installing cv2 in IC `__init__.py` file

Noting that this is a quick temporary fix as eventually we will move away from autoinstalls and rely on deepsparse[integration] to install all integration specific dependencies, however I would like to get this in, to have a working source of truth when doing that migration

### DIFF
--- a/src/deepsparse/image_classification/__init__.py
+++ b/src/deepsparse/image_classification/__init__.py
@@ -20,12 +20,23 @@ from collections import namedtuple
 
 
 _LOGGER = _logging.getLogger(__name__)
-_Dependency = namedtuple("_Dependency", ["name", "version", "necessary"])
+_Dependency = namedtuple("_Dependency", ["name", "import_name", "version", "necessary"])
 
 
 def _auto_install_dependencies():
     dependencies = [
-        _Dependency(name="torchvision", version=">=0.3.0,<=0.14.0", necessary=True),
+        _Dependency(
+            name="torchvision",
+            import_name="torchvision",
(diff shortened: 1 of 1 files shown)


src/deepsparse/image_classification/__init__.py
```diff
@@ -20,12 +20,23 @@
 
 
 _LOGGER = _logging.getLogger(__name__)
-_Dependency = namedtuple("_Dependency", ["name", "version", "necessary"])
+_Dependency = namedtuple("_Dependency", ["name", "import_name", "version", "necessary"])
```
rahul-tuli, line 23: Added this field, cause import name can differ from install name for some libraries, example `opencv-python`

bfineran (approved):

KSGulin (approved):
LGTM

dbogunowicz (approved):

PR #941, opened 2023-03-07
Benchmark UI demo

InquestGeronimo:
Example for spinning up a Gradio UI for benchmarking select SparseZoo models. Integration includes the ability to spin up DigitalOcean Droplets.


examples/benchmark-ui/elements.py
```diff
@@ -0,0 +1,102 @@
+
+class TextElements:
+    
+    # Markdown Text
+    md_title = "# Benchmarking Sparse Models on DigitalOcean's CPUs!"
+
+    md_body = '''
+
+    ## Hey There 👋
```
mgoin, line 23: maybe regular text size? there already is a title header that describes it

mgoin (commented):
Looks pretty nice! 
I think the `Droplet` class might be unnecessary but fine with keeping it if you like the UX.

examples/benchmark-ui/requirements.txt
```diff
@@ -0,0 +1,4 @@
+deepsparse-nightly
+deepsparse[onnxruntime]
```
mgoin, line 2: is the nightly needed or is 1.4 good enough for now?

examples/benchmark-ui/elements.py
InquestGeronimo: can we lower the heading size to "### " instead of normal text size? Keeping it regular text size looks a bit weird aesthetically. 🤷‍♂️

examples/benchmark-ui/requirements.txt
InquestGeronimo: still need nightly as for right now

mgoin: can you remove `deepsparse[onnxruntime]` and edit `deepsparse-nightly[onnxruntime]`

mgoin, line 2:
```suggestion
deepsparse-nightly[onnxruntime]
```

mgoin (approved):

rahul-tuli (approved):
Nice example, ran without any issues in one go! Great work @InquestGeronimo

mgoin (approved):
Also took an easy run through the local server, LGTM

Issue #937, opened 2023-03-07
Setting `aggregation_strategy` in TokenClassificationPipeline fails


mgoin:
**Describe the bug**
Setting `aggregation_strategy` in TokenClassificationPipeline other than the default `"none"` raises avalidation error.

**Expected behavior**

**Environment**
Include all relevant environment information:
1. OS [e.g. Ubuntu 18.04]: Ubuntu 18.04.6 LTS
2. Python version [e.g. 3.7]: Python 3.8.0
3. DeepSparse version or commit hash [e.g. 0.1.0, `f7245c8`]: DeepSparse 1.5.0.20230301
4. ML framework version(s) [e.g. torch 1.7.1]: transformers-4.23.1

**To Reproduce**
Default `"none"` works fine:
```
>>> from deepsparse import Pipeline
>>> p_simple = Pipeline.create(task="ner")
>>> p_none("Mary is flying from Nairobi to New York")
TokenClassificationOutput(predictions=[[TokenClassificationResult(entity='B-PER', score=0.9990960955619812, index=1, word='mary', start=0, end=4, is_grouped=False), TokenClassificationResult(entity='B-LOC', score=0.9995313286781311, index=5, word='nairobi', start=20, end=27, is_grouped=False), TokenClassificationResult(entity='B-LOC', score=0.9995684027671814, index=7, word='new', start=31, end=34, is_grouped=False), TokenClassificationResult(entity='I-LOC', score=0.9975587129592896, index=8, word='york', start=35, end=39, is_grouped=False)]])
```

Using `"simple"` or another like `"first"` fails:
```
>>> p_simple = Pipeline.create(task="ner", aggregation_strategy="simple")
>>> p_simple("Mary is flying from Nairobi to New York")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mgoin/slow/venvs/test/lib/python3.8/site-packages/deepsparse/pipeline.py", line 295, in __call__
    pipeline_outputs = self.process_engine_outputs(
  File "/home/mgoin/slow/venvs/test/lib/python3.8/site-packages/deepsparse/transformers/pipelines/token_classification.py", line 351, in process_engine_outputs
    current_results.append(TokenClassificationResult(**entity))
  File "pydantic/main.py", line 342, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for TokenClassificationResult
index
  field required (type=value_error.missing)
```

[...]

PR #943, opened 2023-03-07
[token classification] only require 'index' on output when aggregation_strategy=none

bfineran:
`index` is only supplied on output of token classification pipelines when the aggregation strategy is None. This caused a pydantic bug due to `index` being required. This PR makes `index` not required.

HF source on the topic: https://github.com/huggingface/transformers/blob/8abe4930d39d9fb379fe89ee19327d371daee29f/src/transformers/pipelines/token_classification.py#L202

fixes: #937 

**test_plan:**
manual


mgoin (approved):
Thanks!

rahul-tuli (approved):

PR #961, opened 2023-03-18
Integrating Batch Inference to AWS Lambda example

InquestGeronimo:
Adding batch inference capability to the current aws-lambda example (now named aws-serverless).


InquestGeronimo:
image is a placeholder for now...

bfineran (commented):
LGTM overall, do you plan on adding support for batch size selection? or is the plan to use the batch size 1/dynamic built in to run the batch job? Wondering if this is the best for performance

PR #963, opened 2023-03-20
[Fix] Make `test_logging_only_system_info` flaky

dbogunowicz:
Well, once in a blue moon, simple tests for server logging fail. I am moderately sure that this is because the test completes before the log "gets" (asynchronously) to the leaf logger. This is a very rare event, a flaky test should account for that.

I think that alternatively, we could put the "sleep" function before `_test_logger_contents(...)`, but I think that making the test flaky is more elegant.

### DIFF
--- a/tests/server/test_loggers.py
+++ b/tests/server/test_loggers.py
@@ -26,6 +26,7 @@ from deepsparse.server.config import (
 from deepsparse.server.helpers import server_logger_from_config
 from deepsparse.server.server import _build_app
 from fastapi.testclient import TestClient
+from flaky import flaky
 from tests.deepsparse.loggers.helpers import fetch_leaf_logger
 from tests.helpers import find_free_port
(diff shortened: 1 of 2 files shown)


tests/server/test_loggers.py
```diff
@@ -98,6 +99,7 @@ def test_data_logging_from_predefined():
         assert log == expected_log
 
 
+@flaky(max_runs=4, min_passes=3)
```
bfineran, line 102: how long does this test take? any reason not to do 2:1?

bfineran (approved):
LGTM - let's monitor if failures get more consistent we should investigate further

tests/server/test_loggers.py
dbogunowicz: It's a quick one, a few ms.
0.75% is more tough to pass then 0.5% ;)

PR #964, opened 2023-03-20
[Fix] Explicitly specify the shape for the benchmarking model

dbogunowicz:
Because the onnx model had at least one dynamic axis (batch, h,w), it was failing in the benchmarking logic.
I fixed the test plus added a warning to avoid using models with dynamic shapes in `generate_random_inputs`.

### DIFF
--- a/src/deepsparse/utils/onnx.py
+++ b/src/deepsparse/utils/onnx.py
@@ -214,6 +214,11 @@ def generate_random_inputs(
         input_tensor_type = external_input.type.tensor_type
         elem_type = translate_onnx_type_to_numpy(input_tensor_type.elem_type)
         in_shape = [int(d.dim_value) for d in input_tensor_type.shape.dim]
+        if 0 in in_shape:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


src/deepsparse/utils/onnx.py
```diff
@@ -214,6 +214,11 @@ def generate_random_inputs(
         input_tensor_type = external_input.type.tensor_type
         elem_type = translate_onnx_type_to_numpy(input_tensor_type.elem_type)
         in_shape = [int(d.dim_value) for d in input_tensor_type.shape.dim]
+        if 0 in in_shape:
+            raise ValueError(
+                "Attempting to benchmark a model with dynamic shape using"
```
dbarbuzzi, line 219: nit: missing a space between these two lines would yield "… usingrandomly …".

```suggestion
                "Attempting to benchmark a model with dynamic shape using "
```

dbarbuzzi (approved):
LGTM

bfineran (approved):

dbarbuzzi (approved):

PR #978, opened 2023-03-27
Update README.md -> `log_save_frequency` in Prometheus logger 

dbogunowicz:
Addresses: https://app.asana.com/0/1201735099598270/1204260174145542/f


mgoin:
@dbogunowicz It looks like this also needs to be changed in the cli docs: https://github.com/neuralmagic/deepsparse/blob/1f147004eb772493e8a95b832860c8a620792f99/src/deepsparse/server/cli.py#L202

bfineran (approved):

mgoin (approved):
thanks!

PR #1002, opened 2023-04-17
Add experimental darwin support

bnellnm:
Enable support for Darwin in deepsparse

### DIFF
--- a/src/deepsparse/cpu.py
+++ b/src/deepsparse/cpu.py
@@ -155,10 +155,9 @@ def cpu_architecture() -> architecture:
 
     :return: an instance of the architecture class
     """
-    if not (sys.platform.startswith("linux") or
-            sys.platform.startswith("darwin")):
+    if not (sys.platform.startswith("linux") or sys.platform.startswith("darwin")):
         raise OSError(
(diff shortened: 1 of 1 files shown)


src/deepsparse/cpu.py
```diff
@@ -140,6 +143,51 @@ def _parse_arch_bin() -> architecture:
...
+        return platform.mac_ver()[0]
+    return None
+
+
+def check_darwin_support() -> bool:
+    """
+    Check if the system is running Darwin and it meets the minimum version
+    requirements.
+    """
+    if sys.platform.startswith("darwin"):
```
KSGulin, line 167: Should we be failing the darwin support check if `NM_ALLOW_DARWIN` is `False`?
```suggestion
    if allow_darwin() and sys.platform.startswith("darwin"):
```

KSGulin (approved):
The long awaited day 🙂

src/deepsparse/cpu.py
bnellnm: Good catch, I missed this while doing refactoring.

tlrmchlsmth (approved):

KSGulin (approved):

bfineran (approved):

PR #1009, opened 2023-04-24
Rs/docs update new

rsnm2:


mgoin (approved):
Reviewed each doc and I think it looks great

PR #1015, opened 2023-04-27
torchscript, cpu and gpu support

horheynm:
Torchscript integration

Example
```python3
model = "path_to_dot_pt" or nn_module or torchscipt_model
device = "cpu" or "cuda"
engine = TorchScriptEngine(model, device=device)
inp = [numpy.random.rand(1, 3, 224, 224).astype(numpy.float32)]
out = engine(inp)
```


setup.py
```diff
@@ -165,6 +165,9 @@ def _parse_requirements_file(file_path):
 _haystack_integration_deps = _parse_requirements_file(_haystack_requirements_file_path)
 
 
+_torch_deps = ["torch==2.0"]
```
bfineran, line 168: let's have the torch version match the `sparseml[torch]` supported versions

src/deepsparse/benchmark/torchscript_engine.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def _select_device(device: str):
+    device = str(device).lower()
+    if device == "cuda" and torch.cuda.is_available():
+        return "cuda"
+    return "cpu"
+
+
+def _validate_jit_model(model):
```
bfineran, line 80: nit 'private' methods to bottom of file

src/deepsparse/benchmark/torchscript_engine.py
```diff
@@ -0,0 +1,325 @@
...
+    if batch_size < 1:
+        raise ValueError("batch_size must be greater than 0")
+
+    return batch_size
+
+
+def _select_device(device: str):
+    device = str(device).lower()
+    if device == "cuda" and torch.cuda.is_available():
+        return "cuda"
```
bfineran, line 76: we should just raise our own error if a user specifies cuda and cuda is not available.

src/deepsparse/benchmark/torchscript_engine.py
```diff
@@ -0,0 +1,325 @@
...
+        self._device = _select_device(device)
+
+        if isinstance(model, torch.nn.Module):
+            self._model = torch.jit.script(model).eval()
+        elif isinstance(model, str):
+            self._model = torch.jit.load(model).eval()
+        else:
+            self._model = model
+
+        _validate_jit_model(self._model)
```
bfineran, line 120: IMO we don't need to convert `Module` objects into `jit` modules, it's fine to keep them as they are, need to update the validation accordingly (ie both `Module` and `ScriptModule` valid)

src/deepsparse/benchmark/torchscript_engine.py
```diff
@@ -0,0 +1,325 @@
...
+
+    @property
+    def output_names(self) -> List[str]:
+        """
+        :return: The ordered names of the outputs.
+        """
+        return [node_arg.name for node_arg in self._eng_net.get_outputs()]
+
+    @property
+    def output_shapes(self) -> List[Tuple]:
```
bfineran, line 212: `eng_net` doesn't exist and will cause these to fail. let's delete for now. can see what we can do for arg/shape inference later

src/deepsparse/benchmark/torchscript_engine.py
```diff
@@ -0,0 +1,325 @@
...
+
+        torch_inputs = [torch.from_numpy(input).to(self.device) for input in inp]
+
+        tensors = self._model(*torch_inputs)
+
+        if isinstance(tensors, torch.Tensor):
+            tensors = [tensors]
+        if self.device == "cuda":
+            tensors = [torch.Tensor.cpu(tensor) for tensor in tensors]
+        return [tensor.detach().numpy() for tensor in tensors]
```
bfineran, line 263: should always be safe to call `cpu` on a tensor, even if not in `cuda` mode
```suggestion
        return [tensor.cpu().detach().numpy() for tensor in tensors]
```

src/deepsparse/benchmark/torchscript_engine.py
```diff
@@ -0,0 +1,325 @@
...
+            The expected order is the inputs order as defined in the ONNX graph.
+        :param val_inp: Validate the input to the model to ensure numpy array inputs
+            are setup correctly for inference.
+        :return: The dictionary of outputs from the model after executing
+            over the inputs
+        """
+        out = self.run(inp, val_inp)
+        return zip(self._output_names, out)
+
+    def _validate_inputs(self, inp: List[numpy.ndarray]):
```
bfineran, line 310: not used

tests/conftest.py
```diff
@@ -90,3 +105,52 @@ def check_for_created_files():
         f"megabytes of temp files created in temp directory during pytest run. "
         f"Created files: {set(end_files_temp) - set(start_files_temp)}"
     )
+
+
+@pytest.fixture
+def torchvision_fixture():
```
bfineran, line 111: verify that this function is run lazily

tests/conftest.py
```diff
@@ -12,13 +12,28 @@
...
 
 import pytest
 from tests.helpers import delete_file
+from tests.utils.torch import find_file_with_pattern
+
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+try:
```
bfineran, line 30: is there any better place to put this outside of conftest? would prefer to keep closer to the place where it's used

tests/test_torchscript.py
```diff
@@ -0,0 +1,56 @@
...
+try:
+    import torch
+
+    torch_import_error = None
+except Exception as torch_import_err:
+    torch_import_error = torch_import_err
+    torch = None
+
+
+def test_cpu_torchscript(torchscript_test_setup):
```
bfineran, line 33: let's test cuda as well but add `pytest.mark.skipif` with condition of `torch.cuda.is_available`

bfineran (commented):
LGTM pending comments - additionally, let's update PR description to include scope of feature as well as example snippet

future work should add zoo support

bfineran (approved):

PR #1018, opened 2023-05-01
default num_cores in `benchmark_model` fn

bfineran:
`num_cores` is properly defaulted in the benchmark model CLI but not python API function call. this can cause errors with operating over the `None` default value in that scenario.

This PR adds a fix to provide a default when `None` is specified


dbarbuzzi (dismissed):
Worked on my end – the example code being tested ran successfully and matched the number of cores used with the equivalent CLI command.

rahul-tuli (approved):

bfineran:
approved prior to quality fixes - merging

PR #1035, opened 2023-05-23
Refactor engine.py to pull out some common functionality

mgoin:


src/deepsparse/engine.py
```diff
@@ -152,7 +153,92 @@ def _validate_scheduler(scheduler: Union[None, str, Scheduler]) -> Scheduler:
...
+    def value(self):
+        return self._deepsparse_context
+
+    @property
+    def num_cores(self):
+        return self._num_cores
+
+    @property
+    def num_streams(self):
+        return self._deepsparse_context.num_streams()
```
bnellnm, line 193: why use the deepsparse_context for one of these methods but not the other?

bnellnm (commented):
Looks ok to me except for the stuff in KVCacheEngine

src/deepsparse/engine.py
mgoin: The context decides on the final number of streams so we need to ask it. We should store this value though

dbogunowicz (approved):
lgtm

PR #1047, opened 2023-05-29
[OPT][Pipeline] Assume that all inputs have zeroth dimension `batch_size`

dbogunowicz:
In the process of figuring out the final design of the pipeline, we are defaulting to the assumption, that every input/output from the engine has a zeroth dimension `batch_size`.

This PR implements this functionality. It was tested using a model generated by the branch specified here:
https://github.com/neuralmagic/sparseml/pull/1589
and outputs the expected generated sequence:
```bash
The president of the United States is the head of the executive branch of the government. The president is the head of the executive branch of the...
```

For now leaving this as a separate PR, so that the backward functionality with the models generated before the seminar is not violated. Once the pipeline is ready to land to `main`, this PR can be automatically pulled into the main feature branch.


bfineran (commented):
looks like this needs rease

PR #1061, opened 2023-06-08
update max datasets version

bfineran:
to be verified by @anmarques

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,7 +147,7 @@ _yolov8_integration_deps = _yolo_integration_deps + ["ultralytics==8.0.30"]
 _transformers_integration_deps = [
     f"{'nm-transformers' if is_release else 'nm-transformers-nightly'}"
     f"~={version_base}",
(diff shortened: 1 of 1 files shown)


mgoin (approved):

dbogunowicz (approved):

anmarques (approved):
Tested datasets==2.11.0 with datasets.transformers.eval_downstream using the following datasets and models:
- SQuAD / oBERT-base
- SQuADv2 / oBERTa-base
- MNLI  / oBERT-base
- QQP / oBERT-base
- goemotions / oBERT-base
- SST-2 / oBERT-base
- CoNLL-2003 / oBERT-base
- IMDB / oBERT-base

All tests succeeded (results in correct accuracy), with the exception of CoNLL-2003. For CoNLL-2003 eval crashed but it seems to be independent of the datasets version, since the same error happens with datasets==1.18.4.

PR #1067, opened 2023-06-13
[WIP] Engine KV Cache Integration

bfineran:
Draft integration of KVCache object into the text generation pipeline. Note that the `KVCache` object will be moved into the engine binaries and only a placeholder is provided

only updates the autoregressive path


src/deepsparse/kv_cache.py
```diff
@@ -0,0 +1,37 @@
...
+class KVCache:
+    def __init__(
+        self,
+        cache: Dict[str, numpy.ndarray],
+        prompt_length: int,
+        frozen_positions: List[int],
+    ):
+        pass
+
+    def reset(self, prompt_length: int, frozen_positions: List[int]):
```
bfineran, line 33: @markurtz @SageMoore curious on the relationship between init and reset here - it seems like some of the information in init is specific to the run of a single query which seems to be better suited for (and is repeated in) reset.  any reason to have in init?

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -109,12 +110,14 @@ def __init__(
...
         self.force_max_tokens = force_max_tokens
 
+        self.engine_args["cache_bools"] = self.cache_bools
         self.engine = Pipeline.create_engine(
             self.onnx_file_path,
             self.engine_type,
             self.engine_args,
             self.context,
         )
+        self.kv_cache = KVCache({}, 0, [0])  # TODO: populate with correct values
```
bfineran, line 120: @markurtz @SageMoore what do we want for the initial kv cache? zeros of the correct shape? do we need param names?

src/deepsparse/engine.py
```diff
@@ -261,11 +264,13 @@ class Engine(BaseEngine):
...
+        will try to distribute them evenly across as few sockets as possible
     :param num_streams: The max number of requests the model can handle
-        concurrently.
-    :param scheduler: The kind of scheduler to execute with. Pass None for the default.
-    :param input_shapes: The list of shapes to set the inputs to. Pass None to use model as-is.
+        concurrently
+    :param scheduler: The kind of scheduler to execute with. Pass None for the default
+    :param input_shapes: The list of shapes to set the inputs to. Pass None to use model as-is
+    :param kv_cache_bools: List of boolean values for each model input in-order of whether
+        they represent the input of  KV-cache
```
dbogunowicz, line 273:
```suggestion
        they represent the input of KV-cache
```

src/deepsparse/engine.py
```diff
@@ -276,6 +281,7 @@ def __init__(
         num_streams: int = None,
         scheduler: Scheduler = None,
         input_shapes: List[List[int]] = None,
+        kv_cache_bools: List[bool] = None,
```
dbogunowicz, line 284:
```suggestion
        kv_cache_bools: Optionall[List[bool]] = None,
```

src/deepsparse/kv_cache.py
```diff
@@ -0,0 +1,37 @@
...
+
+import numpy
+
+
+class KVCache:
+    def __init__(
+        self,
+        cache: Dict[str, numpy.ndarray],
+        prompt_length: int,
+        frozen_positions: List[int],
```
dbogunowicz, line 29: What are frozen positions?

src/deepsparse/kv_cache.py
```diff
@@ -0,0 +1,37 @@
...
+from typing import Dict, List
+
+import numpy
+
+
+class KVCache:
+    def __init__(
+        self,
+        cache: Dict[str, numpy.ndarray],
+        prompt_length: int,
```
dbogunowicz, line 28: Why do we need to know the prompt_length in advance? To allocate the buffers for the kv cache ahead of time?

src/deepsparse/kv_cache.py
```diff
@@ -0,0 +1,37 @@
...
+
+from typing import Dict, List
+
+import numpy
+
+
+class KVCache:
+    def __init__(
+        self,
+        cache: Dict[str, numpy.ndarray],
```
dbogunowicz, line 27: Is this the initial KV cache state?

src/deepsparse/kv_cache.py
dbogunowicz: Oh i see this is not as much as prompt_length, rather the "tokens seen so far" right?

src/deepsparse/transformers/pipelines/text_generation.py
dbogunowicz: Very good question. Also, how will initial kv cache differ whether we are initially running multitoken inference or directly start with the singletoken inference?

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -329,7 +334,9 @@ def autoregressive_inference(
         engine_inputs.update(kv_cache)
         engine_inputs = [engine_inputs[name] for name in self.engine.input_names]
 
-        new_logits, *cache_values = self.engine(engine_inputs)
+        new_logits, *cache_values = self.engine(engine_inputs, kv_cache=kv_cache)
+        self.kv_cache.shift_last(1)
+        # TODO: do we still need to update the cache below since engine handles?
```
dbogunowicz, line 339: If 'shift last' will work correctly also in the case of onnx runtime inference, it would be great! we could get rid of assemble_kv_cache  I think.

dbogunowicz (commented):
I like the interface so far; however there are still many open questions. When are we planning to finalize the engine kv cache integration?

src/deepsparse/kv_cache.py
SageMoore: This data structure is passed into the engine at runtime. Prompt_length is necessary here so that the engine knows how many tokens were used to generate the KV cache since it will be sized internally based on the context length. The prompt length will be 0 for cases where we are building up the KV cache from scratch using a single token model. I called it prompt_length because we typically describe the list of tokens passed into the multi_token model as a prompt. I'm open to suggestions W.R.T naming, though.

src/deepsparse/kv_cache.py
SageMoore: Yes. We are transitioning from bools to ints to support "freezing" numerous tokens at the beginning of a particular kv cache.

src/deepsparse/kv_cache.py
SageMoore: Init and reset are very similar. This data is required to properly execute the run of the engine. I added these arguments to the constructor to avoid.

`cache = KVcache()`
`cache.reset(prompt_length, frozen_positions)`

src/deepsparse/transformers/pipelines/text_generation.py
SageMoore: FWIW, the KVCache object ONLY works with the deepsparse engine. Onnx runtime will not be able to leverage this infrastructure.

src/deepsparse/kv_cache.py
SageMoore, line 29: FWIW, I currently have this implemented as a single int. Switching to a List isn't a problem, but I'm just making sure that there is a use case for having different freezing strategies for different caches.

PR #1071, opened 2023-06-14
[Fix] Fix CLI benchmark errors

dbogunowicz:
Pending adding unittests

### DIFF
--- a/src/deepsparse/utils/onnx.py
+++ b/src/deepsparse/utils/onnx.py
@@ -239,6 +239,7 @@ def override_onnx_batch_size(
         return save_onnx_to_temp_files(model, with_external_data=not inplace)
 
 
+@contextlib.contextmanager
 def override_onnx_input_shapes(
     onnx_filepath: str,
     input_shapes: Union[List[int], List[List[int]]],
@@ -303,7 +304,7 @@ def override_onnx_input_shapes(
             "Overwriting in-place the input shapes of the model " f"at {onnx_filepath}"
         )
         onnx.save(model, onnx_filepath)
-        return onnx_filepath
+        yield onnx_filepath
     else:
         _LOGGER.info(
             f"Saving the input shapes of the model at {onnx_filepath} "


src/deepsparse/utils/onnx.py
```diff
@@ -303,7 +304,7 @@ def override_onnx_input_shapes(
             "Overwriting in-place the input shapes of the model " f"at {onnx_filepath}"
         )
         onnx.save(model, onnx_filepath)
-        return onnx_filepath
+        yield onnx_filepath
```
dbarbuzzi, line 307: Should the `return` on line 307/308 also be changed to `yield`? Given that nothing happens after the `yield`, is the change necessary?

dbarbuzzi: You’re right. Given that, though, the `return` in the `else` clause should definitely be changed to `yield`.

dbogunowicz: Oh, actually this returns a function that in itself yields :)

dbogunowicz: You may be right @dbarbuzzi. Just realized it writing the unit tests.

src/deepsparse/utils/onnx.py
```diff
@@ -303,13 +308,16 @@ def override_onnx_input_shapes(
             "Overwriting in-place the input shapes of the model " f"at {onnx_filepath}"
```
rahul-tuli, line 308: nit
```suggestion
            "Overwriting in-place the input shapes of the model at {onnx_filepath}"
```

mgoin (dismissed):

src/deepsparse/utils/onnx.py
```diff
@@ -60,15 +60,16 @@ def save_onnx_to_temp_files(model: onnx.ModelProto, with_external_data=False) ->
...
+        external_data = tempfile.NamedTemporaryFile(
+            suffix=".data", delete=False, mode="w"
         )
-        has_external_data = save_onnx(model, shaped_model.name, external_data)
-        _LOGGER.info(f"Saving external data to temporary directory: {external_data}")
+        _LOGGER.info(f"Saving external data to temporary file: {external_data.name}")
+        has_external_data = save_onnx(model, shaped_model.name, external_data.name)
     else:
         has_external_data = save_onnx(model, shaped_model.name)
     try:
```
rahul-tuli, line 75: Will the next line ever raise an error? if not I presume the try .. finally block can be removed; specially cause yield is used

tests/deepsparse/utils/onnx.py
```diff
@@ -0,0 +1,71 @@
...
+    with override_onnx_input_shapes(
+        onnx_file_path, input_shapes, inplace=inplace
+    ) as modified_model_path:
+        # Load the modified ONNX model
+        modified_model = onnx.load(modified_model_path)
+        new_input_shapes = [
+            dim.dim_value
+            for dim in modified_model.graph.input[0].type.tensor_type.shape.dim
+        ]
+        assert new_input_shapes == input_shapes
```
rahul-tuli, line 71: should we add a check here after the `with` statement that no new files exist? we could do an list(folder.iterdir()) on parent folder and check that the length is same before and after?

tests/deepsparse/utils/onnx.py
```diff
@@ -0,0 +1,71 @@
...
+
+@pytest.mark.parametrize(
+    "test_model, input_shapes",
+    [
+        (
+            "zoo:cv/classification/mobilenet_v1-1.0/pytorch/sparseml/imagenet/base-none",  # noqa: E501
+            [10, 224, 224, 3],
+        )
+    ],
+    scope="function",
```
rahul-tuli, line 56: Should we scope this to the function or session maybe? to reduce downloads?

tests/deepsparse/utils/onnx.py
rahul-tuli: Or will that check be more involved

dbogunowicz: This happens automatically thanks to conftest!

src/deepsparse/utils/onnx.py
dbogunowicz, line 308:
```suggestion
            f"Overwriting in-place the input shapes of the model at {onnx_filepath}"
```

tests/deepsparse/utils/onnx.py
dbogunowicz: good idea

src/deepsparse/utils/onnx.py
dbogunowicz: This was not originally written by me, but the way I read it is:
- yield `shaped_model.name` and keep it "alive" as long as the context manager is open
- once the context manager closes, run everything in the `finally` statement (i.e do cleanup)

tests/deepsparse/utils/onnx.py
rahul-tuli: Maybe I described it wrong, I would like a test for `save_onnx_to_temp_files ` since it is used by override_onnx_input_shapes I thought this could be a candidate for the same, but I guess a separate test would be cleaner. will leave that upto you!

bfineran (approved):

PR #1080, opened 2023-06-20
Update ONNX utilities to ingest filepath or loaded model

KSGulin:
This PR updates a number of DeepSparse utilities specific to ONNX model handling to be able to ingest a model file path (nominal behavior) or a ModelProto

### DIFF
--- a/src/deepsparse/utils/onnx.py
+++ b/src/deepsparse/utils/onnx.py
@@ -131,7 +131,7 @@ def model_to_path(model: Union[str, Model, File]) -> str:
 def get_external_inputs(onnx_model: Union[str, ModelProto]) -> List:
     """
     Gather external inputs of ONNX model
-    :param onnx_filepath: File path to ONNX model or ONNX model object
+    :param onnx_model: File path to ONNX model or ONNX model object
     :return: List of input objects
     """
     model = onnx_model if isinstance(onnx_model, ModelProto) else onnx.load(onnx_model)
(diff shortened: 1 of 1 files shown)


dbogunowicz (approved):

src/deepsparse/utils/onnx.py
```diff
@@ -128,13 +129,13 @@ def model_to_path(model: Union[str, Model, File]) -> str:
...
-def get_external_inputs(onnx_filepath: str) -> List:
+def get_external_inputs(onnx_model: Union[str, ModelProto]) -> List:
     """
     Gather external inputs of ONNX model
-    :param onnx_filepath: File path to ONNX model
+    :param onnx_model: File path to ONNX model or ONNX model object
     :return: List of input objects
     """
-    model = onnx.load(onnx_filepath)
+    model = onnx_model if isinstance(onnx_model, ModelProto) else onnx.load(onnx_model)
```
dbogunowicz, line 138:
```suggestion
    model = onnx_model if isinstance(onnx_model, ModelProto) else onnx.load(onnx_model, load_external_data=False)
```
Whenever we just take a look at inputs and derive something from them, without actually using the model's weights, let's load without external data to avoid loading big data to memory. This applies to all the functions edited in this PR.

dbogunowicz (changes requested):
Please see my suggestion.

src/deepsparse/utils/onnx.py
KSGulin: Good call. Applied 🙂

dbogunowicz (approved):

bfineran (approved):

PR #1101, opened 2023-06-29
Allow check_transformers_install to warn and pass if NM fork isn't used

mgoin:


src/deepsparse/transformers/__init__.py
```diff
@@ -41,9 +41,9 @@
 def _check_transformers_install():
     import transformers as _transformers
 
-    if not _transformers.NM_INTEGRATED:
+    if not hasattr(_transformers, "NM_INTEGRATED") or not _transformers.NM_INTEGRATED:
```
bfineran, line 44: I guess `not _transformers.NM_INTEGRATED` is a bad check since base won't have this at all.
```suggestion
    if not getattr(_transformers, "NM_INTEGRATED", False):
```

mgoin: Nice call, much more clear

PR #1103, opened 2023-06-30
optionally add logits to model output

mwitiderrick:
```
cv_pipeline = Pipeline.create(
  task='image_classification',
  model_path='zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned95-none',  # Path to checkpoint or SparseZoo stub
  top_k=3,
  return_logits = True,
)
input_image = "cheetah.jpeg" # path to input image
inference = cv_pipeline(images=input_image)
inference

# ImageClassificationOutput(labels=None, scores=None, logits=array([[ 1.8277559 ,  4.486894  ,  2.4551017 ,  4.777242  ,  # 2.744733  ,.... 8.45131   ]],
 #  dtype=float32))
```

### DIFF
--- a/src/deepsparse/image_classification/pipelines.py
+++ b/src/deepsparse/image_classification/pipelines.py
@@ -75,6 +75,7 @@ class ImageClassificationPipeline(Pipeline):
         *,
         class_names: Union[None, str, Dict[str, str]] = None,
         top_k: int = 1,
+        return_logits: bool = False,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -88,6 +89,7 @@ class ImageClassificationPipeline(Pipeline):
 
         self._image_size = self._infer_image_size()
         self.top_k = top_k
+        self.return_logits = return_logits
 
         # torchvision transforms for raw inputs
         non_rand_resize_scale = 256.0 / 224.0  # standard used
@@ -112,7 +114,7 @@ class ImageClassificationPipeline(Pipeline):
             mapping class ids to class labels
         """
         return self._class_names
-
+    
     @property
     def input_schema(self) -> Type[ImageClassificationInput]:
         """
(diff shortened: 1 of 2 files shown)


src/deepsparse/image_classification/schemas.py
```diff
@@ -38,10 +38,15 @@ class ImageClassificationOutput(BaseModel):
...
+    class Config:
+        arbitrary_types_allowed = True
 
     labels: List[Union[int, str, List[int], List[str]]] = Field(
         description="List of labels, one for each prediction"
     )
     scores: List[Union[float, List[float]]] = Field(
         description="List of scores, one for each prediction"
     )
+    logits: Optional[ndarray] = Field(description="Raw model outputs")
```
rahul-tuli, line 52: Could we test this logic with deepsparse server? I remember we had issues earlier when keeping numpy arrays in I/O schemas

src/deepsparse/image_classification/pipelines.py
```diff
@@ -112,7 +114,7 @@ def class_names(self) -> Optional[Dict[str, str]]:
             mapping class ids to class labels
         """
         return self._class_names
-
+
```
dbogunowicz, line 117: Nit: uneccessary

src/deepsparse/image_classification/schemas.py
dbogunowicz: Yes, I second that!

dbogunowicz: also good idea to set the default to None here

src/deepsparse/image_classification/schemas.py
```diff
@@ -38,10 +38,15 @@ class ImageClassificationOutput(BaseModel):
     """
     Output model for image classification
     """
+    ndarray: numpy.ndarray = numpy.ndarray
```
dbogunowicz, line 41: I'd say this adds unnecessary clutter. Let's just use for typing `numpy.ndarray`

src/deepsparse/image_classification/pipelines.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         *,
         class_names: Union[None, str, Dict[str, str]] = None,
         top_k: int = 1,
+        return_logits: bool = False,
```
bfineran, line 78: nit: update docstring as well

mwitiderrick:
I am going to close this PR because I was able to get the logits through the Engine option.  Unless people think it's important to output the logits in the pipelines

PR #1106, opened 2023-07-05
[KV Cache] Implementation of `KVCacheSessionStorage`

dbogunowicz:
Testing pending...


src/deepsparse/transformers/engines/nl_decoder_engine.py
```diff
@@ -23,16 +23,46 @@
...
 
 _LOGGER = logging.getLogger(__name__)
 
-__all__ = ["NLDecoderEngine"]
+__all__ = ["NLDecoderEngine", "synchronise_engines_cache"]
 
 _CACHE_INPUT_NAME = "past_key_values"
 
 
+def synchronise_engines_cache(engines: Set["NLDecoderEngine"]):  # noqa F821
```
bfineran, line 37: What is the desired use case for this? Why do we want to synchronize the *entire* caches between pipelines rather than just the individual cache of the session that should be transferred

src/deepsparse/transformers/engines/nl_decoder_engine.py
```diff
@@ -158,25 +180,18 @@ def __call__(
...
-            state=state,
-            sequence_length=self.sequence_length,
-            freeze_first_position=self._freeze_first_position,
+        _LOGGER.debug(
+            f"Transferring cache storage {cache_storage} to the engine {self.engine}"
         )
-        # maybe set as a property for more control?
-        self._num_tokens = engine._num_tokens
+
+        self.kv_cache_storage = cache_storage
```
bfineran, line 194: how to we handle the fact that different engines will expect different cache shapes? i.e. we can't just set a pipeline with length 64 cache to have the same cache as a single token pipeline w/ a 127 length cache

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -244,12 +268,19 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
 
-        if inputs.session_id is not None:
-            # if session_id is provided, we need to set it in engines
-            self.engine.session_id = inputs.session_id
-            self.multitoken_engine.session_id = inputs.session_id
-
         postprocessing_kwargs = dict(return_logits=inputs.return_logits)
+
+        if inputs.session_ids is not None:
+            session_ids = (
```
bfineran, line 274: why don't we just add this to the validator so we don't need to worry about it in the pipeline?

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -244,12 +268,19 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
+                [inputs.session_ids]
+                if isinstance(inputs.session_ids, str)
+                else inputs.session_ids
+            )
+            # TODO: This is ugly, we should find a better way to pass
+            # session_id to the engine, maybe use the refactored pipeline
+            engine_input.append(numpy.array(session_ids))
```
bfineran, line 281: agreed, limitation of the "EnginePipeline" assumption - had thought we had the ability to do kwargs here for some reason...

bfineran (changes requested):
see comments regarding storage sharing, want to make sure we're on the same page on usage...

src/deepsparse/transformers/engines/nl_decoder_engine.py
dbogunowicz: Agree, changes introduced

src/deepsparse/transformers/engines/nl_decoder_engine.py
dbogunowicz: This is a valid remark, but let's push this requirement until we have the causal mask support. There is no way for us to properly test it for now.

src/deepsparse/transformers/pipelines/text_generation.py
dbogunowicz: Good idea

PR #1110, opened 2023-07-09
Mlserver example

robertgshaw2-redhat:


examples/mlserver/client.py
```diff
@@ -0,0 +1,27 @@
+import requests, threading
+
+NUM_THREADS = 2
+URL = "http://localhost:8080/v2/models/text-classification-model/infer"
+sentences = ["I hate using GPUs for inference", "I love using DeepSparse on CPUs"] * 100
```
mgoin, line 5: Should `* 100` be `* NUM_THREADS` if we are taking only `sentences[:NUM_THREADS]` elements?

examples/mlserver/README.md
```diff
@@ -0,0 +1,75 @@
+# **Step 1: Installation**
+
```
InquestGeronimo, line 2: best to add an intro paragraph to give users a heads up of what this example does.

examples/mlserver/client.py
```diff
@@ -0,0 +1,27 @@
...
+    resp = requests.post(URL, json=inference_request).json()
+    for output in resp["outputs"]:
+        print(output["data"])
+
+
+threads = [threading.Thread(target=tfunc, args=(sentence,)) for sentence in sentences[:NUM_THREADS]]
+for thread in threads:
+    thread.start()
+for thread in threads:
+    thread.join()
```
bfineran, line 27: it looks like this creates `NUM_THREADS` threads to make the request, is that intended?
Might make more sense to create `len(sentences)` threads and execute `NUM_THREADS` at a time.

You can do this out of the box with `ThreadPoolExecutor` with something like:

```suggestion
from concurrent.futures.thread import ThreadPoolExecutor

threadpool = ThreadPoolExecutor(max_workers=NUM_THREADS)

results = threadpool.map(tfunc, sentences)
```

examples/mlserver/client.py
```diff
@@ -0,0 +1,27 @@
+import requests, threading
+
+NUM_THREADS = 2
+URL = "http://localhost:8080/v2/models/text-classification-model/infer"
+sentences = ["I hate using GPUs for inference", "I love using DeepSparse on CPUs"] * 100
+
+def tfunc(text):
```
bfineran, line 7: would rename to something more descriptive like `inference_request`

examples/mlserver/client.py
```diff
@@ -0,0 +1,27 @@
...
+                "name": "sequences",
+                "shape": [1],
+                "datatype": "BYTES",
+                "data": [text],
+            },
+        ]
+    }   
+    resp = requests.post(URL, json=inference_request).json()
+    for output in resp["outputs"]:
+        print(output["data"])
```
bfineran, line 20: executing a list printout while multithreaded may cause a race condition, any reason to not return the value and print in sequence at the end? (ie consider thread 1 and thread 2 happen to execute *exactly* at the same time, they will print their lines at the same time and might not tell which is which)

examples/mlserver/client.py
bfineran: @rsnm2 see suggestion below

examples/mlserver/client.py
```diff
@@ -0,0 +1,27 @@
+import requests, threading
```
bfineran, line 1: would suggest a few in line comments for self-documentation

examples/mlserver/models/text-classification-model/models.py
```diff
@@ -0,0 +1,19 @@
...
+from typing import List
+from deepsparse import Pipeline
+
+class DeepSparseRuntime(MLModel):
+    async def load(self) -> bool:
+        self._pipeline = Pipeline.create(
+            task = self._settings.parameters.task,
+            model_path = self._settings.parameters.model_path,
+            batch_size = self._settings.parameters.batch_size,
+            sequence_length = self._settings.parameters.sequence_length,
```
bfineran, line 12: is there a place for generic kwargs in the settings? Would be cool if we could use that instead to dump extra pipeline args so we can get full generic pipeline support out of the box

examples/mlserver/models/text-classification-model/models.py
```diff
@@ -0,0 +1,19 @@
+from mlserver import MLModel
```
bfineran, line 1: this is great, love that it works out of the box - let's throw in the serving command as a comment just for convenience

jeanniefinks:

PR #1115, opened 2023-07-11
AWS Serverless update

InquestGeronimo:
updating deepsparse version/lambda image to avoid bugs in lambda


mgoin (approved):
Nice cleanup! Glad the topo file isn't needed anymore. Is there any num core setting in the server config?

PR #1121, opened 2023-07-13
[TextGeneration][Timer] text gen specific timings + improved timing tooling

bfineran:
adds:
* bugfix for `TimerManager` workflow for `engine_forward` subtimings
    * since `engine_forward` is always called in a child thread for batch splitting, it will not have access to the `current` timer contextvar
    * see changes in `timer.py` `new_timer_context` for this
* adds `timer.time` helper contextmanager; see docstring for usage
* adds subtimings for prefill, generation and the individual autoregressive passes for each in the text generation pipeline
    * see example in test plan for what these timings may look like

**test_plan:**
Manually for now - once we have more text gen pipeline tests, we can add assertions that the proper stages are generated in the timer manager:

```python
# running text gen / opt pipeline
opt = Pipeline.create(task="opt", model_path=MODEL_PATH, engine_type="onnxruntime", max_generated_tokens=128)
output = opt(sequences="Who is the president of the United States?", return_logits=True)

print(opt.timer_manger)
print(opt.timer_manager.stages)
```

Manager and stages:
```python
TimerManager({'engine_token_generation': 4.507111581042409, 'engine_prompt_prefill_single': 0.03916543936356902, 'engine_prompt_prefill': 0.3927434356883168, 'pre_process': 0.001995994709432125, 'engine_token_generation_single': 0.035451228459050334, 'engine_forward': 4.909945313818753, 'total_inference': 4.912277169525623, 'post_process': 0.00030374620109796524})
```

```python
['engine_token_generation', 'engine_prompt_prefill_single', 'engine_prompt_prefill', 'pre_process', 'engine_token_generation_single', 'engine_forward', 'total_inference', 'post_process']
```

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -35,10 +35,10 @@ __all__ = ["TextGenerationPipeline"]
 
 @dataclass(frozen=True)
 class _TextGenerationTimings:
-    PROMPT_PREFILL = "engine_prompt_prefill"
-    PROMPT_PREFILL_SINGLE = "engine_prompt_prefill_single"
-    TOKEN_GENERATION = "engine_token_generation"
-    TOKEN_GENERATION_SINGLE = "engine_token_generation_single"
+    PROMPT_PREFILL: str = "engine_prompt_prefill"
+    PROMPT_PREFILL_SINGLE: str = "engine_prompt_prefill_single"
+    TOKEN_GENERATION: str = "engine_token_generation"
+    TOKEN_GENERATION_SINGLE: str = "engine_token_generation_single"
 
 
 class TextGenerationInput(BaseModel):
@@ -344,17 +344,19 @@ class TextGenerationPipeline(TransformersPipeline):
             generated_tokens = [tokens[-1]]
             generated_logits = prompt_logits
 
-            timer.start(_TextGenerationTimings.TOKEN_GENERATION)
-            while len(generated_tokens) < max_tokens:
-                with timer.time(_TextGenerationTimings.TOKEN_GENERATION_SINGLE):
-                    token, logits = self.autoregressive_inference(tokens)
-                tokens.append(token)
-                generated_tokens.append(token)
-                generated_logits.append(logits)
+            with timer.time(_TextGenerationTimings.TOKEN_GENERATION):
+                while len(generated_tokens) < max_tokens:
+                    with timer.time(_TextGenerationTimings.TOKEN_GENERATION_SINGLE):
+                        token, logits = self.autoregressive_inference(tokens)
+                    tokens.append(token)
+                    generated_tokens.append(token)
+                    generated_logits.append(logits)
 
-                if token == self.tokenizer.eos_token_id and not self.force_max_tokens:
-                    break
-            timer.stop(_TextGenerationTimings.TOKEN_GENERATION)
+                    if (
+                        token == self.tokenizer.eos_token_id
+                        and not self.force_max_tokens
+                    ):
+                        break
 
         return numpy.array([generated_tokens]), numpy.concatenate(
             generated_logits, axis=1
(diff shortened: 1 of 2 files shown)


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -30,6 +30,12 @@
 __all__ = ["TextGenerationPipeline"]
 
 
+PROMPT_PREFILL = "engine_prompt_prefill"
```
dbogunowicz, line 33: I do not like those variables "floating around" in the script. This Python file is already fairly complex and does not need more clutter. Could we possibly introduce more order to the definitions of the timing stage names? 
A simple idea would be to wrap those four variables inside a data class (like here https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/utils/timer.py#L29)  and import from a helper script?

dbogunowicz (approved):

dbogunowicz (approved):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -304,35 +313,41 @@ def engine_forward(
...
+            while len(generated_tokens) < max_tokens:
+                with timer.time(_TextGenerationTimings.TOKEN_GENERATION_SINGLE):
+                    token, logits = self.autoregressive_inference(tokens)
+                tokens.append(token)
+                generated_tokens.append(token)
+                generated_logits.append(logits)
+
+                if token == self.tokenizer.eos_token_id and not self.force_max_tokens:
+                    break
+            timer.stop(_TextGenerationTimings.TOKEN_GENERATION)
```
rahul-tuli, line 350: Why not use the `timer.time` context manager for `TOKEN_GENERATION`, was this done intentionally to maybe reduce indentation level?
```python
            with timer.time(_TextGenerationTimings.TOKEN_GENERATION):
```

src/deepsparse/utils/timer.py
```diff
@@ -114,6 +114,26 @@ def has_stage(self, stage: str) -> bool:
...
+        ```
+
+        :param stage: the name of the stage to time
+        """
+        self.start(stage)
+
+        try:
+            yield
+        finally:
+            self.stop(stage)
```
rahul-tuli, line 135: The implementation looks good but I don't believe we need finally here;
```suggestion
        self.start(stage)
        yield
        self.stop(stage)
```

Using finally with yield can sometimes cause unforeseen errors, (like executing the finally clause if the generator reference is not stored) more info here: https://stackoverflow.com/a/56063376/21911160

Though using the `@contextmanager` decorator protects us from executing the finally clause(as it stores a reference to the generator) but I think the proposed version is safer and reads better (also reduces an indentation level)

src/deepsparse/utils/timer.py
```diff
@@ -341,4 +367,5 @@ def new_timer_context(self) -> StagedTimer:
         try:
             yield timer
         finally:
-            timer.stop(InferenceStages.TOTAL_INFERENCE)
+            if total_inference:
+                timer.stop(InferenceStages.TOTAL_INFERENCE)
```
rahul-tuli, line 371: Same comment as above, can remove the try.. finally clause

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -30,6 +31,14 @@
 __all__ = ["TextGenerationPipeline"]
 
 
+@dataclass(frozen=True)
+class _TextGenerationTimings:
+    PROMPT_PREFILL = "engine_prompt_prefill"
```
rahul-tuli, line 36: nit: could we add types to the dataclass here?

rahul-tuli (commented):
LGTM  pending response to comments!

bfineran:
review suggestions LGTM

rahul-tuli (approved):

dbogunowicz (approved):

PR #1127, opened 2023-07-19
[Text Generation] Causal Mask Support

dbogunowicz:
Allows the user to set the argument `prompt_processing_sequence_length` of the `TextGeneration` pipeline to a value different from `sequence_length`, effectively enabling running a `multitoken_engine` in a scenario, when we feed it `input_ids` of any length, and robustly provide kv cache support. In other words, it enables prefilling the cache using subsequences of different lengths.

### Manual Testing

Complementary feature (and PR) from Sparseml: https://github.com/neuralmagic/sparseml/pull/1676

```python

def _test_pipeline(engine_type):
    opt = Pipeline.create(task="opt",
                          model_path="/home/ubuntu/damian/sparseml/deployment",
                          engine_type=engine_type,
                          prompt_processing_sequence_length=64,
                          use_deepsparse_cache = False,
                          max_generated_tokens=32)
    print('----------')
    prompt = "def hello_world():" # the prompt is short, will not be processed by self.multitoken_engine
    out = opt(sequences=prompt, return_logits = True)
    print(out.sequences[0])
    print('---------')
    prompt = "def hello_world():" * 20 # the prompt is long, will be processed by self.multitoken_engine
    out = opt(sequences=prompt, return_logits=True)
    print(out.sequences[0])

_test_pipeline(engine_type ="onnxruntime")
_test_pipeline(engine_type ="deepsparse")
```

```bash
#### INFERENCE WITH ONNXRUNTIME ####
2023-07-20 05:56:04 deepsparse.transformers.engines.nl_decoder_engine INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/damian/sparseml/deployment/model.onnx
2023-07-20 05:56:10 deepsparse.utils.onnx INFO     Overwriting in-place the batch size of the model at /home/ubuntu/damian/sparseml/deployment/model.onnx
2023-07-20 05:56:18 deepsparse.transformers.engines.nl_decoder_engine INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/damian/sparseml/deployment/model.onnx
2023-07-20 05:56:24 deepsparse.utils.onnx INFO     Overwriting in-place the batch size of the model at /home/ubuntu/damian/sparseml/deployment/model.onnx
----------

    return 'Hello World!'

def hello_world_2():
    return 'Hello World!'

def hello_world_3(): # output is identical to the one from the torch model (baseline)
---------
def hello_world():def hello_world():def hello_world():def hello_world():def hello_world():def hello_world():def hello # output is identical to the one from the torch model (baseline)

#### INFERENCE WITH ONNXRUNTIME ####
# Note that we do not see a warning stating that multitoken engine must default to onnxruntime, because the deepsparse runtime cannot support it! Now both single token and multi token engines are running using deepsparse!

/home/ubuntu/damian/deepsparse/src/deepsparse/transformers/pipelines/text_generation.py:126: UserWarning: AVX512 support not detected, disabling internal management of KV cache which may affect performance. To enable full performance, deploy on an AVX512-compatible system.
  warnings.warn(
Using pad_token, but it is not set yet.
2023-07-20 05:56:37 deepsparse.transformers.engines.nl_decoder_engine INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/damian/sparseml/deployment/model.onnx
DeepSparse, Copyright 2021-present / Neuralmagic, Inc. version: 1.6.0.20230711 COMMUNITY | (34a5203e) (release) (optimized) (system=avx2, binary=avx2)
2023-07-20 05:58:58 deepsparse.transformers.engines.nl_decoder_engine INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/damian/sparseml/deployment/model.onnx
----------

    return 'Hello World!'

def hello_world_2():
    return 'Hello World!'

def hello_world_3():
---------
def hello_world():def hello_world():def hello_world():def hello_world():def hello_world():def hello_world():def hello
```

(in this scenario we run with the default `prompt_processing_sequence_length=64`, but setting it to `32` gives the same result naturally)

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -13,12 +13,14 @@
 # limitations under the License.
 
 import logging
+import warnings
 from typing import List, Optional, Tuple, Type, Union
 
 import numpy
 from pydantic import BaseModel, Field
 
 from deepsparse import Pipeline
+from deepsparse.cpu import cpu_avx512_compatible
 from deepsparse.pipeline import DEEPSPARSE_ENGINE
 from deepsparse.transformers.engines import NLDecoderEngine
 from deepsparse.transformers.pipelines import TransformersPipeline
@@ -115,9 +117,17 @@ class TextGenerationPipeline(TransformersPipeline):
         # TODO: Set this to 64 once we modify the OPT injection logic
         prompt_processing_sequence_length: int = 128,
         force_max_tokens: bool = False,
-        use_deepsparse_cache: bool = False,
+        use_deepsparse_cache: bool = True,
         **kwargs,
     ):
+        if not cpu_avx512_compatible() and kwargs["engine_type"] == DEEPSPARSE_ENGINE:
+            warnings.warn(
+                "AVX512 support not detected, disabling internal management "
+                "of KV cache which may affect performance. To enable full "
+                "performance, deploy on an AVX512-compatible system."
+            )
+            use_deepsparse_cache = False
+
         if use_deepsparse_cache:
             if kwargs["engine_type"] != DEEPSPARSE_ENGINE:
                 raise ValueError(
@@ -126,10 +136,6 @@ class TextGenerationPipeline(TransformersPipeline):
                     f"is {kwargs['engine_type']}. "
                     f"Make sure to set `engine_type` to {DEEPSPARSE_ENGINE}"
                 )
-            raise NotImplementedError(
-                "The deepsparse kv cache is not yet "
-                "supported for text generation pipelines"
-            )
 
         super().__init__(
             **kwargs, _delay_engine_initialize=True, _delay_overwriting_inputs=True
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -255,33 +255,18 @@ class TextGenerationPipeline(TransformersPipeline):
             truncation=truncate,
         )
 
-        if (
-            len(input_tokens[0]) < self.sequence_length
-            and not inputs.fixed_sequences_length
-        ):
-            # if the input is shorter than the sequence length,
-            # we need to pad it to the sequence length
-            padding = "max_length"
-            input_tokens = self.tokenizer(
-                inputs.sequences,
-                return_tensors="np",
-                max_length=self.sequence_length,
-                padding=padding,
-                truncation=truncate,
-            )
-
         attention_mask = input_tokens["attention_mask"]
 
         positions = attention_mask.cumsum(1) * attention_mask
         positions -= 1  # assert that positions start at 0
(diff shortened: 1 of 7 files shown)


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -261,15 +255,33 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
-        # let's make it optional in the future
         positions = attention_mask.cumsum(1) * attention_mask
         positions -= 1  # assert that positions start at 0
         positions_input = dict(positions=positions)
 
-        input_tokens = {**input_tokens, **positions_input}
+        causal_mask = create_causal_mask(
+            input_tokens["input_ids"], input_tokens["attention_mask"]
+        )
+        causal_mask_input = dict(causal_mask=causal_mask)
```
bfineran, line 282: nit: why do we need to wrap it into a dict, can't we just below do
```python
input_tokens = dict(**input_tokens, positions=positions, causal_mask=causal_mask)
```

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -429,12 +432,57 @@ def autoregressive_inference(
...
+        Takes a list of tokens and turns the first
+        `self.prompt_processing_sequence_length` tokens into
+        appropriate engine inputs for the multitoken engine.
+
+        :param tokens: the list of tokens to process
+        :return: a generator of engine inputs
+        """
+        # TODO: Make it yield multiple engine_inputs
+        # Once this is done, we can update the docstrings and be
+        # more verbose about what this function does
```
bfineran, line 455: can you add a few more in-line comments to this function especially the logic branches?
potentially expand the docstring as well?

Also is this function to support multitoken or causal mask? checking understanding as this is the causal mask PR (good either way)

bfineran (changes requested):

src/deepsparse/transformers/pipelines/text_generation.py
dbogunowicz: yes, what you propose is the same thing but shorter

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -261,15 +255,33 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
             truncation=truncate,
         )
 
+        if (
+            len(input_tokens[0]) < self.sequence_length
+            and not inputs.fixed_sequences_length
+        ):
+            # if the input is shorter than the sequence length,
+            # we need to pad it to the sequence length
+            padding = "max_length"
```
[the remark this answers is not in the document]
dbogunowicz: this second conditional is rubbish, I just realized that couple of days ago but forgot to update :)

PR #1128, opened 2023-07-19
[Text Generation][Fix] Update the logic from sampling from a distribution when `deterministic=False`

dbogunowicz:
The function `self.generate_token` expects to receive a single logits array of size `(vocab_size,)` and not `(batch_size, vocab_size)`.


src/deepsparse/transformers/engines/nl_decoder_engine.py
```diff
@@ -155,7 +156,7 @@ def __call__(
         else:
             logits = out[0]
 
-        token = self.generate_token(logits=logits[:, -1, :])
+        token = self.generate_token(logits=logits[0, -1, :])
```
bfineran, line 159:
```suggestion
        # select batch idx 0, batch is always 1
        token = self.generate_token(logits=logits[0, -1, :])
```

bfineran (approved):

PR #1140, opened 2023-07-23
Implement `streamer` for text-generation and add `context` arg to `Pipeline.engine_forward`

mgoin:
Allows usage of the [Transformers TextStreamer and TextIteratorStreamer classes](https://huggingface.co/docs/transformers/internal/generation_utils#transformers.TextStreamer) with DeepSparse

```python
from transformers import TextStreamer

pipe = Pipeline.create(
    task="codegen",
    model_path="zoo:nlg/text_generation/codegen_mono-350m/pytorch/huggingface/bigpython_bigquery_thepile/base-none",
    max_generated_tokens=64,
    prompt_processing_sequence_length=1,
    use_deepsparse_cache=False,
)

streamer = TextStreamer(pipe.tokenizer)

# While also returning the usual output, the streamer will also print the generated text to stdout
_ = pipe(sequences="def fib():", streamer=streamer)
```

https://github.com/neuralmagic/deepsparse/assets/3195154/ef17fd95-a54f-4709-98df-1ce6412a2911

===

This also extends beyond printing to stdout by using TextIteratorStreamer

```python
from transformers import TextIteratorStreamer
from threading import Thread

pipe = Pipeline.create(
    task="codegen",
    model_path="zoo:nlg/text_generation/codegen_mono-350m/pytorch/huggingface/bigpython_bigquery_thepile/base-none",
    max_generated_tokens=64,
    prompt_processing_sequence_length=1,
    use_deepsparse_cache=False,
)

streamer = TextIteratorStreamer(pipe.tokenizer)

generation_kwargs = dict(sequences="def fib():", streamer=streamer)
thread = Thread(target=pipe, kwargs=generation_kwargs)
thread.start()
generated_text = ""
for new_text in streamer:
    generated_text += new_text
print(generated_text)

"""
def fib():
    a, b = 0, 1
    while True:
        yield a
        a, b = b, a + b

def fib2(n):
    a, b = 0, 1
    while a < n:
        yield a
        a, b
"""
```

### DIFF
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -21,6 +21,7 @@ from abc import ABC, abstractmethod
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from functools import partial
 
 import numpy
 from pydantic import BaseModel, Field
@@ -247,7 +248,12 @@ class Pipeline(BasePipeline):
             )
 
             # submit split batches to engine threadpool
-            batch_outputs = list(self.executor.map(self.engine_forward, batches))
+            engine_forward_with_context = partial(
+                self.engine_forward, context=postprocess_kwargs
+            )
+            batch_outputs = list(
+                self.executor.map(engine_forward_with_context, batches)
+            )
 
             # join together the batches of size `self._batch_size`
             engine_outputs = self.join_engine_outputs(batch_outputs, orig_batch_size)
@@ -486,7 +492,9 @@ class Pipeline(BasePipeline):
         """
         return split_engine_inputs(items, batch_size)
 
-    def engine_forward(self, engine_inputs: List[numpy.ndarray]) -> List[numpy.ndarray]:
+    def engine_forward(
+        self, engine_inputs: List[numpy.ndarray], context: Dict = {}
+    ) -> List[numpy.ndarray]:
         """
         :param engine_inputs: list of numpy inputs to Pipeline engine forward
             pass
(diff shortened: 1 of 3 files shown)


dbogunowicz (dismissed):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -282,6 +294,8 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
             self.engine.session_id = inputs.session_id
             self.multitoken_engine.session_id = inputs.session_id
 
+        self.streamer = inputs.streamer
```
bfineran, line 297: why are we assigning state to the model based on an input? would prefer to keep in the context of a single inference as this may cause issues with multi stream and other scenarios

mgoin: @bfineran I couldn't find a way to pass kwargs into the map function used by the threadpool (as I tried adding the streamer to `postprocess_kwargs` and providing `postprocess_kwargs` to `engine_forward()`
```python
            batch_outputs = list(self.executor.map(self.engine_forward, batches))
```

src/deepsparse/pipeline.py
```diff
@@ -247,7 +248,12 @@ def __call__(self, *args, **kwargs) -> BaseModel:
             )
 
             # submit split batches to engine threadpool
-            batch_outputs = list(self.executor.map(self.engine_forward, batches))
+            engine_forward_with_context = partial(
+                self.engine_forward, context=postprocess_kwargs
```
bfineran, line 252: thinking we should rename `postprocess_kwargs` to `context` here since it's used for more than postprocess

bfineran (approved):
LGTM let's also update the PR title to include that context is added to engine forward - in terms of what's added there I think it looks good as a first step, we might want to think through mechanism/rules for updating the context dictionary in engine forward, if we want to allow that at all

src/deepsparse/pipeline.py
mgoin: cool i agree, addressed this in the latest commit!

bfineran (approved):

PR #1162, opened 2023-08-02
[Text Generation][Tests] Text Generation Pipeline

dbogunowicz:
Adds basic testing to harden the most brittle parts of the pipeline.


tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -12,33 +12,214 @@
...
+
+
+@pytest.mark.parametrize(
+    # TODO: Change to stubs
+    "model_path, model_name, uses_bos_token",
+    [
+        ("/home/ubuntu/damian/sparseml/deployment_opt", "facebook/opt-350m", True),
+        (
+            "/home/ubuntu/damian/sparseml/deployment_codegen",
+            "salesforce/codegen-350m-multi",
```
bfineran, line 59: can't land until we update here

bfineran (changes requested):
LGTM overall, 
1. let's move this away from the core pytest pathway as is right now since it will be both very long running and have a larger memory req
2. after landing, let's schedule a follow up to talk about how we can generalize and rearchitecture

dbogunowicz:
Will merge this and introduces the appropriate changes on the feature branch level.

PR #1169, opened 2023-08-04
Generalize disabling batch size across engine interfaces

mgoin:
This PR proposes changes to allow for the disabling of batch size checks, using batch size during random input generation, and disabling the engine's default batch size override behavior so the user doesn't need to define `NM_BATCH_SIZE_OVERRIDE` themselves

If a batch size of `None` or less than 1 is provided, a warning is logged and NM_DISABLE_BATCH_OVERRIDE is set. This provides a mechanism to disable the batch size override throughout the engine.

---
  - https://app.asana.com/0/0/1205173540568863


bfineran (commented):
code LGTM - can you add a PR description with motivation + test scope?

PR #1173, opened 2023-08-08
[Transformers] Add an argument `trust_remote_code` on transformers pipeline initialization.

dbogunowicz:
Give a user an option to enable running remote code through pipeline argument e.g.

```
Pipeline.create(task="...", trust_remote_code = True)
```

By default `trust_remote_code=False`.


mgoin (approved):
Thank you!
Side note: I think there is a problem with discoverability for arguments that are inherited by pipeline. Maybe this would be "solved" by autogenerated docs but I'm not sure how to make it more visible just looking at the constructors

Satrat (approved):

PR #1179, opened 2023-08-11
update mock engine to support cached bools arg

bfineran:
fixes broken GHA tests


bfineran:
GHA test passed, merging

PR #1183, opened 2023-08-15
DS eval for OPT on WikiText

natuan:
This PR enables verifying perplexity of OPT models on wikitext, matching the numbers in pytorch eval flow after one-shot prune/quantize with SparseGPT.

Qualification: tested with 1.3, 2.7b base, W8A8linear and pruned50_quantW8A8 models.

### DIFF
--- a/src/deepsparse/transformers/eval_downstream.py
+++ b/src/deepsparse/transformers/eval_downstream.py
@@ -109,6 +109,44 @@ def perplexity_eval(args, batch_size=16, dataset_name="openai_humaneval"):
     return perplexity_metrics
 
 
+def perplexity_opt_eval(args, batch_size=16, dataset_name=""):
+    if args.max_samples:
+        batch_size = min(batch_size, args.max_samples)
+
+    dataset = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")
+
+    text_generation = Pipeline.create(
+        task="text-generation",
+        model_path=args.model_path,
+        engine_type=args.engine,
+        num_cores=args.num_cores,
+        sequence_length=args.max_sequence_length,
+        prompt_processing_sequence_length=args.max_sequence_length,
+        max_generated_tokens=1,
+    )
+    perplexity_metrics = Perplexity(pipeline=text_generation, batch_size=batch_size)
+    active_engines = [
+        engine
+        for engine in [text_generation.engine, text_generation.multitoken_engine]
+        if engine
+    ]
+    print("Engine info: ")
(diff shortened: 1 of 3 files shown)


src/deepsparse/transformers/eval_downstream.py
```diff
@@ -109,6 +112,82 @@ def perplexity_eval(args, batch_size=16, dataset_name="openai_humaneval"):
     return perplexity_metrics
 
 
+def perplexity_wikitext_eval(args, batch_size=16, dataset_name=""):
```
dbogunowicz, line 115: why are we creating a duplicate of `perplexity_eval` function? i do not see a reason why we should have a separate function, that mostly is just adding boilerplate code.

src/deepsparse/transformers/metrics.py
```diff
@@ -122,22 +122,6 @@ def add_batch(self, predictions: List[str]):
 
             logits = out.logits
 
-            if not self._pipeline.has_cache:
```
dbogunowicz, line 125: why are we moving this logic from the metric to the evaluation script? this does not make sense to me from the design point of view and would only add unnecessary code.

dbogunowicz, line 125: `has_cache` has been renamed to `cache_support_enabled` btw

dbogunowicz (changes requested):
Let's clean up the code, there is no reason to write so much unnecessary lines.

src/deepsparse/transformers/metrics.py
natuan: I tried to follow the ppl and metrics design here, but got bad results. I did it this way to match how it was done from the pytorch flow, and mainly to make sure we got the right ppl numbers; plus for the sake of time. If you have concrete suggestions how to move these into the framework, please help.

src/deepsparse/transformers/eval_downstream.py
natuan: Tried to reuse this func, swapping in wikitext, got bad results.

src/deepsparse/transformers/metrics.py
abhinavnmagic: I think this pathway is important to support perplexity evaluation of non-cache models. Wouldn't we include likelihood for pad tokens if we remove this? For testing, did you test non-cache models?

natuan:
Closing because this PR will be covered [here](https://github.com/neuralmagic/deepsparse/pull/1197)

PR #1190, opened 2023-08-17
[Text Generation] Optimize the slow `update` method in the KVCacheDecoder

dbogunowicz:
As reported by @mgoin and investigated by myself, the `update` method in the KVCacheDecoder is very slow.
Profiling has shown that this is due to the repeated use of the `numpy.delete` function:

image

This discussion:
https://stackoverflow.com/questions/30399534/shift-elements-in-a-numpy-array
hints that the most elegant and quite efficient replacement for `numpy.delete` would be slicing the arrays. This is the change that this PR introduces.

Short benchmarking numbers:

image

### DIFF
--- a/src/deepsparse/transformers/utils/decoder_kv_cache.py
+++ b/src/deepsparse/transformers/utils/decoder_kv_cache.py
@@ -103,78 +103,77 @@ class DecoderKVCache:
             Corresponds to `input_ids.shape[1]`
         """
         self.total_num_processed_tokens += input_ids_len
-        total_cache_capacity = state[list(state.keys())[0]].shape[
+
+        input_state_capacity = state[list(state.keys())[0]].shape[
             self._sequence_len_axis
         ]
-        # total_capacity = num_tokens (num of non-blank tokens) +
-        # + num_padded_entries (num of blank tokens)
-        num_padded_entries = max(
-            0, total_cache_capacity - self.total_num_processed_tokens
-        )
         num_entries_to_delete = input_ids_len
 
-        if num_padded_entries:
-            """
-            Transforms input KV cache that contains blank entries.
-            It removes the rightmost blank entries from the cache.
-            Example 1:
-            (entries in the cache denote the order in which they were
-            added to the cache, zero is to denote a blank entry)
-            ```
-            state["state_name"]: (1, 1, 5, 1) = array([[[[0], [0], [1], [2], [3]]]])
-            -> num_padded_entries = 2
-            -> num_entries_to_delete = 1
-            -> num_padded_entries > num_entries_to_delete
-            # there are more blank entries than entries to delete
-            results in:
-            state["state_name"]: (1, 1, 4, 1) = array([[[[0], [1], [2], [3]]]])
-            ```
-            Example 2:
-            ```
-            state["state_name"]: (1, 1, 6, 1) = array([[[[0], [0], [0], [1], [2], [3]]]]) # noqa: E501
-            -> num_padded_entries = 3
-            -> num_entries_to_delete = 5
-            -> num_padded_entries < num_entries_to_delete
-            # there are less blank entries than entries to delete
-            results in:
-            state["state_name"]: (1, 1, 3, 1) = array([[[[1], [2], [3]]]])
-            ```
-            """
-            num_padded_entries_to_delete = min(
-                num_padded_entries, num_entries_to_delete
-            )
-            idxs_to_remove = [
-                num_padded_entries - i - 1 for i in range(num_padded_entries_to_delete)
-            ]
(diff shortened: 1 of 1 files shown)


src/deepsparse/transformers/utils/decoder_kv_cache.py
```diff
@@ -103,79 +103,78 @@ def update(
...
+        new_cache_array = cache_array[
+            :,
+            :,
+            bool(self._freeze_first_position) + num_non_padded_entries_to_delete :,
+            :,
+        ]
+        if self._freeze_first_position:
+            bos_entries = cache_array[:, :, :1, :]
+            new_cache_array = numpy.concatenate(
+                [bos_entries, new_cache_array], axis=self._sequence_len_axis
```
bfineran, line 174: could try benchmarking against idx selection

bfineran (commented):
LGTM pending comment

src/deepsparse/transformers/utils/decoder_kv_cache.py
dbogunowicz: Code used:

```python
from deepsparse.transformers.utils import DecoderKVCache
import numpy as np
import timeit

def update_session():
    session = DecoderKVCache()
    state = np.random.random((1, 32, 1000, 64))
    session.setup(session_id="dummy_id",
                  state=state,
                  num_processed_tokens=0,
                  freeze_first_position=1,
        )
    for _ in range(1000):
        session.remove_non_padded_entries(state, 500)

# run timeit on the test function
print(timeit.timeit("update_session()", setup="from __main__ import update_session", number=10))
```

src/deepsparse/transformers/utils/decoder_kv_cache.py
```diff
@@ -103,79 +103,78 @@ def update(
             Corresponds to `input_ids.shape[1]`
         """
         self.total_num_processed_tokens += input_ids_len
-        total_cache_capacity = state[list(state.keys())[0]].shape[
+
```
rahul-tuli, line 106: Could we also update the docstring above on line 94; there's a typo
`taking the kv cache output of from the forward pass` --> `taking the kv cache output from the forward pass`

src/deepsparse/transformers/utils/decoder_kv_cache.py
```diff
@@ -103,79 +103,78 @@ def update(
...
+        starting from the second oldest entry.
+        """
+        new_cache_array = cache_array[
+            :,
+            :,
+            bool(self._freeze_first_position) + num_non_padded_entries_to_delete :,
+            :,
+        ]
+        if self._freeze_first_position:
+            bos_entries = cache_array[:, :, :1, :]
```
rahul-tuli, line 172: What does bos stand for? beginning of sentence?

dbogunowicz: yes!

src/deepsparse/transformers/utils/decoder_kv_cache.py
dbogunowicz: Good catch!

src/deepsparse/transformers/utils/decoder_kv_cache.py
dbogunowicz: @rahul-tuli Ben probably referred to this function: https://github.com/neuralmagic/deepsparse/blob/cbc63319e3e85d2f9237559ddd0a5c7aad52735b/src/deepsparse/transformers/pipelines/text_generation.py#L422. This was how we used to (naively) do the "cache reassembly" back in the days. My code above tries to mimic this (however, avoiding the use of np.delete function). I can quickly benchmark your idea as well.

dbogunowicz: @rahul-tuli 
Tried this implementation:
```python
B, H, L, E = cache_array.shape
new_cache_array = numpy.empty((B, H, L - num_non_padded_entries_to_delete, E))
if self._freeze_first_position:
    new_cache_array[:,:,0,:] = cache_array[:,:,0,:]
    new_cache_array[:,:,1:,:] = cache_array[:,:,num_non_padded_entries_to_delete+1:,:]
else:
    new_cache_array[:,:,:,:] = cache_array[:,:,num_non_padded_entries_to_delete:,:]
```

roughly 1,25x slower then current implementation when freezing tokens
roughly 60x slower then current implementation when not freezing tokens

rahul-tuli: Thanks Damian! makes sense to me now :)

rahul-tuli (approved):

bfineran (approved):

PR #1191, opened 2023-08-17
[Text Generation] [Fix] Raise error when we use deepsparse engine and `prompt_processing_length == sequence_lenght`

dbogunowicz:
When we try to create a multi-token deepsparse engine where `prompt_processing_length == sequence_length`, this engine will default to onnxruntime:
```bash
[nm_ort 7f9cd3579740 >WARN<  is_supported_graph src/onnxruntime_neuralmagic/supported/ops.cc:144] Warning: Optimized runtime disabled - Detected dynamic input past_key_values.0.key dim 2. Set inputs to static shapes to enable optimal performance.
```
this is a known issue reported by the runtime team. The crux is: if `prompt_processing_length == sequence_length`, the initial kv cache tensors will be empty, and this is prohibitive in the engine - we cannot have empty kv cache values in there.

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -224,7 +224,7 @@ class TextGenerationPipeline(TransformersPipeline):
                     "process a prompt with a larger processing length. "
                     "However, it is assumed that `prompt_processing_sequence_length` "
                     "is smaller than the `sequence_length`. "
-                    "Please adjust the `prompt_processing_sequence_length` "
+                    "Adjust the `prompt_processing_sequence_length` "
                     "argument accordingly."
                 )
 
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -214,6 +214,20 @@ class TextGenerationPipeline(TransformersPipeline):
         engine, multitoken_engine = None, None
 
         if self.cache_support_enabled:
+            if (
+                self.engine_type == DEEPSPARSE_ENGINE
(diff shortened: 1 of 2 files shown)


rsnm2:
@dbogunowicz 

This makes sense.

However, the eval_downstream script sets the prompt_prefill and sequence_length equal to the same value. So now running eval_downstream will fail with your error message

https://github.com/neuralmagic/deepsparse/blob/main/src/deepsparse/transformers/eval_downstream.py#L89

dbogunowicz:
@rsnm2 yeah, missing commit. fixed :)

rahul-tuli:
Note: failing tests should be fixed by https://github.com/neuralmagic/sparsezoo/pull/357

bfineran (dismissed):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -214,6 +214,20 @@ def initialize_engines(
...
+                self.engine_type == DEEPSPARSE_ENGINE
+                and self.sequence_length <= self.prompt_processing_sequence_length
+                and self.enable_multitoken_prefill
+            ):
+                raise ValueError(
+                    "Attempting to initialize auxiliary DeepSparse engine to "
+                    "process a prompt with a larger processing length. "
+                    "However, it is assumed that `prompt_processing_sequence_length` "
+                    "is smaller than the `sequence_length`. "
+                    "Please adjust the `prompt_processing_sequence_length` "
```
dbogunowicz, line 227:
```suggestion
                    "Adjust the `prompt_processing_sequence_length` "
```

Satrat (approved):

rahul-tuli (approved):

PR #1192, opened 2023-08-17
Bug Fix: Make Numpy Array Outputs JSON Serializable for Server

Satrat:
**Asana Ticket:** https://app.asana.com/0/1205229323407165/1205248270105783/f 

When a TextGeneration pipeline is run through DeepSparse Server with `return_logits=true`, the request fails because the logits are returned as `numpy.ndarray` which is not JSON serializable. To fix this, I added a helper function that converts numpy arrays to lists before returning them to the client

### Testing
Launch a server:
`deepsparse.server --config_file server_error_config.yaml`

server_error_config.yaml:
```
{
  "name": "opt_model_server_config",
  "endpoints": [
    {
      "task": "text_generation",
      "model": "zoo:nlg/text_generation/codegen_multi-350m/pytorch/huggingface/bigquery_thepile/base-none",
      "route": "/gen",
    }
  ]
}
```

Send request from python, it should now be processed correctly
```
>>> r = requests.post("http://0.0.0.0:5543/gen", json={"sequences": ["def is_palindrome(word):"], "return_logits": True})
>>> r.status_code
200
>>>r.text
(This will print the very long logits array)
```

### DIFF
--- a/src/deepsparse/server/helpers.py
+++ b/src/deepsparse/server/helpers.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 from typing import Dict, List, Optional
+from pydantic import BaseModel
+import numpy
 
 from deepsparse import (
     BaseLogger,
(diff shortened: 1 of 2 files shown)


bfineran (approved):
LGTM overall - need to make sure that there are no pipelines that by default return large numpy arrays as tolist can slow down the pipeline significantly

rahul-tuli (approved):

PR #1197, opened 2023-08-23
Refactor of perplexity computation

anmarques:
Refactor intended to simplify perplexity computation and add support for different datasets into the same codebase. Among the changes, these are the highlights:

1. The perplexity class is now agnostic of the pipeline. It only sees predictions and targets. It does have an argument called "accumulate" that indicates whether to compute ppl for each sample separately or accumulate across samples.
2. Handle masking of tokens and logits in the perplexity_eval function. To avoid complications regarding the attention mask not being uniform within a batch, process each sample separately even if the pipeline is executed in batched mode.
3. Add logic to split data for wikitext such that each sample has the same number of tokens.

**Testing plan:**

Verified ppl for base Codegen 350M mono:
- deepsparse.transformers.eval_downstream <path to codegen model> --batch-size 16 --max-sequence-length 1024 -d openai_humaneval
Result: mean ppl: 3.60 (PyTorch: 3.60)

Verified ppl for OPT base 1.3b:
- deepsparse.transformers.eval_downstream <path to opt model> --batch-size 16 --max-sequence-length 2048 -d wikitext
Result: mean ppl: 14.62 (PyTorch: 14.63)

~~**NOTE**: This pipeline was only tested for non-cached models. It should work with kv-cache models as well. Right now the pipeline is created with `sequence_length=args.max_sequence_length` and `prompt_processing_sequence_length=args.max_sequence_length`. As soon as the kv-cache issues around this case are resolved we should test ppl evaluation again.~~

**Update**: Added support to c4 dataset in a way that complies with both the subsets defined in SparseGPT and LLM-foundry. Validated on cached models as well.


src/deepsparse/transformers/eval_downstream.py
```diff
@@ -71,16 +71,56 @@
...
 from datasets import load_dataset, load_metric  # isort: skip
 
 
-def perplexity_eval(args, batch_size=16, dataset_name="openai_humaneval"):
-    if args.max_samples:
-        batch_size = min(batch_size, args.max_samples)
-
-    dataset = load_dataset(dataset_name)["test"]
+def perplexity_eval(args, dataset_name="openai_humaneval"):
+    if dataset_name == "wikitext":
```
bfineran, line 79: would be nice to move this into a helper function

bfineran (approved):

src/deepsparse/transformers/eval_downstream.py
anmarques: Do you mean the split of the dataset? Just pushed an update

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -89,22 +102,64 @@ def perplexity_eval(args, batch_size=16, dataset_name="openai_humaneval"):
...
+                # Need to remove tokens that were masked
+                input_ids = prediction.input_tokens["input_ids"][s].flatten()
+                logits = prediction.logits[s]
+                attention_mask = prediction.input_tokens["attention_mask"][s].flatten()
+
+                sequence_length = logits.shape[0]
+                attention_mask = attention_mask[:sequence_length]
+                input_ids = input_ids[:sequence_length]
+
+                print(attention_mask.shape)
```
dbogunowicz, line 148: Is this on purpose?

src/deepsparse/transformers/metrics.py
```diff
@@ -36,136 +31,88 @@
...
-    def __init__(self, pipeline: Pipeline, batch_size: int = 16):
+    def __init__(self, accumulate: bool = False):
         """
-        Given the pipeline, compute the perplexity of the model
-        on the given text input.
-
-        Code adapted from:
-        https://huggingface.co/spaces/evaluate-metric/perplexity/blob/main/perplexity.py # noqa: E501
+        Class for computing perplexity.
+        """
```
dbogunowicz, line 37: missing params for `accumulate`

src/deepsparse/transformers/eval_downstream.py
anmarques: Nope. That was for debugging 😁 . Removed

src/deepsparse/transformers/metrics.py
anmarques: Done

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -75,36 +76,97 @@
...
+    # Along with the token, it returns the logits for input sequence
     text_generation = Pipeline.create(
         task="text-generation",
         model_path=args.model_path,
         engine_type=args.engine,
         num_cores=args.num_cores,
         sequence_length=args.max_sequence_length,
         max_generated_tokens=1,
+        trust_remote_code=args.trust_remote_code,
+        batch_size=args.batch_size,
```
dbogunowicz, line 109:
```suggestion
```

@anmarques this causes the error.

Currently, we cannot run LLM models in the engine with `batch_size != 1`. This is the batch size of the engine, to be clear, not the batch size of the evaluation script.

Now:
The engine inputs at batch 16, get split into a input of a batch size 16 and this gives you the error.


**After you accept the changes:**
The engine inputs at batch 16, will be properly split into 16 separate inputs of batch size 1 band each batch will be processed by the engine that assumes batch_size=1

anmarques: Addressed

dbogunowicz (dismissed):

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -462,6 +521,77 @@ def _split_train_val(train_dataset, val_ratio, seed=42):
     return train_ds, val_ds
 
 
+def _process_concatenated_datasets(dataset_name, model_path, max_sequence_length, kwargs):
```
bfineran, line 524: can we move these to a helpers file? eval_downstream is getting a bit long so trying to avoid too much task specific logic

anmarques: Done. To limit the scope of the PR to the PPL evaluation, mI moved only the helper functions defined in this PR to utils/eval_helpers. It's probably best to move other helper functions in a future PR.

dbogunowicz (dismissed):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -179,6 +183,18 @@ class TextGenerationOutput(BaseModel):
...
+    input_tokens: Optional[
+        Any
+    ] = Field(  # dictionary mapping "token_ids" and "attention_mask" to numpy arrays
+        default=None,
+        description="The output of the tokenizer."
+        "Dictionary containing token_ids and attention_mask, "
+        "both mapping to arrays of size "
+        "[batch_size, sequence_length]",
+    )
+    session_id: Optional[str] = Field(
```
dsikka, line 195: need rebase. `session_id` is no longer part of this class.

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -179,6 +183,18 @@ class TextGenerationOutput(BaseModel):
         "prompt provided. If streamng is enabled, the next generated token is returned."
         "Otherwise, the full generated sequence is returned."
     )
+    input_tokens: Optional[
```
dsikka, line 186: do we want to add additional attributes to the text_generation output? cc @markurtz  @bfineran

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -88,6 +88,10 @@ class Config:
         alias="prompt",
         description="The input sequences to generate the text from.",
     )
+    return_input_tokens: bool = Field(
```
dsikka, line 91: same comment as below

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -70,41 +70,101 @@
...
+    # We'll use the text generation pipeline to generate a single token.
+    # Along with the token, it returns the logits for input sequence
     text_generation = Pipeline.create(
         task="text-generation",
         model_path=args.model_path,
         engine_type=args.engine,
         num_cores=args.num_cores,
         sequence_length=args.max_sequence_length,
-        max_generated_tokens=1,
+        trust_remote_code=args.trust_remote_code,
```
dsikka, line 107: Because this is so dependent on the pipeline being created correctly, please add a unit test which tests a pipeline created with these args. This would help catch any pipeline updates which may no longer comply with your creation

src/deepsparse/transformers/utils/eval_helpers.py
```diff
@@ -0,0 +1,94 @@
+from transformers import AutoTokenizer
+from datasets import load_dataset
+import numpy
+
+
+def process_concatenated_datasets(dataset_name, model_path, max_sequence_length, kwargs):
```
dsikka, line 6: docstring, typing

src/deepsparse/transformers/utils/eval_helpers.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+        else:
+            raw_dataset = load_dataset(
+                "allenai/c4",
+                "allenai--c4",
+                split="validation",
+            )
+        if raw_samples is not None:
+            raw_dataset = raw_dataset[:raw_samples]
+        raw_text = raw_dataset["text"]
```
dsikka, line 33: This function uses a ton of hardcoded names and expects a certain pattern for file names. 

- Please add a test to ensure the filename assumptions hold and the different if/else conditions are valid for whatever inputs we expect
- Could we define constants at the top and use them here?

src/deepsparse/transformers/utils/eval_helpers.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Dataset is split into sections that contain "max_sequence_length" tokens.
+    # To split the dataset, first tokenize text
+    tokenizer = AutoTokenizer.from_pretrained(model_path)
+    return _split_text_by_tokens(
+        raw_text, eos, bos, tokenizer, max_sequence_length, kwargs.get("max_text_length", None)
+    )
+
+
+def _split_text_by_tokens(text, eos, bos, tokenizer, sequence_length, max_text_length):
```
dsikka, line 43: docstring, typing

src/deepsparse/transformers/metrics.py
```diff
@@ -231,19 +183,15 @@ def compute(self) -> Dict[str, float]:
...
-    """
-    Import and return the required torch module. Raises an ImportError if torch is not
-    installed.
-
-    :raises ImportError: if torch is not installed
-    :return: torch module
-    """
-    try:
-        import torch
+def _cross_entropy(predictions, targets, reduction="mean"):
```
dsikka, line 186: docstring, typing

src/deepsparse/transformers/metrics.py
```diff
@@ -36,139 +31,96 @@
...
-            self.perplexities.extend(perplexity_batch.numpy().tolist())
+            # Store perplexities
+            if self._perplexities is None:
+                self._perplexities = perplexities
+            else:
+                self._perplexities = numpy.concatenate(
+                    (self._perplexities, perplexities)
+                )
 
     def compute(self) -> Dict[str, Any]:
```
dsikka, line 108: nit: generic

src/deepsparse/transformers/metrics.py
```diff
@@ -36,139 +31,96 @@
 
 
 class Perplexity:
-    def __init__(self, pipeline: Pipeline, batch_size: int = 16):
+    def __init__(self, accumulate: bool = False):
```
dsikka, line 34: Could you provide an lifecycle in the docstring?

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -70,41 +70,101 @@
...
+            run_inference = True
+            end_evaluation = True
+
+        if (idx + 1) % args.batch_size == 0 or idx == dataset_length - 1:
+            run_inference = True
+
+        if run_inference:
+            # Perform single token generation
+            prediction = text_generation(
+                sequences=batch_samples,
```
dsikka, line 135: similarly, we should have a test which matches the inputs that are being used here

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -70,41 +70,101 @@
...
+                input_ids = prediction.input_tokens["input_ids"][s].flatten()
+                logits = prediction.generations[s].score
+                attention_mask = prediction.input_tokens["attention_mask"][s].flatten()
+
+                effective_sequence_length = logits.shape[0]
+
+                input_ids = input_ids[-effective_sequence_length:]
+                attention_mask = attention_mask[-effective_sequence_length:]
+
+                logits = numpy.compress(attention_mask, logits, axis=0)[:-1, :]
```
dsikka, line 155: we always expect these dimensions/indexing?

dsikka (changes requested):
Code is slightly hard to follow at the moment without docstrings or examples.  We should also probably be adding more testing to verify the pipeline being used in the current configuration is current/up-to-date with the inputs expected. Since we have a dependency on many hard-coded names/dimensions/indexes, we should probably add tests for them as well, to make sure they're always valid.

src/deepsparse/transformers/eval_downstream.py
```diff
@@ -604,7 +675,24 @@ def parse_args():
         type=bool,
         default=False,
     )
-
+    parser.add_argument(
```
bfineran, line 678: this is only supported for text gen as of this PR - should add a comment and a check or warning to make the user aware

src/deepsparse/transformers/utils/eval_helpers.py
bfineran: @anmarques was this adapted from somewhere or new code we are writing? If adapted, let's get a link in.  In either case let's make sure to include a docstring that specifies what is returned and the scope of the function

src/deepsparse/transformers/pipelines/text_generation.py
anmarques: Fixed

src/deepsparse/transformers/utils/eval_helpers.py
anmarques: Fixed

src/deepsparse/transformers/metrics.py
anmarques: Fixed

src/deepsparse/transformers/metrics.py
anmarques: I don't understand the comment

src/deepsparse/transformers/eval_downstream.py
anmarques: Yes

src/deepsparse/transformers/metrics.py
anmarques: I don't understand the comment

src/deepsparse/transformers/utils/eval_helpers.py
anmarques: Fixed

src/deepsparse/transformers/eval_downstream.py
anmarques: Fixed

src/deepsparse/transformers/metrics.py
dsikka: this function name is very generic. if we're matching hf specs, then it's fine

src/deepsparse/transformers/metrics.py
dsikka: Provide an example of what functions are called when, from start to finish

src/deepsparse/transformers/eval_downstream.py
anmarques: Added tests. Would appreciate if you could review them. First time I adde checks in this framework.

src/deepsparse/transformers/metrics.py
anmarques: Yes, using hf specs

src/deepsparse/transformers/eval_downstream.py
anmarques: Added tests. Would appreciate if you could review them. First time I adde checks in this framework.

src/deepsparse/transformers/metrics.py
anmarques: Done

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -666,6 +666,28 @@ def _test_kv_cache_state(
                 x[:, :, -start_index:-end_index, :], y, atol=_PRECISION
             )
 
+    def test_pipeline_for_ppl_eval(self, ):
+        pipeline = self.get_pipeline(
```
rahul-tuli, line 670: nice!

rahul-tuli (dismissed):
Lgtm pending comments and quality!

bfineran (approved):

src/deepsparse/transformers/eval_downstream.py
dsikka: where are the tests added to?

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: bumping this @markurtz @bfineran

dsikka (commented):
looks a lot better. still need to verify testing cases. Could you point to where those files are? Don't seem to be a part of this PR.

bfineran (dismissed):

dbogunowicz (dismissed):

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -104,6 +104,29 @@ def test_token_generation_non_deterministic(pipeline, prompt):
     assert len(set(text_outputs)) == 3
 
 
+def test_pipeline_for_ppl_eval(self):
```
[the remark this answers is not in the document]
anmarques: Yes, the test does not contemplate accuracy, only functionality. It assesses that the pieces needed from the pipeline are present, so changes to the pipeline don't break ppl evaluation. I don't think that assessing accuracy would be in the scope of a unit test.

I just realized that the testing interface changed and that the unit test should take a pipeline and a prompt instead of "self". I'm fixing that

dsikka (approved):

dbogunowicz (approved):

PR #1204, opened 2023-08-25
[Feature] Add callback for Text Generation Pipelines

rahul-tuli:
This PR adds a callback attribute to TextGenerationInput

`callback` is a callable that will be invoked on each generated token, and returns a boolean value.

If callback returns false, further generation of tokens is stopped

Test Code:

```python

start = 0
def dummy_callback(token):
    global start
    start += 1
    return start < 3


if __name__ == "__main__":
    codegen_pipeline = Pipeline.create(
        task="text_generation",
        model_path = "zoo:nlg/text_generation/codegen_mono-350m/pytorch/huggingface/bigpython_bigquery_thepile/base-none"
    )
    
    inputs = {
        "sequences": ["def fib(a, b, accumulator=0)"]*3,
        "callback": dummy_callback,
        "return_logits": True,
    }
    
    outputs = codegen_pipeline(**inputs)
    assert ( shape:=outputs.logits.shape[1]) == 3, f"Expected 3 tokens, got {shape}"
```

Also added above logic in an automated test (This is not yet active on GHA, but tested locally)

---
  - https://app.asana.com/0/0/1205276886236977

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -469,7 +469,7 @@ class TextGenerationPipeline(TransformersPipeline):
                         break
 
                     if callback is not None and callback(token) is False:
-                        _LOGGER.info(
+                        _LOGGER.debug(
                             "callback %s returned False, stopping generation."
                             % callback.__qualname__
                         )
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -16,7 +16,7 @@ import logging
 import os
 import warnings
 from dataclasses import dataclass
-from typing import Any, Dict, Generator, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Type, Union
 
 import numpy
 import onnx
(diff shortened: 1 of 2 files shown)


dbogunowicz (dismissed):
LGTM, but still not sure how useful this function is. Was this a request from product/QA?

Satrat (dismissed):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -457,6 +466,13 @@ def engine_forward(
                     ):
                         break
 
+                    if callback is not None and not callback(token):
+                        _LOGGER.info(
```
bfineran, line 470: This should be a debug log. Shouldn't log this on each inference

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -91,6 +91,14 @@ class Config:
         "`streamer.put(token_ids)` and the streamer is responsible "
         "for any further processing.",
     )
+    callback: Optional[Callable[[Any], bool]] = Field(
+        default=None,
+        description="Callable that will be invoked "
+        "on each generated token. The callable must return a "
+        "Boolean value. If invocation returns `True`, the "
```
bfineran, line 98: Boolean value not required. Can return None for no effect

src/deepsparse/transformers/pipelines/text_generation.py
rahul-tuli: No, this *only* logs when the callback returns False; which should be done right?

src/deepsparse/transformers/pipelines/text_generation.py
rahul-tuli: Updated logic to only stop when `False` is returned

Satrat (approved):

bfineran (approved):

PR #1206, opened 2023-08-25
Add stop argument to TextGenerationInput

rahul-tuli:
This PR adds a stop argument to TextGenerationInputSchema

Argument description:
```
A string or a list of strings that will be used as stop tokens. (token generation will stop when any of the stop tokens is generated). Set to `None` to ignore this parameter. Default is `None`."
```

Local Test:

```python



codegen_pipeline = Pipeline.create(
    task="text_generation",
    model_path = "zoo:nlg/text_generation/codegen_mono-350m/pytorch/huggingface/bigpython_bigquery_thepile/base-none"
)

inputs = {
    "sequences": "def fib(a, b, accumulator=0)",
    "stop": ["def", "int"],
}

outputs = codegen_pipeline(**inputs)
    
print(outputs.sequences)
```

Output:
```bash
)
2023-08-25 15:12:00 deepsparse.transformers.utils.helpers INFO     Overwriting in-place the input shapes of the transformer model at /home/rahul/.cache/sparsezoo/neuralmagic/codegen_mono-350m-bigpython_bigquery_thepile-base/model.onnx
2023-08-25 15:12:08 deepsparse.transformers.pipelines.text_generation INFO     Stop token int generated. Stopping generation.
```

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -497,7 +497,7 @@ class TextGenerationPipeline(TransformersPipeline):
                             else decoded_token.strip()
                         )
                         if decoded_token in stop:
-                            _LOGGER.info(
+                            _LOGGER.debug(
                                 "Stop token %s generated. Stopping generation."
                                 % decoded_token
                             )
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -16,7 +16,18 @@ import logging
 import os
 import warnings
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Type, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 import numpy
 import onnx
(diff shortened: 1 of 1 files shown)


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -456,6 +465,20 @@ def engine_forward(
...
+                    if stop is not None:
+                        decoded_token = self.tokenizer.decode(token)
+
+                        decoded_token = (
+                            decoded_token
+                            if decoded_token.isspace()
+                            else decoded_token.strip()
+                        )
+                        if decoded_token in stop:
+                            _LOGGER.info(
```
bfineran, line 477: Should be debug log

Satrat (commented):
Instead of decoding each token to compare it to the stop, why don't we just encode the stop tokens to make the comparison easier?

src/deepsparse/transformers/pipelines/text_generation.py
rahul-tuli: Updated, was doing it nonetheless when I saw your comment on other PR

rahul-tuli:
> Instead of decoding each token to compare it to the stop, why don't we just encode the stop tokens to make the comparison easier?

I tried that approach, the issue there is as follows:


Let's consider the same example as in the test code above, where the `stop=["def", "int"]`
Issue is that sometimes the model produces `" int"` as a token, which will have a different token_id compared to `"int"`
```python
>>> self.tokenizer("int")
{'input_ids': [600], 'attention_mask': [1]}

>>> self.tokenizer(" int")
{'input_ids': [493], 'attention_mask': [1]}
```
And this happens quite a few times; we do need the string to be able to at least remove white spaces, however I'm open to any other ideas?

bfineran (dismissed):

Satrat (dismissed):

Satrat (approved):

bfineran (approved):

PR #1208, opened 2023-08-26
Increase default text_generation sequence_length to 512

mgoin:
The default sequence length of 128 for all transformers pipelines is uncommonly small for text generation. I chose 512 because it seemed like a good middle-ground that all models can support but isn't so large as to affect performance or memory, like 2048. Final value is open to discussion but making it an explicit top-level argument is needed IMO


dbogunowicz (approved):
lgtm

PR #1142, opened 2023-07-24
[Text Generation] Automatically benchmark in auto-regressive setting

dbogunowicz:
When benchmarking an LLM, assert that input_ids length is one, so that benchmarks emulate the correct data.

## Manual Testing

1. Export the sample model
```bash
python kv_cache_injector.py --input-file deployment/model.onnx --output-file deployment/model_kvcache.onnx
```
2. Inject kv cache
```bash
python kv_cache_injector.py --input-file deployment/model.onnx --output-file deployment/model_kvcache.onnx
2023-07-24 12:50:46 sparseml.exporters.transforms.kv_cache.configs INFO     Loaded config file deployment/config.json for model: codegen
2023-07-24 12:50:46 sparseml.exporters.transforms.kv_cache.configs INFO     Properly configured arguments for KV Cache Transformation
2023-07-24 12:50:48 sparseml.exporters.transforms.onnx_transform INFO     [CacheKeysAndValues] Transformed 40 matches
2023-07-24 12:50:52 sparseml.exporters.transforms.onnx_transform INFO     [PositionsAdjustmentCodeGen] Transformed 7 matches
Modified model saved to: deployment/model_kvcache.onnx
```
3. Benchmark
```bash
deepsparse.benchmark /home/ubuntu/damian/sparseml/deployment/model_kvcache.onnx --sequence_length 256

2023-08-01 10:09:08 deepsparse.benchmark.benchmark_model INFO     Thread pinning to cores enabled
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
2023-08-01 10:09:11 deepsparse.transformers.utils.helpers INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/damian/sparseml/deployment/model.onnx
2023-08-01 10:09:16 deepsparse.benchmark.benchmark_model INFO     Found model that contains KV cache support. Benchmarking the autoregressive model with sequence length: 256.
DeepSparse, Copyright 2021-present / Neuralmagic, Inc. version: 1.6.0.20230727 COMMUNITY | (3cb4a3e5) (optimized) (system=avx2, binary=avx2)
2023-08-01 10:11:00 deepsparse.benchmark.benchmark_model INFO     deepsparse.engine.Engine:
        onnx_file_path: /home/ubuntu/damian/sparseml/deployment/model.onnx
        batch_size: 1
        num_cores: 23
        num_streams: 1
        scheduler: Scheduler.default
[...]

PR #1212, opened 2023-08-28
Automatically analyze in auto-regressive setting

mgoin:
Take the implementation for benchmark in https://github.com/neuralmagic/deepsparse/pull/1142 and apply to debug_analysis. Also update help usage and change default seq_len to 512

### DIFF
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -17,31 +17,40 @@ Benchmarking script for ONNX models with the DeepSparse engine.
 
 ##########
 Command help:
-usage: deepsparse.benchmark [-h] [-b BATCH_SIZE] [-shapes INPUT_SHAPES]
-                            [-ncores NUM_CORES] [-s {async,sync,elastic}]
-                            [-t TIME] [-w WARMUP_TIME] [-nstreams NUM_STREAMS]
-                            [-pin {none,core,numa}]
-                            [-e {deepsparse,onnxruntime}] [-q]
+usage: deepsparse.benchmark [-h] [-b BATCH_SIZE] [-seq_len SEQUENCE_LENGTH]
+                            [-input_ids_len INPUT_IDS_LENGTH]
+                            [-i INPUT_SHAPES] [-ncores NUM_CORES]
+                            [-s {async,sync,elastic}] [-t TIME]
+                            [-w WARMUP_TIME] [-nstreams NUM_STREAMS]
+                            [-pin {none,core,numa}] [-e ENGINE] [-q]
                             [-x EXPORT_PATH]
                             model_path
 
 Benchmark ONNX models in the DeepSparse Engine
 
 positional arguments:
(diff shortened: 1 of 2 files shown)


src/deepsparse/benchmark/benchmark_model.py
```diff
@@ -153,25 +167,21 @@ def parse_args():
...
         "-seq_len",
         "--sequence_length",
         type=int,
-        default=2048,
-        help="The sequence length to run the "
-        "KV cache supported model benchmarks for. "
-        "Must be greater than 0, default is 2048",
+        default=512,
+        help="The sequence length to run the KV cache supported model "
+        "benchmarks for. Must be 1 <= seq_len, default is 512",
```
tlrmchlsmth, line 176: >Must be 1 <= seq_len,

do normal people write it this way?

src/deepsparse/debug_analysis.py
```diff
@@ -49,10 +54,25 @@
...
                         Override the shapes of the inputs, i.e. -shapes
                         "[1,2,3],[4,5,6],[7,8,9]" results in input0=[1,2,3]
                         input1=[4,5,6] input2=[7,8,9]
+  --use-kvcache         Enable internal KVCache
+  --kv-cache-prev-num-tokens KV_CACHE_PREV_NUM_TOKENS
+                        KVCache: The amount of previous tokens that will be
+                        read from the external KV cache on the first inference
+  --kv-cache-num-frozen-tokens KV_CACHE_NUM_FROZEN_TOKENS
+                        KVCache: The amount of first tokens that we want to
+                        keep permanently in the KV cache
```
tlrmchlsmth, line 75: Could you default these to 0? I don't see us setting them very often.

src/deepsparse/debug_analysis.py
```diff
@@ -307,10 +349,31 @@ def main():
...
+    print(f"Analyzing model: {orig_model_path}")
 
     batch_size = args.batch_size
 
+    if has_model_kv_cache(model_path):
+        if batch_size != 1:
+            raise ValueError(
+                "Unable to run models with KV cache support "
+                "for batch size different than one."
+                "Please set batch size to 1 and try again"
```
tlrmchlsmth, line 361: Should we use `--use-kvcache` here instead of querying `has_model_kv_cache`?

mgoin: `--use-kvcache` just enables the internal kv cache and we want to be able to measure the difference. Probably `--use-kvcache` should be renamed

src/deepsparse/debug_analysis.py
mgoin: Yup good call and cleans up the logic

src/deepsparse/benchmark/benchmark_model.py
mgoin: lol i was making it the same as input_ids_length

tlrmchlsmth (approved):
lgtm, thanks!

PR #1214, opened 2023-08-29
[Text Generation][Enhancement] If `prompt_processing_sequence_length` == 1, do not inititalize multitoken_engine

dbogunowicz:
A simple improvement that streamlines the pipeline.

If `prompt_processing_sequence_length == 1`, we are essentially running single-token prompt prefill , so we should not be initializing and running the additional engine.

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -197,9 +197,14 @@ class TextGenerationPipeline(TransformersPipeline):
             _delay_engine_initialize=True,
             _delay_overwriting_inputs=True,
         )
-        self.enable_multitoken_prefill = self.causal_mask_input_present(
-            model_path=self.onnx_file_path
+        # enable multitoken prefill if
+        # - the model graph is supporting it (causal_mask input is present)
+        # - prompt_processing_sequence_length != 1 (identical to single-token prefill)
+        self.enable_multitoken_prefill = (
+            self.causal_mask_input_present(model_path=self.onnx_file_path)
+            and prompt_processing_sequence_length > 1
         )
+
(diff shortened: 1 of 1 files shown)


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -680,20 +685,14 @@ def engine_inputs_for_prefill(
...
-                            .reshape(1, -1)
-                            .astype(numpy.int64)
+                    engine_input = (
+                        numpy.arange(
+                            num_cached_entries,
+                            num_cached_entries + self.prompt_processing_sequence_length,
                         )
+                        .reshape(1, -1)
+                        .astype(numpy.int64)
+                    )
```
tlrmchlsmth, line 695: what's the reason for this change?

dbogunowicz: Handling the case where `prompt_processing_sequence_lenght == 1` vs `!=1` required some additional lines of code. If you look closely, I am actually removing these superfluous lines of code, and keeping only the essential logic for the creation of the engine_inputs for multitoken_engine.

dbogunowicz: (lines 684-686 are not needed anymore)

tlrmchlsmth: Will this function no longer work when `self.prompt_processing_sequence_length == 1`? Should we add an assert?

dbogunowicz: Those lines will be never triggered, will never end up using the multitoken engine in this scenario.

mgoin (approved):
awesome! a special case i'll take advantage of :)

PR #1215, opened 2023-08-29
Remove dependency on `transformers` for running `debug_analysis` and `benchmark_model` on KV-Cache supported LLMs

dbogunowicz:
Tested using running the relevant pathways and making sure that no errors are raised.


src/deepsparse/debug_analysis.py
```diff
@@ -363,17 +363,24 @@ def main():
                 "Please set batch size to 1 and try again"
             )
 
+        if args.input_ids_length > args.sequence_length:
+            raise ValueError(
+                f"input_ids_length: {args.input_ids_length} "
+                f"must be less than sequence_length: {args.sequence_length}"
```
mgoin, line 369: It can be less than or equal to, right?

mgoin (approved):
The runtime team appreciates it! Just left one comment on an error message

bfineran (approved):

src/deepsparse/debug_analysis.py
dbogunowicz: No. This would mean that we have an "empty' cache (of length zero), and that will crash the engine I think.

PR #1216, opened 2023-08-29
[Feature Branch][LLM Testing] Full Testing Harness for LLMs

dbogunowicz:
The implementation of the test harness for LLMs. By default, the tests are turned off so that we do not choke GHA.
To enable tests: remove `@pytest.mark.skip(reason="Those tests are too heavy to run as a normal part of the CI.")`
@pytest.mark.skip(reason="Those tests are too heavy to run as a normal part of the CI.")
and run `pytest tests/deepsparse/transformers/pipelines/test_text_generation.py`
Future consideration: adding config and utilizing small toy models to make tests extremely lightweight.

Includes PRs:
- #1227 
- #1219


dbogunowicz:
Note to myself: add a scenario when we run the same prompt twice with a different session id and the kv cache management gets solved properly.

tests/deepsparse/transformers/pipelines/helpers.py
```diff
@@ -0,0 +1,210 @@
...
+        self.session = onnxruntime.InferenceSession(self.model_onnx_path)
+        self.model_inputs = [
+            x.name
+            for x in onnx.load(
+                self.model_onnx_path, load_external_data=False
+            ).graph.input
+        ]
+
+    def tokenize(self, prompt: str):
+        return self.tokenizer(
```
dsikka, line 89: For this class, we pull the model from sparsezoo but the tokenizer is through hf. Why not pull it from the sparsezoo as well?

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -12,262 +12,396 @@
...
+
+        if self.use_deepsparse_cache:
+            pytest.skip(
+                "Cannot run ORT pipeline with the internal deepsparse cache enabled."
             )
-            hf_output = tokenizer.decode(generated_ids[0], skip_special_tokens=True)
-            hf_outputs.append(hf_output)
-        return hf_outputs
+        model_stub, _, _, torch_ground_truth = setup
+        pipeline = Pipeline.create(
```
dsikka, line 207: Ton of `Pipeline.create` calls throughout our tests. Is there anyway we could unify some of them/store it in fixtures?

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -12,262 +12,396 @@
...
-        (
-            "zoo:nlg/text_generation/opt-1.3b/pytorch/"
-            "huggingface/opt_pretrain/base-none",
-            "facebook/opt-1.3b",
-            True,
-        ),
         (
             "zoo:nlg/text_generation/codegen_mono-350m/pytorch/"
             "huggingface/bigpython_bigquery_thepile/base-none",
             "salesforce/codegen-350m-mono",
```
dsikka, line 38: If we're using codegen would it make sense to add a codegen-specific prompt? It may not matter since we're comparing it to the groundtruth from hugging face but both will essentially produce garbage for the current prompt?

dsikka (commented):
How confident are we in our test coverage? Possibly add tests when running with `deterministic` off or multiple input sequences?

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -12,262 +12,396 @@
...
+        #  "facebook/opt-1.3b",
+        #  True,
+        #  None),
     ],
     scope="class",
 )
-@pytest.mark.skip(
-    reason="Those tests are too heavy to " "run as a normal part of the CI."
-)
 class TestTextGenerationPipeline:
```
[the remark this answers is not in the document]
dbogunowicz: Yes. With a lower frequency (nightly)

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -12,262 +12,396 @@
...
+        self.num_tokens_generate = 216
+        self.prompt = """
+           Didn't know what time it was, the lights were low
+           I leaned back on my radio
+           Some cat was layin' down some rock 'n' roll
+           "Lotta soul," he said
+           Then the loud sound did seem to fade
+           Came back like a slow voice on a wave of phase
+           That weren't no DJ, that was hazy cosmic jive
+           """
```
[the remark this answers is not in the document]
dbogunowicz: Those tests will be long anyway :)
I'd rather test them for long prompts and a large number of generated sequences to be sure that the logits do not diverge from the ground truth in some early stages.

tests/deepsparse/transformers/pipelines/helpers.py
dbogunowicz: Good idea, I like that.

tests/deepsparse/transformers/pipelines/test_text_generation.py
dbogunowicz: Unifying is difficult, since each of those pipelines will differ significantly from the other.

tests/deepsparse/transformers/pipelines/test_text_generation.py
dbogunowicz: Hmm, adds a bit more complexity to the whole harness, but you have a good point there.

tests/deepsparse/transformers/pipelines/helpers.py
dbogunowicz: We need to use the model name anyway, plus what happens if we mess up and have a wrong tokenizer in our stub? Then our ground truth will not be ground truth anymore, it will have hidden dependency on the thing that we are actually testing.

tests/deepsparse/transformers/pipelines/test_text_generation.py
dbogunowicz: Acknowledged

tests/deepsparse/transformers/pipelines/test_text_generation.py
dbogunowicz: Did a nice trick to be less wasteful. (see get_prompt func)

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -12,262 +12,396 @@
...
-            "huggingface/opt_pretrain/base-none",
-            "facebook/opt-1.3b",
-            True,
-        ),
         (
             "zoo:nlg/text_generation/codegen_mono-350m/pytorch/"
             "huggingface/bigpython_bigquery_thepile/base-none",
             "salesforce/codegen-350m-mono",
             False,
+            15.5,
```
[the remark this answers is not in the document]
dbogunowicz: It is once again!

dsikka (changes requested):
- Remove ORT ground truth class and use deepsparse pipeline instead

bfineran (approved):
LGTM overall. as discussed offline - will need some refactors to move cleanly to a config based method

dsikka (approved):
LGTM. Support for getting this to run on a nightly basis is still pending?

PR #1217, opened 2023-08-29
[TextGeneration] max token refactor

dsikka:
For this ticket: 

https://app.asana.com/0/1201735099598270/1205276886236972/f

### Summary:
- Refactors the TextGeneration Constructor to remove the `max_tokens` argument and makes it part of the pipeline input
- Adds `num_generated_predictions` to the input as well, which dictates the number of sequences that are generated for a given input. Similar to the hugging face implementation, we repeat the input based on the number provided, defaulting to 1. When `num_generated_predictions` is > 1,  the engine's `deterministic` property is togged to False
- In the case where the value is > 1, the output is a list of lists, where each list includes the generated sequences for a given prompt. This updates the sequences output to be one of `str, List[str], and List[List[str]]`

### Testing
- Tested locally using the new input arguments
- Also added new tests to evaluate the `num_generated_predictions`

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -376,6 +376,14 @@ class TextGenerationPipeline(TransformersPipeline):
                 "Set `max_tokens` to 1 to support that scenario."
             )
 
+        # If the num_generated_predictions > 1, check how many times each prompt given
+        # as an input was repeated. If just once, the number of generated predictions
+        # for that particular prompt is equal to num_generated_predictions. This is done
+        # by repeating the prompt num_generated_predictions times If that
+        # prompt was given as an input multiple times, ignore the value of
+        # num_generated_predictions and produce n predictions, where n is the number of
+        # times the prompt was given as an input. Also, update the engine so that
+        # deterministic is set to False.
         if inputs.num_generated_predictions > 1:
             counter = Counter(inputs.sequences)
             repeated_seq = []
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -64,7 +64,7 @@ class TextGenerationInput(BaseModel):
         description="The number of text generations to create from a single prompt. If "
         "the same sequence is given as an input multiple times, the number of generated"
         "the number of generated predictins is equivalent to the number of times the "
-        "the sequence is repeated."
+        "the sequence is repeated.",
     )
     max_tokens: int = Field(
         default=1024,
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -385,6 +385,8 @@ class TextGenerationPipeline(TransformersPipeline):
         # times the prompt was given as an input. Also, update the engine so that
         # deterministic is set to False.
         if inputs.num_generated_predictions > 1:
+            if isinstance(inputs.sequences, str):
+                inputs.sequences = [inputs.sequences]
             counter = Counter(inputs.sequences)
             repeated_seq = []
             for seq in inputs.sequences:
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -15,6 +15,8 @@
 import logging
 import os
 import warnings
+from collections import Counter
+
 from typing import (
     Any,
     Callable,
@@ -56,6 +58,15 @@ class TextGenerationInput(BaseModel):
 
     sequences: Union[str, List[str]] = Field(
         description="The input sequences to generate the text from.",
+    )
+    num_generated_predictions: int = Field(
+        default=1,
+        description="The number of text generations to create from a single prompt",
+    )
+    max_tokens: int = Field(
+        default=1024,
+        description="Maximum number of tokens to generate per output sequence. If no "
+        "value is provided, will default to 1024.",
     )
     return_logits: bool = Field(
         default=False,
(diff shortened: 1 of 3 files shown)


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -358,6 +358,29 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
         :return: the inputs for the engine
         """
+        if not self.cache_support_enabled and inputs.max_tokens > 1:
+            raise ValueError(
+                "The model used for inference does not support kv cache. It is "
+                "assumed that it maps from the token sequence to predicted logits."
+                "Set `max_tokens` to 1 to support that scenario."
+            )
+
+        if inputs.num_generated_predictions > 1:
```
dbogunowicz, line 368: could you add a short comment that briefly describes what happens here (may not be obvious to somebody who picks up the code later)

dbogunowicz (commented):
The good: much less additional code and complexity that I thought
The bad and ugly: could you add appropriate tests in tests/deepsparse/transformers/pipelines/test_text_generation.py ?

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: yes will add. much of the code is if the user provides the same prompt as an input multiple times (as seems to be a common testing case). For that case, we ignore `num_generated_predictions` but we can also just treat them the same way.

bfineran (dismissed):

mgoin (commented):
Could `max_tokens` default to the `prompt_length - sequence_length` so we don't risk running out of kv cache context? I'm not sure what happens there actually, especially when using internal kv cache

dsikka:
> Could `max_tokens` default to the `prompt_length - sequence_length` so we don't risk running out of kv cache context? I'm not sure what happens there actually, especially when using internal kv cache

@dbogunowicz would like to get your opinion on this

dbogunowicz:
@dsikka this is a very good idea.

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -375,6 +374,39 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
         :return: the inputs for the engine
         """
+        if not self.cache_support_enabled and inputs.max_tokens > 1:
+            raise ValueError(
+                "The model used for inference does not support kv cache. It is "
+                "assumed that it maps from the token sequence to predicted logits."
+                "Set `max_tokens` to 1 to support that scenario."
+            )
+
+        # If the num_generated_predictions > 1, check how many times each prompt given
```
bfineran, line 384: let's move this to a helper function

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -375,6 +374,39 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
+        # as an input was repeated. If just once, the number of generated predictions
+        # for that particular prompt is equal to num_generated_predictions. This is done
+        # by repeating the prompt num_generated_predictions times If that
+        # prompt was given as an input multiple times, ignore the value of
```
bfineran, line 388: what's the reasoning behind this?

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -375,6 +374,39 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
...
+        # num_generated_predictions and produce n predictions, where n is the number of
```
bfineran, line 389: is there a reason we don't postprocess the generated sequences and group them by the original sequence idx now with multiple generated values? that way the user has a 1:1 mapping between their actual inputs and what they get back

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: If the user is already providing the same input multiple times, the number of generated predictions seems to be implied. Otherwise, if they provide an input 2 times, but then also set num_generated_predictions to 3 for example, that would be 6 predictions. But it makes more sense to just output 2 predictions.

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: Do you mean in general when the user provides multiple input sequences which are identical?

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: Caveat is that this doesn't follow hf behaviour so can remove

dsikka:
> Could `max_tokens` default to the `prompt_length - sequence_length` so we don't risk running out of kv cache context? I'm not sure what happens there actually, especially when using internal kv cache

Talking to the MLE team, I think for now we want to keep the defaults as is and update them once we've established best practices.

Satrat (approved):

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -135,6 +136,16 @@ def test_model_output_sequences(self, setup):
...
+    def test_num_generated_predictions(self, setup):
+        pipeline = setup[0]
+        short_prompt = setup[3]
+
+        output_sequences = pipeline(
+            sequences=[short_prompt], num_generated_predictions=2
+        )
+
+        assert len(output_sequences.sequences) == 2
+
```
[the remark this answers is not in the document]
dsikka: Can do!

Satrat (approved):

bfineran (approved):

PR #1218, opened 2023-08-29
[TransformersPipeline] Add in and refactor TransformersPipeline args

dsikka:
For this ticket: https://app.asana.com/0/1201735099598270/1205276886236966/f

Summary:
- Updates the `TransformersPipeline` constructor to add in two arguments: `config` and `tokenizer` 
- For both of these, the user can provide a string, path or transformers object which will be used as opposed to relying on a `deployment` directory with the expected json files. By default, these will both be `None` and in that case, the normal `deployment` directory workflow will be used. 
- Additionally, the `config` argument may also be a dictionary
- To support this functionality, the  `get_onnx_path_and_configs` is refactored/separated into two separate functions,  `get_hugging_face_configs` and `get_onnx_path`

## Testing:

- Tested locally using a variety of combinations for `config` and `tokenizer`

Example:

```
from transformers import LlamaConfig, LlamaTokenizerFast

tokenizer = LlamaTokenizerFast.from_pretrained("hf-internal-testing/llama-tokenizer")
config = {
   "_name_or_path": None,
   "architectures": [
      "LlamaForCausalLM"
   ],
   "bos_token_id": 1,
   "eos_token_id": 2,
   "hidden_act": "silu",
   "hidden_size": 5120,
   "initializer_range": 0.02,
   "intermediate_size": 13824,
   "max_position_embeddings": 4096,
   "model_type": "llama",
   "num_attention_heads": 40,
   "num_hidden_layers": 40,
   "num_key_value_heads": 40,
   "pretraining_tp": 1,
   "rms_norm_eps": 1e-05,
   "rope_scaling": None,
   "tie_word_embeddings": "false",
   "torch_dtype": "float16",
   "transformers_version": "4.31.0.dev0",
   "use_cache": "true",
   "vocab_size": 32000
}

llama = Pipeline.create(
   task="text-generation",
   model_path="/home/dsikka/models_llama/deployment_13",
   engine_type="onnxruntime",
   deterministic=False,
   config=config,
   tokenizer=tokenizer
)

inference = llama(sequences=["Hello?"])
for s in inference.sequences:
   print(s)
```


src/deepsparse/transformers/helpers.py
```diff
@@ -119,19 +130,21 @@ def get_onnx_path_and_configs(
...
     else:
         raise ValueError(
-            f"model_path {model_path} is not a valid file, directory, or zoo stub"
+            f"model_path {model_path} is not a valid directory or zoo stub"
+        )
+
+    if config_path is None or tokenizer_path is None:
+        warnings.warn(
+            f"Unable to find model or tokenizer configs for model_path {model_path}. "
+            f"model_path must be a directory containing config.json, and/or "
```
dbogunowicz, line 141: "and/or"? not "and"?

src/deepsparse/transformers/pipelines/pipeline.py
```diff
@@ -104,30 +120,51 @@ def sequence_length(self) -> Union[int, List[int]]:
...
+        if not self.config or not self.tokenizer:
+            config_found, tokenizer_found = get_hugging_face_configs(self.model_path)
+            if config_found:
+                self.config = config_found
+            if tokenizer_found:
+                self.tokenizer = tokenizer_found
+
+        if isinstance(self.config, dict):
+            local_config_path = os.path.join(self.model_path, "config.json")
+            with open(local_config_path, "w") as f:
```
dbogunowicz, line 144: do we really need to save an intermediate file to disk?

dsikka: From the hugging face docs, it seems like the `AutoConfig.from_pretrained` requires a file path of some sort: https://huggingface.co/docs/transformers/v4.32.1/en/model_doc/auto#transformers.AutoConfig.from_pretrained.pretrained_model_name_or_path but definitely open to not saving if there's a way around it.

src/deepsparse/transformers/helpers.py
dsikka: If only one is present it's fine as we no longer require a `deployment` folder with both files and can expect a config or tokenizer through one of the other now supported input types (paths, dict, etc)

bfineran (approved):

PR #1220, opened 2023-08-30
topk topp penalties

horheynm:


src/deepsparse/transformers/utils/helpers.py
```diff
@@ -211,3 +212,91 @@ def create_causal_mask(
...
+    #     while init_p < top_p:
+    #         next_most_likely_prob = next(probs_iter)
+    #         init_p += next_most_likely_prob
+    #         i += 1
+    #     probs = probs[:i]
+
+    return numpy.random.choice(len(probs), p=probs)
+
+
+def apply_sampling_penalties(
```
horheynm, line 267: apply_sampling_penalties

src/deepsparse/transformers/utils/helpers.py
```diff
@@ -211,3 +212,91 @@ def create_causal_mask(
     causal_mask[:, :, :, :num_zeros] = 0
 
     return causal_mask
+
+
+def generate_token(
```
dbogunowicz, line 217: NLDecoderEngine  --> calls generate_token(logits)

generate_token(logits) -> token
takes care of converting logits into token, i.e also handles all the sampling in between

sampling methods, I like if they are split into separate helper functions

PR #1222, opened 2023-08-31
[server] pin anyio support to <4.0.0

bfineran:
this is for compatibility with the maximum currently allowed version of `starlette`. `anyio` just pushed a 4.0.0 that is incompatible with earlier versions of `starlette`. 

Note that this is now fixed in later versions (`deepsparse` has not yet tested against newer versions of `fastapi` and `starlette`). See: https://github.com/encode/starlette/pull/1936

from internal bug report:
```
Our Deepsparse GHA is failing with starlette error: AttributeError: module 'anyio' has no attribute 'start_blocking_portal'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -124,6 +124,7 @@ _server_deps = [
     "python-multipart>=0.0.5",
     "prometheus-client>=0.14.1",
     "psutil>=5.9.4",
+    "anyio<4.0.0",
 ]
 _onnxruntime_deps = [
     "onnxruntime>=1.7.0",


mgoin (approved):
Looks good from GHA testing for that error. There are still other asserts for another issue

bfineran:
GHA failure looks unrelated and on the sparsezoo side - pushing this patch

bfineran:
Confirmed with @tdg5 the issue may be due to an internal migration.  Issue is now tracked internally with fix underway

PR #1224, opened 2023-09-01
[Text Generation][Fix] Running same prompt through KV cache models N times bad result when n > 1

dbogunowicz:
This fix rectifies the problem with erroneous KV cache management during prefill. The problem manifests itself under the following conditions:
- When we run the same pipeline twice.
- When we run it using the auxiliary multitoken engine, meaning there is a need for session sharing between the two engines.

Explanation:

The first time we run the inference, we follow these steps:
1. We call `self.multitoken_engine.reset_kv_cache()`, creating an empty KV cache session attached to the multi-token engine. This session is located in memory at `location_1`.
2. We run multitoken prefill, correctly populating the KV cache session object at `location_1` with KV cache values.
3. We call `engine.reset_kv_cache()`, creating an empty KV cache session attached to the engine. This session is located in memory at `location_2`.
4. In the scenario where we use the `multitoken_engine`, we also run `self.engine.transfer_cache_state(cache=self.multitoken_engine.kv_cache)`. The KV cache session of the `self.multitoken_engine` is now transferred to the `self.engine`. From this point on, both engines share the same session with the same memory location: `location_1`. This is a crucial fact.
5. Finally, we finish the inference by running the autoregressive inference.

The second time we run the inference, both engines share the same session object. This means that:
1. We call `self.multitoken_engine.reset_kv_cache()`, creating an empty KV cache session attached to the multi-token engine. This session is located in memory at `location_1`.
2. We run multitoken prefill, correctly populating the KV cache session object at `location_1` with KV cache values.
3. We call `engine.reset_kv_cache()`, which resets the KV cache session object at `location_1` and thus loses all the information collected during the multitoken prefill.

The solution is to introduce an `if-else` condition to reset the engine's KV cache only if we are running without multi-token engine support.

This may not be the most elegant solution, but it is the one that works and does not require too many code changes. In the future, I will ensure that this edge case is covered in testing and taken into consideration when refactoring -> landing `KVCacheSessionStorage` support.


dbogunowicz:
@SageMoore this fix is temporary, we will move to a different, improved architecture once I get the testing harness in place. I will definitely run the new PR through you, to make sure it looks good from your end.

mgoin (approved):
Appreciate the detailed analysis of what is happening!

Satrat (approved):

PR #1225, opened 2023-09-01
[TextGeneration] Samling arguments for generation

horheynm:
1. Add sampling into the text-generation pipeline
2. Refactor token generation flow

I also need https://github.com/neuralmagic/deepsparse/pull/1234 to land to check on failures caused by this pr if any

PR #1229, opened 2023-09-07
Chatbot demo

dbogunowicz:


examples/chatbot_llm/main.py
```diff
@@ -0,0 +1,25 @@
...
+    pipeline = Pipeline.create(
+        task="text-generation",
+        model_path=args.model_path,
+        max_generated_tokens = args.max_generated_tokens,
+    )
+
+    while True:
+        # get input from user
+        input_text = input("User: ")
+        response = pipeline(sequences=[input_text], session_ids=args.session_id)
```
bfineran, line 15: What's the point of having session id as an arg?

dbogunowicz: Yeah, for this small example probably should be a fixed string.

dbogunowicz:
@rahul-tuli my few cents on the design of CLI.
We should allow the user to modify:
- whether the generation is deterministic or not
- the temperature of the logits distribution
- max generated length 
- sequence_length 
- Maybe the parameters introduced in this PR by @horheynm : https://github.com/neuralmagic/deepsparse/pull/1225. I do not have yet a good understanding of how to tune them to get the most impressive conversion flow, but it feels like they can be helpful
Limitations:
- I'd probably keep it simple. Let's only input one sequence at a time.

Maybe beautifying the interface with some colors in stdout and emoticons (LLaMa2 famously is able to output emojis: https://github.com/abetlen/llama-cpp-python/issues/608)

dbogunowicz:
Closing, as this is now the duplicate of https://github.com/neuralmagic/deepsparse/pull/1235

bfineran:
out of date - closing

PR #1230, opened 2023-09-07
use tokenizer_config path if tokenizer is None

horheynm:
Opt models do not have tokenzier.json inside gitlab.
One approach is to use tokenizer_config.json instead.


src/deepsparse/transformers/helpers.py
```diff
@@ -111,12 +112,20 @@ def get_onnx_path_and_configs(
...
-        tokenizer_path = _get_file_parent(
-            zoo_model.deployment.default.get_file(_MODEL_DIR_TOKENIZER_NAME).path
-        )
         tokenizer_config_path = zoo_model.deployment.default.get_file(
             _MODEL_DIR_TOKENIZER_CONFIG_NAME
         )
+        tokenizer_path = _get_file_parent(
+            zoo_model.deployment.default.get_file(_MODEL_DIR_TOKENIZER_NAME).path
+        )
+        if tokenizer_path is None:
```
bfineran, line 121: what do we do if tokenizer config is also missing?

horheynm: tokenizer config is not mandatory?

dbogunowicz: tokenizer config should always be present. are you suggesting to introduce an explicit error-raising logic if this is the case, Ben?

dbogunowicz:
latest commit:
```python
2023-09-12 06:49:19 deepsparse.transformers.pipelines.text_generation INFO     Compiling an auxiliary engine to process a prompt with a larger processing length. This improves performance, but may result in additional memory consumption.
2023-09-12 06:49:19 deepsparse.utils.onnx INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/.cache/sparsezoo/neuralmagic/opt-1.3b-opt_pretrain-base/model.onnx
DeepSparse, Copyright 2021-present / Neuralmagic, Inc. version: 1.6.0.20230831 COMMUNITY | (f9ca3574) (release) (optimized) (system=avx2, binary=avx2)
2023-09-12 06:49:25 deepsparse.utils.onnx INFO     Overwriting in-place the input shapes of the transformer model at /home/ubuntu/.cache/sparsezoo/neuralmagic/opt-1.3b-opt_pretrain-base/model.onnx
```

Satrat (approved):

dbogunowicz:
@horheynm can you move this from a draft to a PR and add a PR description? I think we should get it reviewed and land it asap.

bfineran (approved):

PR #1232, opened 2023-09-08
Remove dependency on trust-remote-code

anmarques:
This PR is motivated to remove the dependency on the "trust-remote-code" argument for deepsparse transformers pipelines.

In addition to an additional flag, this argument also meant that custom scripts needed for the PyTorch definition of models needed to be carried over in the deployment folder, even though the PyTorch definition is not needed for deepsparse inference.

The dependency was introduced by using the AutoConfig class to load the model config. This PR replaces this with PretrainedConfig instead. AutoConfig is a class that wraps PretrainedConfig and does additional checks to guarantee that the loaded config matches what the model definition uses. By doing this replacement these checks are skipped and the model definitions is not needed any longer.

### Testing plan
 
 The model config at this moment is only used in the text-classification and token-classification pipelines. To validate this PR I evaluated accuracy on these pipelines and verified that it is the same as before. Here are the commands used for testing:

- `deepsparse.transformers.eval_downstream zoo:nlp/token_classification/obert-base/pytorch/huggingface/conll2003/pruned90_quant-none -d conll2003`

- `deepsparse.transformers.eval_downstream zoo:nlp/multilabel_text_classification/obert-base/pytorch/huggingface/goemotions/pruned90_quant-none -d go_emotions`

- `deepsparse.transformers.eval_downstream zoo:nlp/text_classification/obert-base/pytorch/huggingface/mnli/pruned90_quant-none -d mnli`

- `deepsparse.transformers.eval_downstream zoo:nlp/text_classification/obert-base/pytorch/huggingface/qqp/pruned90_quant-none -d qqp`

### Update

@mgoin Pointed out that some models (e.g., replit) can have custom tokenizers and thus trust_remote_code would still be needed for the tokenizer. Hence, updated the code to only remove the argument from the config loading, but still keep it around for the tokenizer.

### DIFF
--- a/src/deepsparse/transformers/pipelines/pipeline.py
+++ b/src/deepsparse/transformers/pipelines/pipeline.py
@@ -24,7 +24,7 @@ from typing import Any, Dict, List, Mapping, Optional, Union
 
 import numpy
 import transformers
-from transformers.models.auto import AutoConfig, AutoTokenizer
+from transformers.models.auto import AutoTokenizer
 
 from deepsparse import Bucketable, Pipeline
 from deepsparse.transformers.helpers import (
@@ -154,28 +154,16 @@ class TransformersPipeline(Pipeline, Bucketable):
             else:
                 self.config_path = os.path.join(self.config, "config.json")
 
-            old_config = AutoConfig.from_pretrained(
-                self.config,
-                trust_remote_code=self._trust_remote_code,
-                finetuning_task=self.task if hasattr(self, "task") else None,
-            )
-
-            new_config = transformers.PretrainedConfig.from_pretrained(
+            self.config = transformers.PretrainedConfig.from_pretrained(
                 self.config,
                 finetuning_task=self.task if hasattr(self, "task") else None,
             )
-
-            print(new_config)
-            print(old_config)
-            exit()
-
 
         if isinstance(self.tokenizer, (str, Path)):
             self.tokenizer_config_path = os.path.join(self.tokenizer, "tokenizer.json")
 
             self.tokenizer = AutoTokenizer.from_pretrained(
                 self.tokenizer,
-                trust_remote_code=self._trust_remote_code,
                 model_max_length=self.sequence_length,
             )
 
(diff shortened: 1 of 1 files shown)


src/deepsparse/transformers/pipelines/pipeline.py
```diff
@@ -132,28 +132,30 @@ def setup_onnx_file_path(self) -> str:
         """
         onnx_path = get_onnx_path(self.model_path)
 
+        print(self.config)
```
dsikka, line 135: no prints

src/deepsparse/transformers/pipelines/pipeline.py
```diff
@@ -162,7 +164,6 @@ def setup_onnx_file_path(self) -> str:
 
             self.tokenizer = AutoTokenizer.from_pretrained(
                 self.tokenizer,
-                trust_remote_code=self._trust_remote_code,
                 model_max_length=self.sequence_length,
```
dsikka, line 167: If we're removing this, we have to update the pipeline args as we're now taking in an argument that isn't being used anywhere

dsikka (changes requested):

src/deepsparse/transformers/pipelines/pipeline.py
anmarques: debugging prints 😁 . Removed

dbogunowicz (dismissed):
great add @anmarques

src/deepsparse/transformers/pipelines/pipeline.py
mgoin: @anmarques it is probably still necessary to allow for the tokenizer for some cases. i had to include this when using replit-code for example https://huggingface.co/replit/replit-code-v1-3b/blob/main/replit_lm_tokenizer.py

bfineran (approved):

src/deepsparse/transformers/pipelines/pipeline.py
anmarques: Updated to code to leave trust_remote_code in place as optional argument for tokenizer

dbogunowicz (approved):

PR #1233, opened 2023-09-08
SparseZoo V2 Compatibility

Satrat:
This PR fixes two issues related to the SparseZoo V2 upgrade:
1. `sample_originals` was deleted from some models but still being referenced in tests. The tested were updated to use `sample_inputs` instead
2. ~~The new sparsezooAPI includes subfolder names in the `display_name` field of `Directories` when previously it did not~~
~~**before(V1):** `{'display_name': 'config.json', 'file_size': 799, 'model_id': '41dc6dcb-88cd-45f4-ae5e-8c34c0eaef00', 'file_type': 'training', 'url': 'https://api.neuralmagic.com/v2/models/41dc6dcb-88cd-45f4-ae5e-8c34c0eaef00/files/config.json'}`~~
~~**after(V2):** {'display_name': 'deployment/config.json', 'file_size': 799, 'model_id': '41dc6dcb-88cd-45f4-ae5e-8c34c0eaef00', 'file_type': 'deployment', 'url': 'https://api.neuralmagic.com/v2/models/41dc6dcb-88cd-45f4-ae5e-8c34c0eaef00/files/deployment/config.json'}~~
~~References to config.json, tokenizer.json and tokenizer_config.json were updated to reflect this change.~~ This is now handled by a change to SparseZoo instead: https://github.com/neuralmagic/sparsezoo/commit/9d9b2ba450b3f26ad79a22acf4c09255f41eefc0

~~NOTE: A bit unsure about hardcoding the deployment folder name, but since we are already searching for that particular directory name in the sparsezoo repo(see `ALLOWED_FILE_TYPES`) I think it should be OK to do here too. **However, this does remove compatibility with V1**~~

The remaining errors are caused by test cases referencing sparsezoo stubs that are deprecated. George is working on replacing these stubs in a separate PR: [https://github.com/neuralmagic/deepsparse/pull/1234](https://github.com/neuralmagic/deepsparse/pull/1234)

### DIFF
--- a/tests/deepsparse/image_classification/test_pipelines.py
+++ b/tests/deepsparse/image_classification/test_pipelines.py
@@ -65,7 +65,7 @@ def test_image_classification_pipeline_preprocessing(
         data_originals_path = zoo_model.sample_inputs.path
     for idx, sample in enumerate(load_numpy_list(data_originals_path)):
         image_raw = list(sample.values())[0].astype(numpy.uint8)
-        image_raw = numpy.transpose(image_raw, axes=[1,2,0])
+        image_raw = numpy.transpose(image_raw, axes=[1, 2, 0])
         image_raw = Image.fromarray(image_raw)
 
         preprocessed_image_pipeline = ic_pipeline.process_inputs(
--- a/tests/deepsparse/image_classification/test_pipelines.py
+++ b/tests/deepsparse/image_classification/test_pipelines.py
@@ -59,12 +59,13 @@ def test_image_classification_pipeline_preprocessing(
     ic_pipeline = Pipeline.create("image_classification", model_path=zoo_stub)
     zoo_model = Model(zoo_stub)
     data_originals_path = None
-    if zoo_model.sample_originals is not None:
-        if not zoo_model.sample_originals.files:
-            zoo_model.sample_originals.unzip()
-        data_originals_path = zoo_model.sample_originals.path
+    if zoo_model.sample_inputs is not None:
+        if not zoo_model.sample_inputs.files:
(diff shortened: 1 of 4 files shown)


horheynm (approved):

bfineran (approved):

dbogunowicz:
Feeling a little uneasy about the backward compatibility here. @bfineran should we care about it?

tests/deepsparse/image_classification/test_pipelines.py
```diff
@@ -59,12 +59,13 @@ def test_image_classification_pipeline_preprocessing(
     ic_pipeline = Pipeline.create("image_classification", model_path=zoo_stub)
     zoo_model = Model(zoo_stub)
     data_originals_path = None
-    if zoo_model.sample_originals is not None:
-        if not zoo_model.sample_originals.files:
-            zoo_model.sample_originals.unzip()
-        data_originals_path = zoo_model.sample_originals.path
+    if zoo_model.sample_inputs is not None:
```
rahul-tuli, line 62: Do we not have sample_originals anymore?

horheynm: Some models do not have originals.

horheynm (approved):

horheynm:
> Feeling a little uneasy about the backward compatibility here. @bfineran should we care about it?

sample-originals were uploaded, deleted from the source that triggers the deployment to save to db and blob storage, but it does not trigger deletion. .
sample-input is more reliable, if the tests are not hardcoded to work with originals.

bfineran (approved):

PR #1237, opened 2023-09-11
[server] Update server routes to be compliant with MLServer

dsikka:
### Summary:
- Includes two other PRs: #1300 and #1288 
- Updates the base routes such that they follow the mlserver convention

### Testing
- Updated all the server tests 
- Tested locally using custom routes as well as routes built for the user given the model name

Sample Config:

```
num_cores: 2
num_workers: 2
endpoints:
  - task: question_answering
    route: some_route/pruned
```

This will now produce the following endpoints:
Screenshot 2023-10-10 at 11 59 39 AM


src/deepsparse/server/server.py
```diff
@@ -237,16 +249,76 @@ def _add_endpoint(
...
+            endpoint_config.route = clean_up_route(endpoint_config.route)
+            route_ready = f"{endpoint_config.route}/ready"
+            route_meta = endpoint_config.route
+        else:
+            route_ready = f"/v2/models/{endpoint_config.name}/ready"
+            route_meta = f"/v2/models/{endpoint_config.name}"
+
+    elif integration == INTEGRATION_SAGEMAKER:
+        route_ready = "/invocations/ready"
+        route_meta = "/invocations"
```
Satrat, line 310: This might be overcomplicating things, but I'm thinking it could make sense here to break this out into a class, especially if we plan on supporting more integrations in the future. Ie a base route class and then subclasses for each integration.

src/deepsparse/server/server.py
```diff
@@ -275,26 +347,24 @@ def _predict_from_files(request: List[UploadFile]):
...
+        )
+        route = clean_up_route(route)
 
         routes_and_fns.append((route, _predict))
         if hasattr(input_schema, "from_files"):
             routes_and_fns.append((route + "/from_files", _predict_from_files))
+
     elif integration == INTEGRATION_SAGEMAKER:
-        route = "/invocations"
+        route = "/invocations/infer"
```
Satrat, line 362: Same thought here about having a base class with children for each integration, to clean up the if statements

src/deepsparse/server/server.py
dsikka: Yah I agree. Was planning on doing something similar as part of OpenAI integration. This is just updating routes for existing integrations

src/deepsparse/server/server.py
dsikka: same comment as above

src/deepsparse/server/server.py
```diff
@@ -142,12 +152,14 @@ def _home():
     def _info():
         return server_config
 
-    @app.get("/ping", tags=["general"], response_model=bool)
-    @app.get("/health", tags=["general"], response_model=bool)
-    @app.get("/healthcheck", tags=["general"], response_model=bool)
-    @app.get("/status", tags=["general"], response_model=bool)
```
bfineran, line 148: I forget which, but at least one is required for sagemaker (I think it's ping)

bfineran (commented):
LGTM overall - as discussed offline, agree with both @dsikka and @Satrat it's time we move route creation to a separate file / fleshed out system

dbogunowicz (dismissed):
LGTM, clean

bfineran (dismissed):
LGTM - as mentioned before we need to sync w/ QA before landing

mgoin (commented):
This seems like it would cause breaking changes to any application built using the old endpoint structure. I think if you could make a README or document showing how to transition current examples of usage, that would be helpful for other teams and users to have a summary of the changes.
For instance, how should we update this Digital Ocean getting started guide? https://marketplace.digitalocean.com/apps/deepsparse-inference-runtime

src/deepsparse/server/openai_server.py
```diff
@@ -0,0 +1,365 @@
...
+from fastapi import BackgroundTasks, FastAPI, Request
+from fastapi.responses import StreamingResponse
+
+
+_LOGGER = logging.getLogger(__name__)
+
+OPENAI_CHAT_NOT_SUPPORTED = ["logit_bias", "best_ok", "ignore_eos", "use_beam_search"]
+OPENAI_TO_DEEPSPARSE_MAPPINGS = {
+    "max_tokens": "max_length",
+    "frequency_penalty": "repetition_penalty",
```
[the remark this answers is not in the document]
dsikka: we also following hugging face convention in our pipeline.

bfineran (dismissed):

Satrat (dismissed):

Satrat (approved):

bfineran (approved):

PR #1240, opened 2023-09-14
[TextGeneration] Add `GeneratedText` and update `TextGenerationOutput`

dsikka:
## Summary
- Add `GeneratedText` schema to the `TextGeneration` pipeline
- Update the `TextGenerationOutput` to use the `GeneratedText` class for each output sequence
- Use the `FinishReason` enum for the logged reason for generation to complete. Question - missing reason for any generic callback?
- Following internal docs: https://github.com/neuralmagic/internal-docs/blob/0bd3bf122c1f8c5871b085916d04cb0266a0a332/requirements/deepsparse/natural-language-generation/main.md
- One missing piece in in the `GeneratedText` is the use of alias

- Also, we're keeping the `prompt` in the `TextGenerationOutput` as opposed to the corresponding `GeneratedText` which seems counterintuitive?

## Testing
- Tested locally - sample output:

```

pipeline = Pipeline.create(
   task="text_generation",
   model_path="/home/dsikka/.cache/sparsezoo/neuralmagic/opt-1.3b-opt_pretrain-quantW8A8/deployment",
   engine_type="onnxruntime"
)

inference = pipeline(sequences=["what", "hello?"], num_generated_predictions=2)
generations = inference.generations
print("First Generation")
print(generations[0])
print("\n")
print("Second Generation")
print(generations[1])
```

**First Generation**
[GeneratedText(text=' color is the willow fern?\nI got it from Panera years ago. Will do a full review of this Matt veg rose', score=None, finished=True, finished_reason='stop token generated.'), GeneratedText(text=" are your plans for your website?\nI was in the process of completing that actually. I mean, I'm probs not going to add more features like a hub video or anything but as far as making the website look prettier it's a work in progress.  Anyways, by the look of the username I'm guessing you have some contacts in the film world? Want to work on something together?\nNot sure. LuckyIknew.com isn't exactly sparkling when I tried to navigate through it just now. Uhm... When are you going to put together all the footage and do it so the audio matches up with the footage?\nI'll do that now. I've had to go to my basement for a few weeks though, all the footage reels are in there. Will happen as soon as I can get back in there and redo it properly. Short of a whole bunch of stuff happening, I won't be able to put too much effort into it, sadly.", score=None, finished=True, finished_reason='stop token generated.')]

**Second Generation**
[GeneratedText(text=' houdini dragon edition? Oh every1 is spelling honey a weird way.... THE HELL DO YOU MEAN?\nMy word is brown?', score=None, finished=True, finished_reason='stop token generated.'), GeneratedText(text="!! hello?! where's her obsession with Yugoslavs?\nthe Yugoslavian attitudes were everyone stands still, only steps on the yard is path, slowly walks to the other side returns, dont, coulor-moning... all with a mix of 'joking' and half serious and i think thats why i like it", score=None, finished=True, finished_reason='stop token generated.')]

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import datetime
 import logging
 import os
 import warnings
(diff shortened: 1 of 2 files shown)


bfineran (dismissed):
LGTM overall - let's check any downstream usages or tests of the pipeline that changing the output like this may break

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -51,6 +53,12 @@
 __all__ = ["TextGenerationPipeline"]
 
 
+class FinishReason(Enum):
+    STOP = "stop token generated."
+    LENGTH = "max token generation reached."
+    TIME = "max time reached."
```
mgoin, line 59: I would advocate for making these more terse i.e. simply `stop`, `length`, `time` or matching the OpenAI Completions API https://platform.openai.com/docs/guides/gpt/chat-completions-response-format

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -122,15 +130,34 @@ class Config:
...
+    )
+    score: Optional[Any] = Field(
+        description="The score for the generated token or sequence. "
+        "The scores have the shape [sequence_length, vocab_size]"
+    )
+    finished: bool = Field(description="Whether generation has stopped.")
+    finished_reason: str = Field(
+        description="The reason for generation to stop. "
+        "Defined by FinishReason. One of stop, length, or time."
+    )
```
mgoin, line 146: Tempted to think that `finished_reason` could just be an `Optional[str]` so we know it is unfinished if `None`, but happy either way

dsikka: Yup - there’s a draft PR to enable streaming where I update this to be optional

dbogunowicz:
LGTM, let's add some testing and make sure that it doesn't break things downsteam (perplexity computation).

dbogunowicz (dismissed):

bfineran (approved):

dbogunowicz (approved):
Nice

PR #1241, opened 2023-09-14
[Text Generation][Fix] Make warning more precise when prompt_sequence_length == 1

dbogunowicz:
Fix the ambiguous warning, that occurs when we run the pipeline with `prompt_sequence_lenght=1`.

e.g.
```

pipeline = Pipeline.create(
   task="text_generation",
   model_path="zoo:nlg/text_generation/opt-1.3b/pytorch/huggingface/opt_pretrain/pruned50_quantW8A8-none",
   prompt_sequence_length=1
)

inference = pipeline(sequences=["hello?"], return_logits=True)
```

**Before:**
you'd get a warning:

"This ONNX graph does not support processing the prompt in "
"with processing length > 1. Creation of an auxiliary engine for "
"Creation of an auxiliary engine for "
"processing the prompt at a larger processing length is disabled. "
"The prompt will be processed in with processing length 1."

...which is clearly not true, there is nothing wrong with the ONNX graph!

**Now:**
"Creation of an auxiliary engine for "
"processing the prompt at a larger processing length is disabled. "
"The prompt will be processed in with processing length 1." -> you'd see this warning

and this warning

"This ONNX graph does not support processing the prompt in "
"with processing length > 1. 

...will only pop up only if your ONNX model really does not support processing length > 1


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -199,6 +199,7 @@ def __init__(
         # enable multitoken prefill if
         # - the model graph is supporting it (causal_mask input is present)
         # - prompt_sequence_length != 1 (identical to single-token prefill)
+
```
dbogunowicz, line 202:
```suggestion
```

mgoin (approved):
Thank you very much for the clarification!

PR #1244, opened 2023-09-14
[Fix] Temporary suppress transformers logging

dbogunowicz:
#1232  introduced the new pathway for instantiating transformer configs. The new pathway emits a warning stdout that gets printed on every instantiation of the pipeline. 

```bash
 f"You are using a model of type {config_dict['model_type']} to instantiate a model of type "
                f"{cls.model_type}. This is not supported for all configurations of models and can yield errors."
```

This print is not only potentially confusing for the user, but also breaks our tests (as some tests explicitly look for a string "error" in the stdout).

This PR introduces a lightweight context manager, that enables temporary reduction/increase of the config level for the particular module logger.

### DIFF
--- a/src/deepsparse/transformers/helpers.py
+++ b/src/deepsparse/transformers/helpers.py
@@ -17,7 +17,6 @@ Helper functions for working with ONNX exports of transformer models and deepspa
 """
 
 
-import logging
 import os
 import re
 import warnings
@@ -42,7 +41,6 @@ __all__ = [
     "fix_numpy_types",
     "get_transformer_layer_init_names",
     "truncate_transformer_onnx_model",
-    "TemporaryLoggingLevel",
 ]
 
 _LOGGER = get_main_logger()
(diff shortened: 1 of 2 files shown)


mgoin (approved):

bfineran (commented):
@dbogunowicz were we not able to suppress with `warnings.filterwarning`? https://docs.python.org/3/library/warnings.html#warnings.filterwarnings

dbogunowicz:
@bfineran yeah, did not work for me. Did not spend too much time investigating though.

Satrat (approved):

bfineran (approved):

dbarbuzzi:
I’ve done some digging and have a solution to target and filter this specific warning. `warnings.filterwarnings` will not work because the transformers code is using `logging.warn` which is separate from `warnings`. Instead, we can create a class to filter the logs and attach it to the transformer's logger's `StreamHandler` (attaching directly to the logger won’t work). The overall code looks something like this:

```python
# the filter class
class TransformersLogFilter(logging.Filter):
    def filter(self, record):
        # return 1 to log, return 0 to ignore/suppress
        return not record.getMessage().startswith("You are using a model of type")

# applying the filter class to the handler by replacing the existing streamhandler
# with the modified version
l = logging.getLogger("transformers")
for handler in l.handlers:
    if isinstance(handler, logging.StreamHandler):
        l.removeHandler(handler)
        handler.addFilter(TransformersLogFilter())
        l.addHandler(handler)
```

PR #1252, opened 2023-09-18
[Test] Make test suite independent from torchvision

dbogunowicz:
Fixes one of the irritating properties of our deepsparse tests, i.e. it was implicitly assumed that one has `torchvision` installed when running the test suite.

 I know that this is not a problem for QA, as they always install all the dependencies before running their tests. However, when testing locally, it was something that started to drive me mad :) If I am testing transformer models, why do I need to have torchvision installed?!

Before:
`conftest.py` on initialization imported `find_file_with_pattern`. This function could only be correctly imported if torchvision was installed.
Now:
`find_file_with_pattern` is uncoupled from `torchvision` dependency.

### DIFF
--- a/src/deepsparse/benchmark/benchmark_model.py
+++ b/src/deepsparse/benchmark/benchmark_model.py
@@ -17,11 +17,11 @@ Benchmarking script for ONNX models with the DeepSparse engine.
 
 ##########
 Command help:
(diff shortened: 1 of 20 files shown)


rahul-tuli (approved):

rahul-tuli:
Nice diff!

bfineran (approved):

PR #1257, opened 2023-09-19
[TextGeneration] Update `sampling_temperature` and `deterministic` to be part of the input

horheynm:
## Summary:
- This removes the `temperature` and `sampling_temperature` from the constructor to the inputs. Specifically, we're mapping `temperature` to the `sampling_temperature` config attribute and `deterministic` to `do_sample`
- Note: `do_sample` being True means `deterministic` is False

```
pipeline = Pipeline.create(
   task="text_generation",
   model_path="/home/dsikka/.cache/sparsezoo/neuralmagic/opt-1.3b-opt_pretrain-quantW8A8/deployment",
   engine_type="onnxruntime",
)
config = {
   "num_return_sequences": 3,
   "do_sample": True
}
inference = pipeline(prompt=["how are you?"], generation_config=config)
print(inference)
```

## Testing:
- All tests passing locally and updated to add two new tests with the new attributes

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -136,7 +136,7 @@ class TextGenerationInput(BaseModel):
         description="GenerationConfig file consisting of parameters used to control "
         "sequences generated for each prompt. The current supported parameters are: "
         "max_length, max_new_tokens, num_return_sequences, output_scores, top_p, "
-        "top_k, repetition_penalty.",
+        "top_k, repetition_penalty, do_sample",
     )
 
     kwargs: Optional[Dict] = Field(
@@ -190,10 +190,6 @@ class TextGenerationPipeline(TransformersPipeline):
     """
     Pipeline for text generation tasks.
 
-    :param deterministic: if False, the pipeline will sample from
-        the probability distribution computed from the logits.
-        If True, the pipeline will get the next token by applying
-        an argmax function to the logits.
     :param sampling_temperature: the temperature to use when sampling
         from the probability distribution computed from the logits.
         Higher values will result in more random samples. Should
@@ -212,7 +208,6 @@ class TextGenerationPipeline(TransformersPipeline):
 
     def __init__(
         self,
-        deterministic: bool = True,
         sampling_temperature: float = 1.0,
         prompt_sequence_length: int = 64,
         sequence_length: int = 1024,
@@ -253,7 +248,6 @@ class TextGenerationPipeline(TransformersPipeline):
             if "WAND_OPT_FLAGS" not in os.environ:
                 os.environ["WAND_OPT_FLAGS"] = "default,~pyramids"
 
-        self.deterministic = deterministic
         self.sampling_temperature = sampling_temperature
         self.prompt_sequence_length = prompt_sequence_length
         self.force_max_tokens = force_max_tokens
@@ -452,8 +446,7 @@ class TextGenerationPipeline(TransformersPipeline):
             )
 
         # If the num_return_sequences > 1, repeat the prompt
-        # num_return_sequences times. Also, update the engine so that deterministic
-        # is set to False.
+        # num_return_sequences times.
         original_inputs = inputs.sequences
         if generation_config.num_return_sequences > 1:
             if isinstance(inputs.sequences, str):
@@ -461,7 +454,6 @@ class TextGenerationPipeline(TransformersPipeline):
             inputs.sequences = repeat_inputs(
                 inputs.sequences, generation_config.num_return_sequences
             )
-            self.deterministic = False
 
         if inputs.fixed_sequences_length or not self.cache_support_enabled:
             # to enforce a fixed sequence length, we need to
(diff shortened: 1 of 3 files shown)


rahul-tuli (commented):
Nice diff!
Should we also add a test to check sampling_temperature is accepted during inference? and that it actually affects the something in output?

dbogunowicz (dismissed):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -435,10 +431,6 @@ def process_inputs(self, inputs: TextGenerationInput) -> List[numpy.ndarray]:
             inputs.sequences = repeat_inputs(
                 inputs.sequences, inputs.num_generated_predictions
             )
-            if self.engine:
-                self.engine.deterministic = False
-            if self.multitoken_engine:
-                self.multitoken_engine.deterministic = False
```
bfineran, line 441: we should replace this by updating the `TokenGenerator` args right?

bfineran (approved):

bfineran (changes requested):
meant to req changes

horheynm:
> Nice diff! Should we also add a test to check sampling_temperature is accepted during inference? and that it actually affects the something in output?

That is a good idea. Ill make diff for that, this pr is meant for fixing deterministic vs non-d fixes

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -425,6 +425,39 @@ def test_num_generated_predictions(self, setup):
...
+        pipeline = self.get_pipeline(**pipeline_kwargs)
+        num_generated_predictions = 3
+        inference = pipeline(
+            sequences=["hello?"],
+            num_generated_predictions=num_generated_predictions,
+            max_tokens=100,
+        )
+        sequences = inference.sequences[0]
+        # Outputs should be unique from one another
+        assert len(set(sequences)) == num_generated_predictions
```
Satrat, line 459: Could we add an assert to confirm they differ?

dbogunowicz:
@horheynm can we land this asap? I think this is part of the release deliveries.

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: Where do we set deterministic to be False for this case then? @horheynm

dsikka (commented):
We're adding `do_sample/deterministic` to the GenerationConfig which will conflict with this so would like to get this in sooner than later, once tested with the `test_text_generation.py` tests

tests/deepsparse/transformers/pipelines/test_text_generation.py
horheynm: I think if there are duplicated exact sequences the len(set(seqs)) of those will be 1, so this test should confirm that they differ

horheynm:
There are some test failures in `test_text_generation.py`, will need to debug

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -438,10 +438,7 @@ def process_inputs(
             inputs.sequences = repeat_inputs(
                 inputs.sequences, generation_config.num_return_sequences
             )
-            if self.engine:
-                self.engine.deterministic = False
-            if self.multitoken_engine:
-                self.multitoken_engine.deterministic = False
+            self.deterministic = False
```
bfineran, line 441: @dsikka noting that we cannot change pipeline state (persistent) due to a parameter in a single inference (temporary), should only set at the last minute when passing to token generator based on inputs

bfineran (approved):

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -435,6 +435,45 @@ def test_num_generated_predictions(self, setup):
...
+            top_k=0,
+            top_p=0.0,
+            num_return_sequences=3,
+            do_sample=True,
+        )
+        inference = pipeline(sequences=["hello?"], generation_config=config)
+        generations = inference.generations
+        # Output should be the same from one another
+        text_outputs = [x.text for x in generations[0]]
+        assert len(set(text_outputs)) == 3
```
horheynm, line 475: non-det, so different output, so should be unique from one another?

dsikka: yes

horheynm: whoops you are correct

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -659,12 +644,14 @@ def engine_forward(
             finished_reason = []
             streaming = context.get("streaming")
             generation_config = context.get("generation_config")
+            deterministic = not generation_config.do_sample
```
rahul-tuli, line 647: Will generation config always be present in context? also could we raise an exception if generation_config is not an instance of GenerationConfig? or is that check already done somewhere else?

rahul-tuli (approved):

src/deepsparse/transformers/pipelines/text_generation.py
dsikka: generation config will always be present, because if nothing, we'll provide our defaults and all the checks are done as part of the processing

PR #1259, opened 2023-09-20
benchmarker API

horheynm:
Trigger
deepsparse.benchmark
deepsparse.benchmark_pipeline

Example code in tests

### DIFF
--- a/src/deepsparse/benchmark/api/benchmarker.py
+++ b/src/deepsparse/benchmark/api/benchmarker.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from typing import Optional
+
+from deepsparse.benchmark.benchmark_model import benchmark_model
+from deepsparse.benchmark.benchmark_pipeline import benchmark_pipeline
+
+
+class Benchmarker:
+    def __init__(self, model, pipeline, url):
+        self.model = model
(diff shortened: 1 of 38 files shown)


src/deepsparse/benchmark/api/benchmarker.py
```diff
@@ -0,0 +1,57 @@
...
+
+from deepsparse.benchmark.benchmark_model import benchmark_model
+from deepsparse.benchmark.benchmark_pipeline import benchmark_pipeline
+
+
+class Benchmarker:
+    def __init__(self, model, pipeline, url):
+        self.model = model
+        self.pipeline = pipeline
+        self.url = url
```
bfineran, line 26: let's enforce that only one can be set, no need to add URL yet

bfineran (commented):
Looks good overall - need a way to instantiate model and pipeline from other sources (ie files, kwargs, etc)

src/deepsparse/benchmark/api/benchmarker.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Input arg to `model`, `pipeline` should be one of:
+     - SparseZoo stub
+     - path to a model.onnx
+     - path to a local folder containing a model.onnx
+     - path to onnx.ModelProto
+
+    Provide the stub/path to one of
+     - model to run deesparse.benchmark
+     - pipeline to run deepsparse.benchmark_pipeline
```
bfineran, line 35:
```suggestion
    Provide the stub/path to one of
     - onnx model to run deesparse.benchmark
     - deployment directory to run deepsparse deepsparse.benchmark_pipeline
```

src/deepsparse/benchmark/api/benchmarker.py
```diff
@@ -0,0 +1,78 @@
...
+        self.pipeline = pipeline
+
+    def __call__(self, **kwargs):
+        if self.model:
+            return benchmark_model(model_path=self.model, **kwargs)
+
+        if self.pipeline:
+            return benchmark_pipeline(model_path=self.pipeline, **kwargs)
+
+    @staticmethod
```
bfineran, line 54: can we add a top level `benchmark` function as well so users can just import that if they want to (instead of `Benchmarker.benchmark`

bfineran (commented):
LGTM overall - the only thing I think would be missing is a way to pass an already instantiated pipeline object to benchmarker

something like:

```python
my_pipeline = Pipeline.create(...)

benchmarker = Benchmarker(pipeline=my_pipeline)
```

bfineran (approved):

PR #1262, opened 2023-09-21
explicitly set ORT providers on all InferenceSession creations

bfineran:
latest version of ORT has a hard requirement for providers to be set. this PR updates any inference session that does not have it set 

**test_plan:**
QA Team + GHA


mgoin (approved):

mgoin:
Landing as this is clearly a positive step towards correct GHA

PR #1273, opened 2023-09-25
Remove Dead Code

rahul-tuli:
In one of our recent changes to `TextGenerationPipeline`,  a function within a function was moved outside, but the original function wasn't deleted. This lead to a duplicate "dead" function not used anywhere.

This PR removes that dead function `_create_generated_text_output`

Note: there is already a `_create_generated_text_output` function defined at the class level


dbogunowicz (approved):

dsikka (approved):
good catch!

bfineran (approved):

PR #1277, opened 2023-09-25
Use nullcontext in nl_decoder_engine when timer_manager.current is None

mgoin:
There is an error that seems to only happen when generating the second token when streaming=True, but only shows up in specific pipeline setups such as Gradio

Error:
```
  File "/home/user/.local/lib/python3.10/site-packages/deepsparse/transformers/engines/nl_decoder_engine.py", line 194, in __call__
    with timer.time(f"EXECUTE_ENGINE_SEQ_LEN_{self.sequence_length}"):
AttributeError: 'NoneType' object has no attribute 'time'
```

This PR simply uses nullcontext to replace the timer if it is None

### DIFF
--- a/src/deepsparse/transformers/engines/nl_decoder_engine.py
+++ b/src/deepsparse/transformers/engines/nl_decoder_engine.py
@@ -14,6 +14,7 @@
 import copy
 import logging
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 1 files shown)


mgoin:
Resolved by https://github.com/neuralmagic/deepsparse/pull/1294

PR #1286, opened 2023-09-26
[TextGeneration] Fix max_length = 1, repeat token

dsikka:
Summary:

This fixes the issue where if the user has set max_length=1, finish reason is never populated as we never go through the loop in engine_forward. max_length=1 is needed for our perplexity calculation so this PR will unblock that from being completed.

Testing:

Tested locally

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -746,6 +746,7 @@ class TextGenerationPipeline(TransformersPipeline):
 
                     if len(generated_tokens) == max_tokens:
                         finished_reason.append(FinishReason.LENGTH)
+                        break
 
                     if streaming:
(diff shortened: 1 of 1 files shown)


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -756,13 +757,24 @@ def engine_forward(
                     tokens=token_generator.tokens, kv_cache=session
                 )
                 if streaming:
-                    yield (
-                        numpy.array([token]),
-                        numpy.array([logits]),
-                        [finished_reason[-1]],
-                    )
+                    if max_tokens == 1:
+                        yield (
```
bfineran, line 761: can you add an in-line comment here

bfineran (approved):
LGTM

rahul-tuli (approved):

rahul-tuli (approved):

bfineran (approved):

PR #1289, opened 2023-09-29
Set text-generation default values to match performance and UX needs

mgoin:
Proposal for new defaults for the text-generation pipeline

Includes:
- Reducing `prompt_sequence_length` to 16. This is in line with benchmarking results showing the performance benefits of going larger than 16 is not great enough to offset the padding overhead. 
- Reducing the default number of generated tokens to 200 rather than 1k - we might even want to make this smaller like 50 to start with.
- Due to the above, I propose we make `max_length` None by default since it conflicts with `sequence_length`. Both of these are trying to control the maximum allowed length of `prompt + max_new_tokens` and by default we should be limited by `sequence_length` since we cannot run past that

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation.py
@@ -61,16 +61,21 @@ _LOGGER = logging.getLogger(__name__)
 __all__ = ["TextGenerationPipeline"]
 
 
+# Based off of https://huggingface.co/docs/transformers/main_classes/text_generation#transformers.GenerationConfig # noqa E501
 class GenerationDefaults:
-    num_return_sequences = 1
-    max_length = 1024
-    max_new_tokens = None
-    output_scores = False
-    top_k = 0
-    top_p = 0.0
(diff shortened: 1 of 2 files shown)


bfineran (approved):

src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -61,16 +61,21 @@
...
-    top_p = 0.0
-    repetition_penalty = 0.0
+    # Parameters that control the length of the output
+    max_length = None
+    max_new_tokens = 200
+    # Parameters that control the generation strategy used
     do_sample = False
+    # Parameters for manipulation of the model output logits
     temperature = 1.0
+    top_k = 50
```
dbogunowicz, line 73: Just a heads up, I did not look into it in much detail, but changing top_k or/and top_p from zeros to HF defaults results in pretty wonky generated sequences, that deviate massively from sequences generated by HF torch model. Not sure why, maybe we are having some bug internally with top_k/top_p.

PR #1294, opened 2023-10-03
[TextGeneration][Bug] Fix timer

dsikka:
For this ticket: https://app.asana.com/0/1205229323407165/1205618233879512/f

Tested locally with gradio/non-gradio examples

### DIFF
--- a/src/deepsparse/transformers/engines/nl_decoder_engine.py
+++ b/src/deepsparse/transformers/engines/nl_decoder_engine.py
@@ -185,14 +185,7 @@ class NLDecoderEngine:
 
         :return: The generated token and corresponding logits
         """
-
-        timer = self.timer_manager.current
-        if not timer:
-            with self.timer_manager.new_timer_context(
-                total_inference=False
-            ) as new_timer:
-                timer = new_timer
-
+        timer = self.timer_manager.current_or_new()
         if kv_cache:
             # if model has kv cache enabled, we need
             # to add the kv cache state to the input
--- a/src/deepsparse/transformers/engines/nl_decoder_engine.py
+++ b/src/deepsparse/transformers/engines/nl_decoder_engine.py
@@ -185,7 +185,14 @@ class NLDecoderEngine:
 
         :return: The generated token and corresponding logits
         """
+
         timer = self.timer_manager.current
+        if not timer:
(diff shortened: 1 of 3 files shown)


src/deepsparse/transformers/engines/nl_decoder_engine.py
```diff
@@ -185,7 +185,14 @@ def __call__(
 
         :return: The generated token and corresponding logits
         """
+
         timer = self.timer_manager.current
+        if not timer:
```
bfineran, line 190: briefly mentioned in the code review meeting but let's implement `timer_manager.current_or_new()` so we can splash this pattern where needed

bfineran (changes requested):

src/deepsparse/transformers/engines/nl_decoder_engine.py
dsikka: ah gotcha. I was under the impression that we still had a `timer_manager.new_timer()` method (mentioned in the docstring) and that was what you were referring to. But yes, updating.

src/deepsparse/utils/timer.py
```diff
@@ -338,6 +338,16 @@ def all_times(self) -> Dict[str, List[float]]:
...
 
+    def current_or_new(self) -> StagedTimer:
+        """
+        Return the current timer if there is one, otherwise return a new one.
+        """
+        if self.current:
+            return self.current
+        else:
+            with self.new_timer_context(total_inference=False) as timer:
+                return timer
```
bfineran, line 349: is this tested? just want to make sure it won't exit the context and leave with current not set (even if the timer object still exists)

mgoin (approved):
This works with my test, thanks!
[test.py](https://github.com/neuralmagic/deepsparse/files/12797468/test.txt)

bfineran (approved):

src/deepsparse/utils/timer.py
bfineran: Right, I'm sure it will work, only question I had is if the current that is created will actually live beyond the context in this function

mgoin: Good call, this is worth investigating in a followup

PR #1247, opened 2023-09-15
Pipelines to use deployment dir path and not single files.

dbogunowicz:

PR #1296, opened 2023-10-03
[bugfix] model_to_path should also download .data files from zoo

bfineran:
fixes issue in #1247 where only `model.onnx` would be downloaded by default in `model_to_path` from a deployment directory. this breaks flows such as `benchmark` and `compile_model` which rely on only the `model_to_path` for the onnx file path from zoo.  with this change the function now also searches for and downloads `.data*` files

**test_plan:**
verified locally that data file downloads now trigger with a stub such as `zoo:mpt-7b-dolly_mpt_pretrain-base` where they weren't before


horheynm (approved):
This was for 
`failed:Invalid fd was supplied: -1`?

PR #1300, opened 2023-10-06
[server] Update OpenAI Model Support

dsikka:
## Summary
- This PR is focused on updating the openai integration such that it is more independent of the deepsparse server, leveraging the server refactor
- This PR allows multiple text generation models to hosted using the single `/v1/chat/completions` endpoint, adds the `/v1/models` endpoint, updates the base routes, and also adds a separate openai specific workflow command

## Examples and testing
- On the command line, the following command can be used:
```
deepsparse.openai sample_config.yaml 
```

Here, `sample_config.yaml` has the following structure:

```
num_cores: 2
num_workers: 2
endpoints:
  - task: text_generation
    model: zoo:nlg/text_generation/opt-1.3b/pytorch/huggingface/opt_pretrain/pruned50_quantW8A8-none
```

- This launches a FASTAPI app with the following endpoints:
Screenshot 2023-10-06 at 12 51 05 PM

## Use the `openai` api to send requests:

```python
import openai


openai.api_key = "EMPTY"
openai.api_base = "http://localhost:5543/v1"

# Completion API
stream = False
completion = openai.ChatCompletion.create(
    messages="how are you?",
    stream=stream,
    max_tokens=30,
    model="zoo:nlg/text_generation/opt-1.3b/pytorch/huggingface/opt_pretrain/pruned50_quantW8A8-none",
)

print("Chat results:")
if stream:
    text = ""
    for c in completion:
        print(c)
else:
    print(completion)

```

- The user can provide the model they want to run inference on as part of the payload. If that model was not provided when launching the server, an error will be returned. We will be adding functionality to allow the user to add additional models, after the server has already been launched.

### DIFF
--- a/src/deepsparse/server/openai_server.py
+++ b/src/deepsparse/server/openai_server.py
@@ -36,7 +36,7 @@ from deepsparse.server.protocol import (
     UsageInfo,
     random_uuid,
 )
-from deepsparse.server.server import Server
+from deepsparse.server.server import Server, SystemLoggingMiddleware
 from fastapi import BackgroundTasks, FastAPI, Request
 from fastapi.responses import StreamingResponse
 
@@ -67,16 +67,14 @@ class OpenAIServer(Server):
 
         app.model_list = self.model_list
         app.model_to_pipeline = self.model_to_pipeline
-        app.context = self.context
-        app.executor = self.executor
 
-        @app.get("/v2/models", tags=["model"])
+        @app.get("/v1/models", tags=["model"])
         async def show_available_models():
             """Show available models. Right now we only have one model."""
             return app.model_list
 
         @app.post(
-            "/v2/models/chat/completions",
+            "/v1/chat/completions",
             tags=["model", "inference"],
             response_model=ChatCompletionResponse,
         )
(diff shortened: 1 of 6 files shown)


bfineran (approved):

src/deepsparse/server/openai_server.py
```diff
@@ -53,17 +52,132 @@
...
                 endpoint_config,
             )
 
-        _LOGGER.info(f"Added endpoints: {[route.path for route in app.routes]}")
+        app.model_list = self.model_list
+        app.model_to_pipeline = self.model_to_pipeline
+
+        @app.get("/v1/models", tags=["model"])
+        async def show_available_models():
+            """Show available models. Right now we only have one model."""
```
mgoin, line 73: > Right now we only have one model.

Is this true if we can load multiple models with one config?

src/deepsparse/server/openai_server.py
```diff
@@ -53,17 +52,132 @@
...
+
+            prompt = request.messages
+            request_id = f"cmpl-{random_uuid()}"
+            created_time = int(time.time())
+            model = request.model
+
+            pipeline = app.model_to_pipeline.get(model)
+            if not pipeline:
+                create_error_response(
+                    HTTPStatus.BAD_REQUEST, f"{model} is not " "available"
```
mgoin, line 98:
```suggestion
                    HTTPStatus.BAD_REQUEST, f"{model} is not available"
```

mgoin (commented):
Really nice! Just two nits

mgoin (approved):

src/deepsparse/server/openai_server.py
dsikka: Good catch, updating.

PR #1308, opened 2023-10-10
[deepsparse.benchmark] enable benchmarking internal kv cache

bfineran:


mgoin (approved):
Good to go, tested by me on some sparse quantized models to match up with debug_analysis with internal kv cache - and drops performance without it enabled

PR #1324, opened 2023-10-17
[LLM Tests] Config support for LLM test suite / enable test suite in GHA

dbogunowicz:
Configurable testing suite for LLMs.

Introduces two types of tests:
1. Unit tests - test the smaller functionalities of the LLM pipelines (`text_generation` and `chat`). Those are small, self-contained lightweight tests that use the tinystories model. Those will run on every commit in GHA. It should be easy and simple to add new test cases here by any developer.

2. Integration test for LLMs - test end-to-end all the relevant LLM pipelines. Those are integration tests that are configurable and by default will not run on every commit but with some lower frequency. So far we have tests for codegen and opt models. I have run a limited scope of these tests on my local machine, so I do not guarantee that they will instantaneously all pass on the first go once deployed. Codegen has been thoroughly tested by me, but opt, because of its size, not.

Along with this feature come small fixes to the LLM pipelines that were required to make tests work properly.


tests/deepsparse/transformers/pipelines/configs/gpt_neo.yaml
```diff
@@ -0,0 +1,13 @@
+model_path: "hf:mgoin/TinyStories-1M-deepsparse"
```
dbogunowicz, line 1: Will add configs for codegen/opt/llama/etc after the successful review.

dbogunowicz:
Tests are green yay! :)

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -279,82 +294,55 @@ def test_deepsparse_single_token_prefill(self, setup):
...
-            torch_ground_truth=torch_ground_truth,
-            run_cache_validation=not self.internal_kv_cache,
+            torch_ground_truth=self.torch_ground_truth,
+            # disable kv cache validation if using internal kv cache
+            run_kv_cache_validation=not self.internal_kv_cache,
         )
 
     def test_deepsparse_generation_after_kv_cache_has_been_filled(self, setup):
-        # Test the pipeline that uses deepsparse engine. The test covers the
+        # Test the deepsparse that uses deepsparse engine. The test covers the
```
dbogunowicz, line 335:
```suggestion
        # Test the pipeline that uses deepsparse engine. The test covers the
```

tests/deepsparse/transformers/pipelines/test_chat.py
```diff
@@ -12,34 +12,157 @@
...
+    ):
+
+        tokenizer = pipeline.tokenizer
+        config = GenerationConfig(
+            output_scores=True, max_length=num_generated_tokens, top_k=0, top_p=0.0
+        )
+
+        # make sure that running two prompts one after another
+        # is identical to running a composition of those prompts
+        out_1_ = pipeline(
```
horheynm, line 130: i would name this out_1_session_1 and the other subsequent call out_2_session_1. The other independent one out_1_session_2

tests/deepsparse/transformers/pipelines/test_chat.py
```diff
@@ -12,34 +12,157 @@
...
+            generation_config=config,
+            include_prompt_logits=True,
+        )
+        cache_state_2 = pipeline.storage_kv_cache.get(session_id_2).cached_inputs[
+            "past_key_values.0.key"
+        ]
+        if cache_state_1.shape[0]:
+            # if cache state is not empty, i.e. we are managing kv cache
+            # externally, make sure that the cache state is the same
+            np.allclose(cache_state_1, cache_state_2, atol=self.precision)
```
horheynm, line 167: what is cache_state_n and what does it mean to be +/- self.precision?  If I make session1 and a duplicate of session 1 above, will the cache_state_n of those be the same?

The only difference if stateful between sess1 and sess2  is the number of times pipeline was called w the respective session_id right? is this causing the difference?

tests/deepsparse/transformers/pipelines/helpers.py
```diff
@@ -82,3 +87,53 @@ def _create_tokenizer(model_name):
...
+    expected_cadence = config["cadence"]
+
+    if not isinstance(expected_cadence, list):
+        expected_cadence = [expected_cadence]
+    if cadence in expected_cadence:
+        return config, None
+    return None, "Skipping test for cadence: {}".format(config["cadence"])
+
+
+def validate_cache_management_type(
```
horheynm, line 109: I would change this code to return True or False, not the value of internal_kv_cache. 
depending on the output, I would then skip pytest or continue the test and use internal_kv_cache.

Also where is this used?

tests/deepsparse/transformers/pipelines/helpers.py
```diff
@@ -82,3 +87,53 @@ def _create_tokenizer(model_name):
...
+    # parses the config file provided
+    assert os.path.isfile(config_path), f"config_path {config_path} is not a file"
+    # reads the yaml file
+    with open(config_path, "r") as f:
+        config = yaml.safe_load(f)
+
+    cadence = os.environ.get("CADENCE", "commit")
+    expected_cadence = config["cadence"]
+
+    if not isinstance(expected_cadence, list):
```
horheynm, line 102: What do you expect the types to be if its not a list? 
A good way to validate would be to use pydantic where you can do type checks, parse, and serialization

tests/deepsparse/transformers/pipelines/helpers.py
```diff
@@ -82,3 +87,53 @@ def _create_tokenizer(model_name):
...
+        )
+    if not internal_kv_cache and "external" not in cache_management_type:
+        pytest.skip(
+            "The tests for running the pipeline with "
+            "external kv cache management are disabled."
+        )
+    return internal_kv_cache
+
+
+def helper_test(test_method):
```
bfineran, line 125: what is a "helper" test? can we get a more intuitive name

tests/deepsparse/transformers/pipelines/helpers.py
dbogunowicz: This may be a bit of an overkill. Let's not forget that it is a simple dev-facing interface. If not a list, it should be a string. I am implicitly assuming this, maybe to implicitly.

tests/deepsparse/transformers/pipelines/helpers.py
dbogunowicz: it is used inside the `test_text_generation.py`. The cool thing is that once we hit `skipif`, we automatically skip the whole testing scenario. Else, we return True or False as you are suggesting.

Satrat (approved):
Nice work on this!

bfineran (approved):

PR #1325, opened 2023-10-17
[Pipeline Refactor] Additional functionality, engine operator, linear router and image classification pipeline/operators/example

dsikka:
### Summary
- Further flush out the Pipeline, Router, and Operator functionality
- Add an `EngineOperator` which can be used for majority of our pipelines
- Created a `LinearRouter`
- Use the base classes to create Operators for Image Classification, specifically for pre-processing and post-processing and tie the Operators together using a v2 ImageClassificationPipeline

### Testing
- Updated the initial test case 
- Test ImageClassificationPipeline creation and inference runs:

```python

from deepsparse.v2.image_classification.pipeline import ImageClassificationPipeline

image_path = "/home/dsikka/deepsparse/src/deepsparse/v2/image_classification/buddy.jpeg"

model_path = (
    "zoo:cv/classification/resnet_v1-50/pytorch/sparseml/imagenet/pruned95-none"
)

# Pipeline creation
pipeline = ImageClassificationPipeline(model_path=model_path)

# run using kwargs; will apply schema if defined
inference = pipeline(images=[image_path, image_path])
print(inference)

# run using arg and schema

from deepsparse.v2.image_classification.preprocess_operator import ImageClassificationInput
inference = pipeline(ImageClassificationInput(images=[image_path, image_path]))

```


src/deepsparse/v2/pipeline.py
```diff
@@ -13,88 +13,113 @@
...
 
-class Pipeline(BaseModel):
-    """
-    Pipeline accepts a series of operators, schedulers, and a router which define
-    an end to end ML transformation.
-
-    Calling a pipeline runs these transformations
+class Pipeline(Operator):
     """
+    Pipeline accepts a series of operators, schedulers, and a router. Calling a pipeline
```
dbogunowicz, line 29: I'd expand docstrings here a bit to provide the users with some fundamental info on operators, schedules, routers, and their relationships. Not a full-blown documentation, but something to give the reader an initial idea of what those are.

src/deepsparse/v2/pipeline.py
```diff
@@ -13,88 +13,113 @@
...
+        output of the previous operator.
+
+        :param inp: input to the operator. expected to be of any type that is
+        expected by the operator.
+        :param context: context to store the current the inputs, outputs, and operator
+        for each step of the router.
+
+        """
+        next_step = self.router.START_ROUTE
+        while next_step != self.router.END_ROUTE:
```
dbogunowicz, line 67: Just to make sure: even though we linearly dispatch the consecutive router steps, in reality, the operators are not run sequentially - this all depends on the router's logic.

src/deepsparse/v2/pipeline.py
```diff
@@ -13,88 +13,113 @@
...
             return pipeline_output, context
 
         return pipeline_output
 
     def validate(self):
-        router_validation = self.router.validate(self.stages)
+        """
+        Validate that compatability of the router and operators provided.
+        """
+        router_validation = self.router.validate(self.ops)
```
dbogunowicz, line 118: I see that you are trying to get rid of the name `stages`, I like that. Could we do a sweep through `v2` directory and made sure that all naming of `stages` is deprecated in favour of `ops`?

src/deepsparse/v2/routers/router.py
```diff
@@ -13,61 +13,67 @@
...
+__all__ = ["Router", "LinearRouter"]
 
 
 class Router:
     """
-    Routers must implement a run method which runs a series of operators
-    for a pipeline for a given input. Base Router runs operators linearly
-    in a series
+    Routers dicate the next operator to run. Each Router must implement a next function,
+    which dicates the index or key of the next operator to run.
```
dbogunowicz, line 28:
```suggestion
    which dictates the index or key of the next operator to run.
```

dbogunowicz (commented):
Excellent job @dsikka!
I left a lot of comments, but in general I am very happy about the refactor and the direction that we are heading towards!

src/deepsparse/v2/image_classification/pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+        image_size: Optional[Tuple[int]] = None,
+        top_k: int = 1,
+        engine_kwargs: Dict = None,
+    ):
+        model_path = model_to_path(model_path)
+
+        if not engine_kwargs:
+            engine_kwargs = {}
+        elif engine_kwargs.get("model_path") != model_path:
+            # TODO: swap to use logger
```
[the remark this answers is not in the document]
dsikka: instead of printing, these statements should be logged.

src/deepsparse/v2/image_classification/pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+class ImageClassificationPipeline(Pipeline):
+    def __init__(
+        self,
+        model_path: str,
+        class_names: Union[None, str, Dict[str, str]] = None,
+        image_size: Optional[Tuple[int]] = None,
+        top_k: int = 1,
+        engine_kwargs: Dict = None,
+    ):
+        model_path = model_to_path(model_path)
```
bfineran, line 40: why is this happening here instead of the engine operator? not sure that Pipelines should have much construction logic outside of deferring to operators

src/deepsparse/v2/image_classification/pipeline.py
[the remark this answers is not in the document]
bfineran: should definitely be debug logs

src/deepsparse/v2/image_classification/pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+        postprocess = ImageClassificationPostProcess(
+            top_k=top_k, class_names=class_names
+        )
+
+        engine = EngineOperator(**engine_kwargs)
+
+        ops = [preproces, engine, postprocess]
+        router = LinearRouter(end_route=len(ops))
+        scheduler = [OperatorScheduler()]
```
bfineran, line 61: mentioned before but we should definitely think through how schedulers may be propagated when nesting pipelines

src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -0,0 +1,118 @@
...
+
+DEEPSPARSE_ENGINE = "deepsparse"
+ORT_ENGINE = "onnxruntime"
+
+SUPPORTED_PIPELINE_ENGINES = [DEEPSPARSE_ENGINE, ORT_ENGINE]
+
+__all__ = ["EngineOperator"]
+
+
+class EngineOperator(Operator):
```
bfineran, line 31: definitely should have a defined input/output schema here for routers and schedulers to take advantage of

src/deepsparse/v2/operators/operator.py
```diff
@@ -87,4 +89,36 @@ def __call__(
...
+        """
+        raise NotImplementedError
+
+    def expand_inputs(self, **kwargs):
+        """
+        Generic function to handle expanding values.
+        """
+        raise NotImplementedError
+
+    def condense_inputs(self, **kwargs):
```
bfineran, line 114: what's the reasoning for making this a top level function? are routers always going to call condense/expand?

tests/deepsparse/v2/test_basic_pipeline.py
```diff
@@ -19,21 +35,21 @@ class AddOneOperator(Operator):
...
+    def run(self, inp: IntSchema, context: Context) -> Dict:
+        return {"value": inp.value + 2}
 
 
 AddThreePipeline = Pipeline(
-    stages=[AddOneOperator(), AddTwoOperator()],
-    router=Router(),
+    ops=[AddOneOperator(), AddTwoOperator()],
+    router=LinearRouter(end_route=2),
     schedulers=[OperatorScheduler()],
```
bfineran, line 53: might as well throw a simple composability test in here

src/deepsparse/v2/image_classification/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# flake8: noqa
+from .postprocess_operator import *
+from .preprocess_operator import *
+
+
+from .pipeline import *  # isort:skip
```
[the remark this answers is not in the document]
rahul-tuli: I would argue the dotted imports are cleaner, because they don't need to change if the top level package names are ever updated

src/deepsparse/v2/image_classification/pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+        engine_kwargs["model_path"] = model_path
+
+        preproces = ImageClassificationPreProcess(
+            model_path=model_path, image_size=image_size
+        )
+        postprocess = ImageClassificationPostProcess(
+            top_k=top_k, class_names=class_names
+        )
+
+        engine = EngineOperator(**engine_kwargs)
```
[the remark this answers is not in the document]
rahul-tuli: +1

src/deepsparse/v2/image_classification/postprocess_operator.py
```diff
@@ -0,0 +1,76 @@
...
+    ):
+        self.top_k = top_k
+        if isinstance(class_names, str) and class_names.endswith(".json"):
+            self._class_names = json.load(open(class_names))
+        elif isinstance(class_names, dict):
+            self._class_names = class_names
+        else:
+            self._class_names = None
+
+    def run(self, inp: Any, context: Optional[Context]) -> Dict:
```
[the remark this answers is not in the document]
rahul-tuli: Also is context used here?

src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -0,0 +1,118 @@
...
+        batches_outputs = list(map(self.engine, batches))
+        engine_outputs = self.condense_inputs(
+            batch_outputs=batches_outputs, orig_batch_size=orig_batch_size
+        )
+        return engine_outputs
+
+    def expand_inputs(self, **kwargs):
+        return split_engine_inputs(kwargs["engine_inputs"], self._batch_size)
+
+    def condense_inputs(self, **kwargs):
```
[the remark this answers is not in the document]
dsikka: from product specs

src/deepsparse/v2/operators/operator.py
```diff
@@ -87,4 +89,36 @@ def __call__(
...
+        """
+        Generic function to handle condensing values.
+        """
+        raise NotImplementedError
+
+    def yaml(self):
+        pass
+
+    def json(self):
```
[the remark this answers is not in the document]
dsikka: serialization (will be implemented after the text gen mvp lands)

src/deepsparse/v2/image_classification/pipeline.py
```diff
@@ -0,0 +1,62 @@
...
+        class_names: Union[None, str, Dict[str, str]] = None,
+        image_size: Optional[Tuple[int]] = None,
+        top_k: int = 1,
+        engine_kwargs: Dict = None,
+    ):
+        model_path = model_to_path(model_path)
+
+        if not engine_kwargs:
+            engine_kwargs = {}
+        elif engine_kwargs.get("model_path") != model_path:
```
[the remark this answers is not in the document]
dsikka: I think for usability, it's slightly more clear if the model_path is defined as an explicit argument.

src/deepsparse/v2/image_classification/postprocess_operator.py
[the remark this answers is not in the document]
dsikka: It may or may not be used. Dependent on the Operator. Hence why Optional. I have typing set to Any at the moment to follow the the same function signature as the parent class. We can update to be the specific schema, once we harden inputs/ouputs.

src/deepsparse/v2/pipeline.py
dsikka: not sure I follow your question

src/deepsparse/v2/operators/operator.py
dsikka: keeping for now, will be updated in a following PR once I add in split/joining on the pipeline level

src/deepsparse/v2/pipeline.py
```diff
@@ -13,90 +13,119 @@
...
             raise ValueError(
-                "Only 1 in-line argument may be supplied to Pipeline which "
-                f"must be a Schema, found: {len(args)}"
+                "Only 1 unnamed arg may be supplied to a Pipeline, the input expected"
+                "for the first Operator."
             )
         if args and kwargs:
             raise ValueError(
-                "Pipeline can only run either a single in-line argument schema or a "
+                "Pipeline can only run either a in-line arguments or a "
```
tdg5, line 108: This isn't new code, but I think the inclusion of this exception violates the [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) since it prevents `Pipeline` from being able to act as an `Operator`.  This is because [`Scheduler.submit` explicitly calls a given operator with both an argument and a keyword argument](https://github.com/neuralmagic/deepsparse/blob/e5220256ccbf320df33af4a8c32d61128ceed729/src/deepsparse/v2/schedulers/scheduler.py#L55). YMMV, but @horheynm and I ran into some sadness because of this 😭

tests/deepsparse/v2/test_basic_pipeline.py
dsikka: like router vailidity?

src/deepsparse/v2/pipeline.py
dsikka: I see what you're saying but I guess we've kinda reserved args/kwargs for inputs which may change between operators/pipelines and made them responsible for resolving the appropriate schema needed. So you could apply the condition above to the operator as well and it would be fine. The other positional arguments (`return_context` in pipeline and `context` in operators) are passed in explicitly.

dsikka: Ran through an example from @horheynm and I see where this is causing a problem. Thanks!

src/deepsparse/v2/pipeline.py
Satrat: I think I have a similar question, for instance in the image classification pipline we define the operators:
`ops = [preproces, engine, postprocess]` as a list, and could run the pipeline just stepping through the operators in list order. Iis the purpose of the router for cases where two operators can be run in parallel? `Does the order of ops matter at all?

src/deepsparse/v2/routers/router.py
```diff
@@ -88,8 +95,13 @@ def validate(operators: List[Operator]) -> Union[bool, str]:
...
+                # TODO: Log error message below
+                """
                 return (
                     f"Operator at idx {idx}: {type(operators[idx])} has invalid "
                     f"output schema {current_output_schema} for next operator "
                     f"{type(operators[idx + 1])} which requires {next_input_schema}"
                 )
+                """
+                return False
+        return True
```
Satrat, line 107: Why was this commented out?

src/deepsparse/v2/pipeline.py
Satrat: Ah ok I understand now looking through the router code, we have a base router that just runs the ops linearly. I think I'm still unclear about the use case for multiple types of routers though, since it looks like we are running operators sequentially and not allowing for computing them in parallel? In what case is the linear router not sufficient?

Satrat (approved):
Nice work here, had a few clarifying questions but overall this is really clean!

src/deepsparse/v2/operators/operator.py
```diff
@@ -13,59 +13,57 @@
...
+from deepsparse.v2.utils import Context
 
 
 __all__ = ["Operator"]
 
 
 class Operator(ABC):
     """
-    Base operator class - can represent any part of an ML pipeline
+    Base operator class - an operator should be defined for each aotmic, functional
```
dbogunowicz, line 28:
```suggestion
    Base operator class - an operator should be defined for each atomic, functional
```

src/deepsparse/v2/pipeline.py
dbogunowicz: I think the `GraphRouter` is the answer to my question :)

src/deepsparse/v2/routers/router.py
```diff
@@ -13,61 +13,67 @@
...
+
     """
 
-    @staticmethod
-    def run(
-        inp: OperatorSchema,
-        operators: List[Operator],
-        scheduler: OperatorScheduler,
-    ) -> Tuple[OperatorSchema, Context]:
+    def __init__(self, end_route: Union[str, int], start_route: Union[str, int]):
```
dbogunowicz, line 35: One of. the arguments should also be `route: Union[Dict, List]` no?

dbogunowicz (commented):
Few minor comments.

src/deepsparse/v2/routers/router.py
dsikka: yup. updated in the follow-up PR

src/deepsparse/v2/routers/router.py
dsikka: this returns a boolean, not a string. The strings should be logs, once include a logger.

PR #1328, opened 2023-10-18
[Fix][Text Generation] Fix the. outdated non-kv cache inference pathway

dbogunowicz:
As always, our new features slightly broke the non-kv cache inference. This PR updates the pathway.
As a follow-up, my high priority is to finalize the test suite to also include basic testing of the non-kv cache pathway, working on it as we speak.


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -683,7 +683,10 @@ def engine_forward(
                 )
                 for prompt_logit in prompt_logits:
                     token_generator.generate(prompt_logit)
-                return numpy.array([self.tokens]), prompt_logits
+                yield numpy.array([token_generator.tokens]), prompt_logits, [
```
rahul-tuli, line 686: Could you add quick local test in the description? want to test if yield works as expected here

src/deepsparse/transformers/engines/nl_decoder_engine.py
```diff
@@ -154,18 +154,29 @@ def run(
...
+                return self.engine._eng_net.execute_list_out(
+                    inputs, kv_cache.engine_internal_cache
+                )
+            else:
+                # run the engine assuming external kv cache
+                # management.
+                return self.engine.run(inputs, val_inp, kv_cache)
+        else:
+            # run the engine without the kv cache support
+            return self.engine.run(inputs, val_inp)
```
rahul-tuli, line 179: could we change the structure slightly?

I think following could read better:

```python
if kv_cache is None:
    return ...
return ( ... if bool(kv_cache.engine_internal_cache) else ...)
```

The comments can stay they are nice touch!

src/deepsparse/transformers/pipelines/text_generation.py
dbogunowicz: The tests are added as a separate PR #1330

src/deepsparse/transformers/engines/nl_decoder_engine.py
dbogunowicz: much more readable, thanks for a good suggestion!

Satrat (approved):
Tested that this is working!

bfineran (approved):

PR #1330, opened 2023-10-18
[LLM Tests] Additional tests for non-kv cache support 

dbogunowicz:
Includes fixes described in #1328

### DIFF
--- a/src/deepsparse/transformers/utils/token_generator.py
+++ b/src/deepsparse/transformers/utils/token_generator.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List
+from typing import List, Optional
 
 import numpy
 
@@ -29,7 +29,7 @@ class TokenGenerator:
     def __init__(
         self,
         logits_shape: int,
-        tokens: List[int] = [],
+        tokens: Optional[List[int]] = None,
         deterministic: bool = True,
         sampling_temperature: float = 1.0,
         top_k: int = 0,
@@ -61,7 +61,7 @@ class TokenGenerator:
         self.top_p = top_p
         self.frequency_penalty = frequency_penalty
         self.presence_penalty = presence_penalty
-        self.tokens = tokens
+        self.tokens = [] if tokens is None else tokens
 
         self._initialize_token_frequencies()
 
@@ -168,5 +168,5 @@ class TokenGenerator:
 
     def _initialize_token_frequencies(self):
         unique_tokens, frequencies = numpy.unique(self.tokens, return_counts=True)
-        for token, frequnecies in zip(unique_tokens, frequencies):
-            self.token_frequencies[token] += frequnecies
+        for token, freq in zip(unique_tokens, frequencies):
+            self.token_frequencies[token] += freq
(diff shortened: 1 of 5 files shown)


src/deepsparse/transformers/utils/token_generator.py
```diff
@@ -29,7 +29,7 @@ class TokenGenerator:
     def __init__(
         self,
         logits_shape: int,
-        tokens: List[int] = [],
+        tokens: Optional[List[int]] = None,
```
dbogunowicz, line 32: 🦘

src/deepsparse/transformers/utils/token_generator.py
```diff
@@ -168,5 +168,5 @@ def _update_frequencies(self, token: numpy.ndarray):
 
     def _initialize_token_frequencies(self):
         unique_tokens, frequencies = numpy.unique(self.tokens, return_counts=True)
-        for token, frequnecies in zip(unique_tokens, frequencies):
-            self.token_frequencies[token] += frequnecies
+        for token, freq in zip(unique_tokens, frequencies):
+            self.token_frequencies[token] += freq
```
bfineran, line 172: nit: always prefer to write out variable names unless it forces bad line breaks

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -449,6 +451,51 @@ def test_streaming_mode_returns_generator(self, setup):
...
+        # prompt logits + one logit for the new generated token
+        logits = output.generations[0].score[-(self.prompt_length + 1) :, :]
+        generated_logits, prompt_logits, *_ = self.torch_ground_truth
+        logits_gt = numpy.concatenate(
+            [prompt_logits[0], generated_logits[0, :1, :]], axis=0
+        )
+        assert numpy.allclose(logits, logits_gt, atol=self.precision)
+
+    def _get_model_path_no_cache(self):
+        from sparsezoo import Model
```
bfineran, line 482: why is this import down here?

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -449,6 +451,51 @@ def test_streaming_mode_returns_generator(self, setup):
...
+
+        if not self.model_path.startswith("zoo:"):
+            pytest.skip("For this test, for now only the zoo model is supported")
+        training_dir_path = Model(self.model_path).training.path
```
bfineran, line 486: we don't want to download the entire training directory just to get one file, especially since the files can be so large, let's use `.get_file` instead? if already downloaded for torch files, that's still ok, just don't see a reason to force it here

src/deepsparse/transformers/utils/token_generator.py
dbogunowicz: this was just a quick intervention to 1) fix the existing typo 2) avoid reusing the existing `frequencies` variable name

tests/deepsparse/transformers/pipelines/test_text_generation.py
dbogunowicz: good point. I will only download those files that are absolutely necessary.

PR #1332, opened 2023-10-18
[Text Generation] Fail if `prompt_sequence_length % 4 != 0`

dbogunowicz:


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -252,6 +252,15 @@ def __init__(
             if "WAND_OPT_FLAGS" not in os.environ:
                 os.environ["WAND_OPT_FLAGS"] = "default,~pyramids"
 
+        # the current requirement on the deepsparse engine
+        # is that prompt_sequence_length
+        # must be a multiple of four.
+        # for simplicity let's extend this requirement to all engines
+        if prompt_sequence_length % 4 != 0:
```
mgoin, line 259: what about 1 🥇

rahul-tuli: Michael you got that eagle eye

dbogunowicz: woops

dbogunowicz, line 259:
```suggestion
        if prompt_sequence_length % 4 == 0:
```

dbogunowicz: thanks for erroneous recommendation copilot...

dbogunowicz: im confused. "if x modulo 4 does not equal 0: raise an error" -> this makes sense

mgoin: we also need to allow prompt_sequence_length=1 for the single-token case so `if prompt_sequence_length is 1 or prompt_sequence_length % 4 != 0:`

dbogunowicz: oh yeah, haha, silly.

mgoin (approved):

bfineran (approved):

PR #1354, opened 2023-10-25
[text generation] fix issues with top_p

bfineran:
fixes two issues with `top_p`:

1. setting filtered values to `-inf` is incompatible with `numpy.choice`, was causing issues with numpy.choice in certain scenarios
2. fixes top_p application to grab the true top `p` percent by weight of the logits softmax distribution (previously was sorting  descending causing the least likely tokens would have the highest density)

**minimal reproducible example:**
```python

pipeline = Pipeline.create(
    "text-generation",
    model_path="hf:neuralmagic/mpt-7b-chat-pruned50-quant",
    engine_type="onnxruntime",
    sequence_length=32,
    prompt_sequence_length=1,
)
generation_config = dict(top_p=0.5, max_new_tokens=5, do_sample=True)
pipeline("def hello_world:", generation_config=generation_config)
```

**test_plan:**
@horheynm to add unit tests in ongoing refactor

### DIFF
--- a/src/deepsparse/transformers/utils/token_generator.py
+++ b/src/deepsparse/transformers/utils/token_generator.py
@@ -19,6 +19,9 @@ import numpy
 from deepsparse.utils.data import numpy_softmax
 
 
+_MIN_FLOAT = numpy.finfo(numpy.float32).min
+
+
 class TokenGenerator:
     """
     Responsible for generating tokens, and contains functions that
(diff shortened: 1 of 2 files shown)


horheynm (dismissed):

mgoin (dismissed):
nice catches, thank you!

horheynm (approved):

mgoin (approved):

PR #1355, opened 2023-10-25
added comments like u said in the test_twitter_nlp.py

bhargavshirin:


mgoin:
@bhargavshirin unfortunately we are deprecating and removing this test so closing this PR. Thanks for the comments

PR #1357, opened 2023-10-26
Enable preparing nested OutputSchemas for serialization

dbogunowicz:
Currently, the `prep_outputs_for_serialization` function assumes that `pipeline_outputs` cannot be arbitrarily nested (allows up to one level of nesting). This is why, for more complex `BaseModels` returned by the python (e.g. `TextGenerationOutput`), the function will not properly convert nested numpy arrays to lists for serialization.

My diff makes sure that any `pipeline_output: BaseModel` that contains an arbitrary number and nesting depth of any fields that are either a `BaseModel`, `list` of `numpy.ndarray` is supported.

As a result, we can now serialize logits of the LLM output, that previously was not properly converted from numpy array to a list. 

```python
import requests

prompt =  ["name one former president of the USA"]
server_address = "http://0.0.0.0:5543/v2/models/gen/infer"

payload = {"prompt": prompt, "output_scores": True, "include_prompt_logits": True}
response = requests.post(server_address, json=payload)
response = response.json()

num_tokens_prompt_and_generated = len(response["generations"][0]["score"])

payload = {"prompt": prompt, "output_scores": True}
response = requests.post(server_address, json=payload)
response = response.json()

num_tokens_generated = len(response["generations"][0]["score"])
num_tokens_prompt = num_tokens_prompt_and_generated - num_tokens_generated
print(f"Number of prompt tokens: {num_tokens_prompt}")
print(f"Number of generated tokens: {num_tokens_generated}")
```

returns

```bash
Number of prompt tokens: 6
Number of generated tokens: 3
```

### DIFF
--- a/src/deepsparse/server/helpers.py
+++ b/src/deepsparse/server/helpers.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from http import HTTPStatus
-from typing import Any, Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 import numpy
 from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


src/deepsparse/server/helpers.py
```diff
@@ -75,23 +75,35 @@ def server_logger_from_config(config: ServerConfig) -> BaseLogger:
     )
 
 
-def prep_outputs_for_serialization(pipeline_outputs: Any):
+def prep_outputs_for_serialization(
```
bfineran, line 78: let's add in a quick unit test for this

bfineran (dismissed):

Satrat (approved):

bfineran (approved):

PR #1359, opened 2023-10-27
timer

horheynm:
Timer Draft
Minimal change to existing code, just need to make a super class and add timer.

Unfortunately, we cannot run Pipeline(Operator) as a Operator bc of 
if args and kwargs:
            raise ValueError(
                "Pipeline can only run either a in-line arguments or a "
                f"series of kwargs, found {len(args)} args and {len(kwargs)} kwargs"
            )

Either fix pipeline callable or can add middleware?


dsikka:
@horheynm Can you add sample code of what you were trying to do that was causing problems?

horheynm:
> @horheynm Can you add sample code of what you were trying to do that was causing problems?

here you go!

```python3
from typing import Dict

from pydantic import BaseModel

from deepsparse.v2 import Pipeline
from deepsparse.v2.routers import LinearRouter
from deepsparse.v2.schedulers import OperatorScheduler
from deepsparse.v2.utils import Context
from deepsparse.v2.timer import Timer


from src.deepsparse.v2.timed_pipeline  import time



class IntSchema(BaseModel):
    value: int


class AddOneOperator(Operator):
    input_schema = IntSchema
    output_schema = IntSchema

    def run(self, inp: IntSchema, context: Context) -> Dict:
        return {"value": inp.value + 1}


class AddTwoOperator(Operator):
    input_schema = IntSchema
    output_schema = IntSchema

    def run(self, inp: IntSchema, context: Context) -> Dict:
        return {"value": inp.value + 2}


AddThreePipeline = Pipeline(
    ops=[AddOneOperator(), AddTwoOperator()],
    router=LinearRouter(end_route=2),
)

# timer = Timer()
# AddThreePipeline.timer = timer

timed_pipeline = time(AddThreePipeline)


pipeline_input = IntSchema(value=5)
pipeline_output = timed_pipeline(pipeline_input)

print(pipeline_output.value)

# print(r)

assert pipeline_output.value == 8

"""
python -m profile -s time scratch/pipeline_v2.py
"""

```

PR #1365, opened 2023-10-30
[Pipeline Refactor][Text-Generation][No KV Cache Pipeline] Prepare scaffolding for no-kv cache pipeline

dbogunowicz:
The goal of this PR is to serve as a cornerstone for the next PR that fully implements the `TextGenerationPipelineNoCache`. The main goal of this diff is to establish elegant helper function shared between both TextGenerationPipeline(s)


src/deepsparse/transformers/helpers.py
```diff
@@ -53,39 +55,42 @@ def get_deployment_path(model_path: str) -> Tuple[str, str]:
     for running the transformers model in the deepsparse pipeline
 
     :param model_path: path to model directory, sparsezoo stub, or ONNX file
+    :param onnx_model_name: name of the ONNX file to look for in the deployment
```
dbogunowicz, line 58: This is to be able to explicitly name the model onnx that we are seeking. Useful, when looking for the model with no kv cache injected. Those models do not have predefined names, thus, it is useful to specify what we are looking for.

src/deepsparse/v2/text_generation/pipeline.py
```diff
@@ -34,10 +34,63 @@
...
+    def __init__(
+        self,
+        model_path: str,
+        sequence_length: int = 1024,
+        engine_kwargs: Optional[Dict] = None,
+        onnx_model_name: Optional[str] = None,
+        **kwargs,
+    ):
+
+        self.model_path = self.setup_onnx_file_path(
```
dbogunowicz, line 51: Line 51 to 56 should be ideally executed by the parent of `TextGenerationPipelineNoCache` and `TextGenerationPipeline` - the `TransformersPipeline`.

However, we encounter a bit of a pickle, since we need to pass down the "inheritance chain" ops, router and schedulers, so we cannot call `__init__` before we have those variables in place. This is not an issue for `TextGenerationPipelineNoCache`, but is an issue for `TextGenerationPipeline`. The latter needs to do a lot of class-specific work between calling `self.setup_onnx_file_path` and calling `super().__init__`. Let's think about whether we can improve the implementation.

src/deepsparse/v2/text_generation/pipeline.py
```diff
@@ -34,10 +34,63 @@
     ProcessOutputs,
     TokenGeneratorOperator,
 )
+from deepsparse.v2.transformers.pipeline import TransformersPipeline
 from deepsparse.v2.utils import PipelineState
 
 
-class TextGenerationPipeline(Pipeline):
+class TextGenerationPipelineNoCache(TransformersPipeline):
```
bfineran, line 41: let's move this to its own file

bfineran (changes requested):
as discussed in meeting - let's rework to not use this `TransformersPipeline` and just use helper functions where possible. can eventually add as an interface but following @dsikka want to avoid inheritance between pipelines where possible

src/deepsparse/v2/text_generation/pipeline.py
```diff
@@ -34,9 +35,66 @@
...
+            engine_kwargs=engine_kwargs,
+        )
+        self.verify_no_kv_cache_present()
+
+        # TODO: Setup the operators of this pipeline
+        ops = []
+        router = LinearRouter(route=ops)
+        scheduler = [OperatorScheduler()]
+        super().__init__(
+            **kwargs,
```
dbogunowicz, line 70:
```suggestion
```

PR #1375, opened 2023-11-01
[ContinuousBatching] ContinuousBatchingScheduler Implementation

bfineran:
Uses the utils in #1373 and #1374 to implement a scheduler for continuous batching.
The `ContinuousBatchingScheduler` tracks various `EngineOperators` and manages their input queues with `ContinuousBatchingQueues`.  The scheduler also runs multiple `ContinuousBatchingExecutorThreads` in parallel that consume these queues and actually run the multi-batch engine and return the correct futures from the scheduler `submit`.

**next steps include:**
* integration with KV Cache engine mode for text gen
* implementation of a singleton pattern so that the scheduler can easily be shared across pipelines
* text gen pipeline integration
* more involved heuristic for the `select_fn` to determine execution priority
* server integration
* heavier testing
* README

**test_plan:**
simple single execution unit test included, further tests should test multiple engines/operators/batch sizes with sufficient load to trigger multibatch execution - note that unit tests for multibatch are handled with the helpers

### DIFF
--- a/src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
+++ b/src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
@@ -1,0 +1,141 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from concurrent.futures import Future
+from threading import Lock
+from typing import List
+
+from deepsparse.v2.operators import EngineOperator, Operator
+from deepsparse.v2.schedulers.scheduler import OperatorScheduler
+from deepsparse.v2.schedulers.utils import (
+    ContinuousBatchingExecutorThread,
+    ContinuousBatchingQueues,
+)
+
+
+__all__ = ["ContinuousBatchingScheduler"]
+
+
+class ContinuousBatchingScheduler(OperatorScheduler):
+    """
+    Manages EngineOperator jobs that should be run with continuous batching.
+    Groups requests for the same engine into larger batches and returns
+    the result to the respeictive request threads after scheduled completion
+
+    :param max_workers: maximum number of threads to execute at once, default 1
+    """
+
+    def __init__(self, max_workers: int = 1):
+        self._max_workers = max_workers
+
+        self._mutex = Lock()
+
+        # Dict[EngineOperator, Dict[batch_size, Engine]]
+        self._operators_to_engines = {}  # EngineOperator -> Dict[batch_size, Engine]
+        self._queues = ContinuousBatchingQueues()
+
+        # create and start max number of worker threads
+        self._threads = [
+            ContinuousBatchingExecutorThread(self._queues, self._operators_to_engines)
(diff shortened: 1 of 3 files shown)


src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+__all__ = ["ContinuousBatchingScheduler"]
+
+
+class ContinuousBatchingScheduler(OperatorScheduler):
+    """
+    Manages EngineOperator jobs that should be run with continuous batching.
+    Groups requests for the same engine into larger batches and returns
+    the result to the respeictive request threads after scheduled completion
```
dbogunowicz, line 35:
```suggestion
    the result to the respective request threads after scheduled completion
```

dbogunowicz (approved):
Excellent. Can we maybe describe in PR description what Continuous Batching is, for clarity and posteriority?

dsikka (approved):

src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
```diff
@@ -0,0 +1,141 @@
...
+            worker_thread.start()
+
+    @property
+    def max_workers(self) -> int:
+        """
+        :return: maximum number of threads to execute at once
+        """
+        return self._max_workers
+
+    def submit(self, *args, operator: Operator, **kwargs) -> Future:
```
dsikka, line 64: all the code for the `ContinuousBatchingExecutorThread` is `EngineOperator` specific. Are we ever expecting non-engine operator as the input to this submit function?

bfineran: great observation - I think it's something we could support in the future, but not sure how much gain there would be. we could probably abstract it out, but it might get a little complicated since the engine operators are unique (and foreseeably unique) that they need to have an engine compiled against the target batch size and this must be swapped in at runtime

tests/deepsparse/v2/schedulers/test_continuous_batching_scheduler.py
```diff
@@ -0,0 +1,48 @@
...
+
+from deepsparse.v2.operators import EngineOperator
+from deepsparse.v2.schedulers import ContinuousBatchingScheduler
+
+
+def test_continuous_batching_executor_thread():
+    # simple test that ContinuousBatchingScheduler can be instantiated and return
+    # a result from a request, for testing multi-batch execution, making enough
+    # concurrent requests guarantee batched execution is out of scope
+    scheduler = ContinuousBatchingScheduler()
```
dsikka, line 27: Is the goal that we would initialize (at least) two schedulers, one `ContinuousBatchingScheduler` and the remaining `OperatorSchedulers` as part of the pipeline init? And then swap to use `ContinuousBatchingScheduler` when the operator being executed is of type `EngineOperator`?

dsikka:
One more question: in terms of next steps, you had written down: `integration with KV Cache engine mode for text gen`
Any reason this can't work with the `NLEngineOperator` as is currently? The `NLEngineOperator` inherits from the `EngineOperator`

tests/deepsparse/v2/schedulers/test_continuous_batching_scheduler.py
bfineran: yes exactly, the `ContinuousBatchingScheduler` would come first in the list of schedulers so it can pick up any relevant engine operator jobs

bfineran:
> One more question: in terms of next steps, you had written down: `integration with KV Cache engine mode for text gen` Any reason this can't work with the `NLEngineOperator` as is currently? The `NLEngineOperator` inherits from the `EngineOperator`
> 
> @bfineran

yeah a few things here:
1. the schemas need to implement the split/join since they don't inherit
2. I think we'll need to update the way the engine kwargs are passed so the shared `create_engine` function sets the right internal/external kv cache mode
3. the run function needs to get updated to accept an engine to be swapped out like we do in `EngineOperator`

PR #1377, opened 2023-11-02
[Fix] The benchmark logic when internal kv cache is involved

dbogunowicz:


bfineran (approved):

mgoin (approved):
very nice i appreciate it

PR #1378, opened 2023-11-02
Add hf: stub support to model_to_path

mgoin:
Tada - just like the transformer pipeline, support HF paths in common places like Engine construction and benchmarking

```
deepsparse.benchmark hf:mgoin/TinyStories-33M-ds -q
Fetching 9 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████| 9/9 [00:00<00:00, 139294.23it/s]
DeepSparse, Copyright 2021-present / Neuralmagic, Inc. version: 1.6.0 COMMUNITY | (3d596172) (optimized) (system=avx512_vnni, binary=avx512)
Original Model Path: hf:mgoin/TinyStories-33M-ds
Batch Size: 1
Scenario: sync
Throughput (items/sec): 346.2459
Latency Mean (ms/batch): 2.8814
Latency Median (ms/batch): 2.8814
Latency Std (ms/batch): 0.2968
Iterations: 3463
```

### DIFF
--- a/src/deepsparse/transformers/helpers.py
+++ b/src/deepsparse/transformers/helpers.py
@@ -28,8 +28,11 @@ import onnx
 from onnx import ModelProto
 
 from deepsparse.log import get_main_logger
-from deepsparse.utils.onnx import _MODEL_DIR_ONNX_NAME, truncate_onnx_model
-from sparsezoo import Model
+from deepsparse.utils.onnx import (
+    _MODEL_DIR_ONNX_NAME,
+    model_to_path,
+    truncate_onnx_model,
+)
 from sparsezoo.utils import save_onnx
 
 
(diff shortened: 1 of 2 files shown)


src/deepsparse/transformers/helpers.py
```diff
@@ -76,15 +76,17 @@ def get_deployment_path(model_path: str) -> Tuple[str, str]:
         deployment_path = zoo_model.deployment_directory_path
         return deployment_path, os.path.join(deployment_path, _MODEL_DIR_ONNX_NAME)
     elif model_path.startswith("hf:"):
+        # load Hugging Face model from stub
```
bfineran, line 79: now that we just grab the entire deployment directory, wonder if we can just use model_to_path here now, or even need the function at all.  either way, would be good to move this to a helper function since it's the same as the changes below

bfineran (approved):

src/deepsparse/transformers/helpers.py
mgoin: i'm good with just calling model_to_path here. i'm kicking around the idea of making this snapshot_download more intelligent so it will only download required files and fail before downloading if required files (model.onnx) aren't present

rahul-tuli (approved):
Great addition, thanks Michael!

andy-neuma (approved):
cool. thanks!

PR #1379, opened 2023-11-02
Update sentence_transformer.py with batched padding

mgoin:


src/deepsparse/sentence_transformers/sentence_transformer.py
```diff
@@ -265,6 +269,7 @@ def tokenize(
             # No padding needed
```
dbogunowicz, line 269:
```suggestion
```
Not true anymore

mgoin: Yeah, it should specify it isn't doing static max_length padding though - just padding to the largest sequence in the batch

PR #1384, opened 2023-11-03
[Pipeline Refactor] Split/Join Functionality for multiple prompts

dsikka:
Summary
- Introduce functionality such that multiple prompts can run, using `SPLIT_ROUTE` and `JOIN_ROUTE` nodes.
- Allows us to enable the `num_return_sequences` generation attribute, such that multiple generations can be produced
- Potentially still needs a good testing strategy? 

Testing

```python
pipeline = TextGenerationPipeline(
    model_path, prompt_sequence_length=3
)

def run_requests():
    prompts = ["The dog barked", "Hello!"]
    outputs = []
    input_value = TextGenerationInput(prompt=prompts, generation_kwargs={"num_return_sequences": 2, "do_sample": True, "max_length": 20})
    output = pipeline(input_value)
    return output


output = run_requests()
for x in output.generations:
    print("\n")
    print(x)
```

Outputs:

```bash
[GeneratedText(text=' and ran to the farm. They chased after the birds to chase Sisi and the chubby dog', score=None, finished=True, finished_reason='length'), GeneratedText(text=' and got with far and hotries on his arm. \n\nHis mom grabbed him out and', score=None, finished=True, finished_reason='length')]


[GeneratedText(text=' That was my name, Molly. The elephant was smiling.\n\nThe giant was gentle, and', score=None, finished=True, finished_reason='length'), GeneratedText(text=' Why?" The lady thought for a moment, and then said, "That\'s a tiny bird,', score=None, finished=True, finished_reason='length')]
```

### DIFF
--- a/src/deepsparse/v2/pipeline.py
+++ b/src/deepsparse/v2/pipeline.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 
-from typing import Dict, List, Union
+import copy
+from functools import partial
+from typing import Any, Dict, List, Union
 
 from deepsparse.v2.operators import Operator
 from deepsparse.v2.routers import Router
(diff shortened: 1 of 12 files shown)


src/deepsparse/v2/pipeline.py
```diff
@@ -136,6 +191,18 @@ def __call__(self, *args, **kwargs):
 
         return self.run(*args, **kwargs)
 
+    def expand_inputs(self, *args, **kwargs):
+        """
+        Generic function to handle expanding values.
+        """
+        raise NotImplementedError
```
bfineran, line 198: probably need a specific message here about split routes etc

src/deepsparse/v2/schedulers/scheduler.py
```diff
@@ -52,6 +53,22 @@ def submit(
             **kwargs,
         )
 
+    def can_map(self, *args):
+        """
+        args containing list of inputs to be used for each worker. This function if we
+        have sufficient workes available
+        """
+        if len(args[0]) <= self._threadpool._max_workers:
```
bfineran, line 61: don't need this check here - definitely can have items in same split that are queued up

src/deepsparse/v2/text_generation/process_outputs.py
```diff
@@ -52,19 +51,20 @@ def _create_generated_text_output(
...
-            inp.generated_logits if generation_config.output_scores else None
-        )
-        finished_reason = inp.finished_reason
+        generated_logits = generated_logits if generation_config.output_scores else None
         sequences = self.tokenizer.batch_decode(
             generated_tokens, skip_special_tokens=True
         )
 
-        finished_reason = [f for f in finished_reason if f]
+        finished_reason = [f[-1] for f in finished_reason]
```
dbogunowicz, line 67: What if `f[-1]`?

dbogunowicz (approved):

src/deepsparse/v2/text_generation/process_outputs.py
dsikka: we produce a finish reason for every generated token but it is None for all until the last one.  This is required for streaming

dbogunowicz: Makes sense. Resolving.

src/deepsparse/v2/schedulers/scheduler_group.py
```diff
@@ -56,22 +56,12 @@ def submit(
                     **kwargs,
                 )
 
-    def can_process(
-        self,
-        *args,
-        operator: Operator,
-        **kwargs,
-    ) -> bool:
+    def map(self, *args, func: Callable):
```
bfineran, line 59: would put this function in the base scheduler instead so all schedulers can share

bfineran (approved):

PR #1388, opened 2023-11-06
[Evaluator] Implementation of `EvaluationRegistry`

dbogunowicz:
## Feature Description

Implements `EvaluationRegistry`. The registry inherits from the `from sparsezoo.utils.registry import RegistryMixin`. It requires `sparsezoo` version with this diff included: https://github.com/neuralmagic/sparsezoo/pull/385


src/deepsparse/evaluation/registry.py
```diff
@@ -12,22 +12,109 @@
...
-class BaseEvaluationRegistry(ABC):
-    """
-    Interface for evaluation registry. The detailed implementation will be
-    handled in a separate PR.
-    """
 
-    @abstractmethod
-    def get(
-        self, target: str, integration: str, datasets: Union[List[str], str]
+DATA_LOGGING_REGISTRY = defaultdict(lambda: defaultdict(str))
```
dbogunowicz, line 22:
```suggestion
```

bfineran (commented):
take a look if we can just use the universal registry here (you can extend to add aliases) https://github.com/neuralmagic/sparsezoo/blob/148ee556fe3626ef00d01954eddf1afecd9ab2b4/src/sparsezoo/utils/registry.py#L36

src/deepsparse/evaluation/registry.py
```diff
@@ -11,12 +11,71 @@
...
+"""
+from typing import Any, Callable, List, Optional, Union
 
 from sparsezoo.utils.registry import RegistryMixin
 
 
 __all__ = ["EvaluationRegistry"]
 
 
+def _merge_name_alias(
```
bfineran, line 25: thoughts on just adding alias to the mixin?

dbogunowicz: Potentially beneficial. 
Maybe better idea would be to refactor `RegistryMixin` so it has the following UI:

```python
# register with default name
@Dataset.register()
class ImageNetDataset(Dataset):
    pass

# register with custom name
@Dataset.register(name="cifar-dataset")
class Cifar(Dataset):
    pass
    
# register with multiple names
@Dataset.register(name=["cifar-dataset", "some_alias"])
class Cifar(Dataset):
    pass
```
I find it more elegant and generic.

If you like it, confirm and I can implement it asap (I will need to tweak the `RegistryMixing` source code to small extent anyway)

Satrat: Seconded that having multiple aliases would be useful in the `RegistryMixin`

src/deepsparse/evaluation/registry.py
```diff
@@ -11,12 +11,71 @@
...
+    return name
+
+
 class EvaluationRegistry(RegistryMixin):
-    pass
+    """
+    Extends the RegistryMixin to enable registering and loading of evaluation
+    functions. Adds the ability to register a function under multiple names.
+    Since functions are stateless, we can register the same function under
+    multiple names and load it from the registry using any of the names.
```
Satrat, line 44: From what I can tell we wouldn't need this child class if `RegistryMixin` had the capability for multiple aliases? It might be cleaner to implement all the registry code in SparseZoo

src/deepsparse/evaluation/registry.py
dbogunowicz: Cool. Vox populi vox dei. I'll implement it :)

dbogunowicz:
@Satrat @bfineran I refactored the `RegistryMixin` and thus simplified this PR. Please re-review.

bfineran (approved):

PR #1390, opened 2023-11-07
[Evaluator] Implementation of `Evaluation` (output schema for `evaluation` module)

dbogunowicz:
## Feature Description

Implementation of the logic governing the `Evaluation` object. Any evaluation integration, by default, expects to return the results in the form `List[Evaluation]`.

The goal of this PR is to:
- by default enforce the structure of `List[Evaluation]` to be returned/printed by DeepSparse Evaluation API
- enable serialization of `evaluations: List[Evaluation]` to `yaml` and `json`
- moving the function `prep_outputs_for_serialization` to `prep_for_serialization" to reuse in a broader context (so it is no exclusive to `deepsparse.server` but to `deepsparse`).


tests/server/test_helpers.py
```diff
@@ -35,44 +31,6 @@ class DummyOutputSchema(BaseModel):
...
-            ),
-            DummyOutputSchema(
-                field_1=[[[1, 2, 3]]],
-                field_2={"key_1": [[[1, 2, 3]]]},
-                field_3=DummyOutputSchema(field_1=[0]),
-            ),
-        )
-    ],
-)
-def test_prep_outputs_for_serialization(unserialized_output, target_serialized_output):
```
dbogunowicz, line 55: All this fuss is because I want to reuse `prep_outputs_for_serialization` outside of `server`.
So I:
- change its name so it's more accurate
- change the name in all the places in the code where the function is used
- move the unit tests to appropriate place.

bfineran (approved):

PR #1391, opened 2023-11-07
[continuous batching] singleton pattern for scheduler

bfineran:
singleton pattern for continuous batching scheduler is favorable so that engine requests coming from different pipelines are scheduled together and do not interfere with each other. 

this diff provides a simple implementation and extends the docstring with a small example

### DIFF
--- a/src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
+++ b/src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
@@ -28,11 +28,32 @@ from deepsparse.v2.schedulers.utils import (
 __all__ = ["ContinuousBatchingScheduler"]
 
 
+_GLOBAL_SCHEDULER = None
+
+
 class ContinuousBatchingScheduler(OperatorScheduler):
     """
     Manages EngineOperator jobs that should be run with continuous batching.
     Groups requests for the same engine into larger batches and returns
-    the result to the respeictive request threads after scheduled completion
+    the result to the respective request threads after scheduled completion
+
+    Example code for getting or creating a shared instance for scheduling
+    between pipelines and adding an engine operator to the scheduler
+    within a pipeline
+
+    ```python
+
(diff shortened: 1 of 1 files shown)


src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
```diff
@@ -54,6 +75,17 @@ def __init__(self, max_workers: int = 1):
...
+    @classmethod
+    def get_instance(cls) -> "ContinuousBatchingScheduler":
+        """
+        :return: global instance of the continuous batching scheduler. If one
+            does not exist yet, a scheduler with a single worker thread to
+            schedule all jobs is created and started
+        """
+        if _GLOBAL_SCHEDULER is not None:
+            return _GLOBAL_SCHEDULER
+        return cls(max_workers=1)
```
rahul-tuli, line 87: Do we need to assign the new instance to `_GLOBAL_SCHEDULER`?

src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
```diff
@@ -28,11 +28,32 @@
...
+
+    class MyPipeline(Pipeline):
+
+        def __init__(self):
+            ...
+            engine_operator = EngineOperator(...)
+            ...
+            continuous_batching_scheduler = ContinuousBatchingScheduler.get_instance()
+            continuous_batching_scheduler.add_engine_operator(engine_operator)
+
```
dsikka, line 54: If this is the lifecycle we're proposing, doesnt that mean we can only start with one worker at any given point?

bfineran: yes, I think in general we will only want one worker so that for a given batch the engine will have full access to all resources. otherwise we'd need to run the engines in multistream mode which I am not sure is supported for KV cache right now

PR #1395, opened 2023-11-10
Timer Middleware Hardcoded

horheynm:
Initial implementation of timer middleware hardcoded.

Something that can extend this pr is to add a store, where pipeline can use to communicate to the middlewares.

Example codes in tests


src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+    def end(self, key):
+        end_time = time.time()
+
+        # TODO: while waiting for N seconds <- condition
+        if key in self.start_times:
+            start_time = self.start_times[key]
+            with self._lock:
+                del self.start_times[key]
+            self.measurements[key] = end_time - start_time
+        return None
```
horheynm, line 46: return bool to see if successly written

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+        self.start_times = {}
+
+    def start(self, key):
+        with self._lock:
+            self.start_times[key] = time.time()
+
+    def end(self, key):
+        end_time = time.time()
+
+        # TODO: while waiting for N seconds <- condition
```
horheynm, line 40: add lock at this level

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+from deepsparse.v2.utils.state import InferenceState
+
+
+class Timer:
+
+    _lock = threading.Lock()
+    condition = threading.Condition()
+
+    def __init__(self):
+
```
horheynm, line 29: self._lock = threading.Lock()

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+import time
+from functools import wraps
+from typing import Optional
+
+from deepsparse.v2.utils.state import InferenceState
+
+
+class Timer:
+
+    _lock = threading.Lock()
```
bfineran, line 25: lock should be local to a single timer object, not static across all
why should one timer be able to block another when writing to separate timing dictionaries?

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def end(self, key):
+        end_time = time.time()
+
+        # TODO: while waiting for N seconds <- condition
+        if key in self.start_times:
+            start_time = self.start_times[key]
+            with self._lock:
+                del self.start_times[key]
+            self.measurements[key] = end_time - start_time
```
bfineran, line 45: we should allow for multiple measurements to the same key, so instead of setting to a value, should append to a list by the key

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+
+class TimerMiddleware:
+
+    timer = Timer()
+
+    def start_event(self, name: str, state: Optional[InferenceState] = None):
+
+        is_inference_run = state is not None
+        is_state_coldstart = is_inference_run and not hasattr(state, "timer")
+        is_new_name = (
```
bfineran, line 80: why do we need a new timer for a new operator name?

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+    def end_event(self, name, state: Optional[InferenceState] = None):
+        """
+        Populate the timer state. If state is provided, populate all its
+        measurements into the timer state
+        """
+        if state and hasattr(state, "timer"):
+            state.timer.end(name)
+
+            if name in self.timer.measurements:
+                print(f"warning, {name} already exists")
```
bfineran, line 102: no prints in prod

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,106 @@
...
+            with self.timer._lock:
+                self.timer.measurements.update(state.timer.measurements)
```
bfineran, line 104: should make this a method of the timer rather than using a private var, especially so usage of the lock can be read in the scope of the timer object (for future debugging)

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: why are we reusing the key?

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: Each operator has its own timer

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: Yes will make it to logger.debug or something in that level

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: So each operator has its own timer, and a pipeline has its own timer that will be updated once the operator closes safely. Pipeline-level (or middleware level) timer will have the operator info. This is the op closing, writing to the pipeline-level timer

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,112 @@
...
+        Ends the timer and saves the runtime into measurements
+        :param key: the key to save the run time for
+        """
+        end_time = time.time()
+
+        with self._lock:
+            if key in self.start_times:
+                start_time = self.start_times[key]
+                del self.start_times[key]
+                self.measurements[key] = end_time - start_time
```
bfineran, line 47: let's make `self.measurements` a `Dict[str, List[float]]` so that we can keep multiple measurements for the same key (ie route splitting potentially)

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,112 @@
...
+# limitations under the License.
+
+import threading
+import time
+from typing import Dict, Optional
+
+from deepsparse.v2.utils.state import InferenceState
+
+
+class Timer:
```
bfineran, line 22: nit: need docstring with example usage

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def start_event(self, name: str, state: Optional[InferenceState] = None) -> None:
+        """
+        Start the timer for one of middleware state or inference state.
+        Inference state timer will be used/initialized if state is provided
+        """
+
+        # middleware level timer
+        if state is None:
+            self.timer.start(name)
```
bfineran, line 82: why do we need this? what is this timing?

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,112 @@
...
+        if state is None:
+            self.timer.end(name)
+            return True
+
+        if hasattr(state, "timer"):
+            state.timer.end(name)
+            return True
+        return False
+
+    def update_middleware_timer(self, name: str, state: InferenceState) -> bool:
```
bfineran, line 105: we should extend `end_event` so if the event is the end of inference we automatically call `update_middleware_timer`

bfineran, line 105: why do we need `name`, shouldn't we do this for every time in the timer?

src/deepsparse/v2/pipeline.py
```diff
@@ -51,12 +52,14 @@ def __init__(
         router: Router,
         schedulers: List[OperatorScheduler],
         pipeline_state: PipelineState = None,
+        # [TODO] middlewares: List[]
```
bfineran, line 55: let's take out any changes to pipeline.py for this PR and add it in the next PR that actually introduces middleware. do not want to leave the v2 branch in incomplete state

bfineran (changes requested):

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: this is for starting pipeline/middleware level timer if inference state is not provided

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: Yeah i thought of that, but do we want to populate middleware timer each time inference timer's measurements gets recorded?

horheynm: name is the key to update the middleware timer here. Once run, the updated measurements in middleware state look like
{
  "foo" : ...
   "ops" :  {"some_func": 0.1, "some_func2": ....}
}

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: if we keep a list how do we know what recording is what

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: ok

src/deepsparse/v2/pipeline.py
horheynm: ok

src/deepsparse/v2/middleware/timer_middleware.py
horheynm: ah so right now the middleware state's recording looks like 
```python3
{"foo" : 0.1, "bar": 0.2, 
"ops_1": {"func1": 0.1, ...} 
} 
we can make it 
{
"pipeline": {"foo" : -.1, ...}
"ops_1": {...}
}
```

src/deepsparse/v2/middleware/timer_middleware.py
bfineran: that would be ideal in the future so we could get sub timings for operators that are run in the case of an early exit/error.

for now, we can probably get away with just writing at the end of inference

bfineran: what is the difference in usage between ops and foo here?

horheynm: foo can be any measurements from the pipeline level,
ops can be any measurements from the inference state level

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __init__(self):
+        self._lock = threading.Lock()
+        self.measurements = {}
+        self.start_times = {}
+
+    def start(self, key: str):
+        """
+        Starts the timer
+        :param key: the key to track the timer for
```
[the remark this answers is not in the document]
dbogunowicz: everywhere :)

tests/deepsparse/v2/middleware/timer_middleware/test_timer_middleware.py
```diff
@@ -0,0 +1,83 @@
...
+        ops=[AddOneOperator(), AddTwoOperator()],
+        router=LinearRouter(end_route=2),
+        schedulers=[OperatorScheduler()],
+    )
+    pipeline_input = IntSchema(value=5)
+    pipeline_output = AddThreePipeline(pipeline_input)
+    assert pipeline_output.value == 8
+
+    timer = AddThreePipeline.timer_middleware.timer
+    assert "foo" in timer.measurements
```
dbogunowicz, line 42: leave a comment, make this foo and bar business less confusing

tests/deepsparse/v2/middleware/timer_middleware/utils.py
```diff
@@ -0,0 +1,172 @@
...
+    ):
+        self._save_run_time_to_inference_state("fp32", 1, inference_state)
+        self._save_run_time_to_inference_state("int4", 0.5, inference_state)
+
+        kwargs["inference_state"] = inference_state
+        kwargs["pipeline_state"] = pipeline_state
+
+        time_delta = 0.005
+        assert (
+            abs(1 - getattr(inference_state, "timer").measurements["fp32"]) < time_delta
```
dbogunowicz, line 80: unify the names of measurements

src/deepsparse/v2/pipeline.py
```diff
@@ -120,6 +125,7 @@ def _run_next_step(
         """
         Generic function to run a given func, process the output and determine the next
         step.
+
```
dbogunowicz, line 128:
```suggestion
```

src/deepsparse/v2/middleware/timer_middleware.py
dbogunowicz: I like the idea with nesting the timings, but names for the measurements should correspond automatically to the operator name.

PR #1402, opened 2023-11-14
[Text Generation][V2] End-to-end tests

dbogunowicz:
## Feature Description

Successfully migrated LLM end-to-end tests from v1 to v2.
Missing elements:
- Tests for no-kv cache pipeline are currently disabled (will be enabled once the `TextGenerationPipelineNoCache` is implemented
- We are not testing the kv cache state post-inference. To do so, we need to enable the `debug` flag, as we did previously in `v1`. I do not see it as a blocker for this PR, but I'd propose to scope it as a part of the next sprint.

## Testing

Both 
`pytest tests/deepsparse/v2/unit/`
`pytest tests/deepsparse/v2/integration_tests/`
are green.

### DIFF
--- a/tests/deepsparse/v2/integration_tests/test_llms.py
+++ b/tests/deepsparse/v2/integration_tests/test_llms.py
@@ -15,7 +15,7 @@
 This test suite consumes config files to test the text generation pipeline
 for various scenarios.
 
-A sample config file is a yaml that requires the following fields:
+A sample config file is a yaml that r_equires the following fields:
     cadence: The cadence of the tests. The available options are:
               "nightly", "weekly" and "commit". By default, only
               the tests that have cadence "commit" will be run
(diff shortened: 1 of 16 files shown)


tests/deepsparse/v2/integration_tests/test_llms.py
```diff
@@ -0,0 +1,358 @@
...
+    validate_internal_kv_cache,
+)
+
+
+CONFIGS_DIRECTORY = "tests/deepsparse/v2/integration_tests/configs"
+
+
+@pytest.fixture()
+def max_new_tokens() -> int:
+    return 64
```
dsikka, line 58: is the fix to edit the typo between max_new_tokens and max_length in this PR? or has it already landed?

dsikka (commented):
Marked everything that I think needs to be updated for now. Key things are the `generation_kwargs` being used and making sure the typo for max_length/max_new_tokens being addressed

dsikka (approved):

bfineran (approved):

PR #1408, opened 2023-11-15
[BugFix] Redefine ARG between stages

rahul-tuli:
Recently a bug was caught by @mgoin where our nightly docker images were not actually installing nightlies;

This was because the `BRANCH` arg was not redefined b/w stages and docker does not persist arg values across stages;
This PR fixes that


mgoin (approved):
interesting and thanks for adding the deepsparse pip check!

mgoin (approved):

PR #1409, opened 2023-11-16
[Pipeline Refactor][Text Generation][Continuous Batching] Integration

dsikka:
# Summary
- Integrates the `ContinuousBatchingScheduler` with the `TextGeneration` Pipeline
- This requires updating the `NLEngineOperator` such that batches greater than size 1 can be handled. Specifically includes updating the input/output schema to include split/join methods required by the `ContinuousBatchingScheduler` and update the `run` method such that inputs can be handled by any scheduler
- Refactors the join/split functionality such that instead of launching individual threads for each `SPLIT` node, poll to identify the operator that is currently finished running and schedules the next operator in its split route, until all the operators within each sub graph (i.e. split route) have finished running. This is done through the help of a dataclass `SplitRoute`

# Testing
- Tested with ORT and Deepsparse Engine + External kv_Cache. For internal kv_cache, the normal scheduler is used.
- Stills need more vigorous unit testing


src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -13,14 +13,14 @@
 # limitations under the License.
```
dbogunowicz, line 13: General comment: we should also add continuous batching to the e2e testing suite. This is why I'd first land the e2e tests PR ASAP, rebase this PR, and add tests as a part of this PR (optionally, we can add a feature branch, so that this can get reviewed and landed quickly to a feature branch; on top of that feature branch we add tests)

src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -29,12 +29,12 @@
...
 SUPPORTED_PIPELINE_ENGINES = [DEEPSPARSE_ENGINE, ORT_ENGINE]
 
-__all__ = ["EngineOperator"]
+__all__ = ["EngineOperator", "EngineOperatorInputs", "EngineOperatorOutputs"]
 
 
 class EngineOperatorInputs(BaseModel):
     engine_inputs: List = Field(description="engine_inputs")
-    engine: Optional[Engine] = Field(
+    engine: Optional[Any] = Field(
```
dbogunowicz, line 37: Why this becomes `Any`?

src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -131,6 +131,7 @@ def batch_size(self) -> int:
         """
         return self._batch_size
 
+    # TODO: maybe add a few args to make this less opaque?
```
dbogunowicz, line 134: Could you elaborate?

src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -142,7 +143,8 @@ def create_engine(
             constructor/compilation
         :return: inference engine
         """
-        onnx_file_path = self.model_path
+
+        onnx_file_path = kwargs.pop("model_path", self.model_path)
```
dbogunowicz, line 147: Why do we need to pop?

src/deepsparse/v2/operators/engine_operator.py
```diff
@@ -13,14 +13,14 @@
 # limitations under the License.
```
dbogunowicz, line 14: General comment: for easier reviewing, a quicker feedback loop, and higher velocity, consider breaking up PRs like this into smaller logical units. I know that it is often not possible. But doing so, would allow you to sneak in at least few additional tests, and would spare us going through a PR with 16 edited scripts :)

src/deepsparse/v2/pipeline.py
```diff
@@ -50,39 +53,41 @@ def __init__(
...
         self.ops = ops
         self.router = router
         self.schedulers = schedulers
         self.pipeline_state = pipeline_state
+        self._continuous_batching_scheduler = continuous_batching_scheduler
         self.validate()
 
         self._scheduler_group = SchedulerGroup(self.schedulers)
 
-    def _run_sequential(
```
dbogunowicz, line 64: I'd leave a bit more comments on what's going on. A developer who takes a peek into this functionality will probably have a lot of questions, regardless of her domain knowledge. The fact that we create new engines on demand is not easy to capture from the code alone (and why we do it?)

src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
```diff
@@ -161,8 +165,18 @@ def add_engine_operator(
...
+
+            override_model_path = None
+            # text generation/NLEngineOperator specific; could add generic method
+            # for all engine_operators, if desired
+            if hasattr(engine_operator, "override_model"):
+                override_model_path, _, _ = engine_operator.override_model(
+                    model_path=engine_operator.model_path, batch_size=batch_size
+                )
+
+            # will break for internal kv_cache; needs additional argument
```
dbogunowicz, line 177: I propose for now just raising `NotImplementedError`, we can figure it out later.

src/deepsparse/v2/text_generation/nl_engine_operator.py
```diff
@@ -59,17 +124,16 @@ def __init__(
...
             onnx_file_path,
             output_indices_to_be_cached,
             kv_cache_data_type,
-        ) = overwrite_onnx_model_inputs_for_kv_cache_models(
-            onnx_file_path=kwargs.get("model_path"),
-            batch_size=kwargs.get("batch_size", 1),
-            sequence_length=sequence_length,
-            input_ids_length=input_ids_length,
-        )
+        ) = self.override_model(self.model_path, batch_size=1)
```
dbogunowicz, line 136:
```suggestion
        ) = self.override_model_inputs(self.model_path, batch_size=1)
```

- The Verbose Damian

src/deepsparse/v2/text_generation/nl_engine_operator.py
```diff
@@ -86,43 +150,80 @@ def __init__(
 
         kwargs["engine_kwargs"] = engine_kwargs
         kwargs["model_path"] = onnx_file_path
+
         super().__init__(**kwargs)
 
-        self.input_ids_length = input_ids_length
+    def override_model(self, model_path: Union[str, Path], batch_size: int):
```
dbogunowicz, line 156: docstrings incomplete, return types absent

src/deepsparse/v2/text_generation/nl_engine_operator.py
```diff
@@ -86,43 +150,80 @@ def __init__(
...
+
+        if bool(kv_cache[0].engine_internal_cache):
             # conventionally, before dispatching
             # inputs to the engine, we validate them
             # if val_inp=True. However, in this case
             # we want to pass the empty kv cache inputs
             # (batch_size=0) to the engine. Therefore,
             # we skip the validation
+
+            # Internal kv_cache works for batch_size of 1 atm
```
dbogunowicz, line 193: I'd rather leave explicit NotImplementedError than a comment

src/deepsparse/v2/text_generation/nl_engine_operator.py
```diff
@@ -86,43 +150,80 @@ def __init__(
...
+        ) = overwrite_onnx_model_inputs_for_kv_cache_models(
+            onnx_file_path=model_path,
+            batch_size=batch_size,
+            sequence_length=self.sequence_length,
+            input_ids_length=self.input_ids_length,
+        )
+        return onnx_file_path, output_indices_to_be_cached, kv_cache_data_type
 
-    def run(self, inp: NlEngineInput, **kwargs) -> Any:
+    def run(self, inp: NLEngineInputs, **kwargs) -> NLEngineOutputs:
```
dbogunowicz, line 173: The run method got super bloated. Can we possibility refactor it, e.g. creating helper methods for better readibility?

src/deepsparse/v2/text_generation/pipeline.py
```diff
@@ -47,13 +51,13 @@ def __init__(
         internal_kv_cache: bool = True,
         force_max_tokens: bool = False,
         generation_config=None,
+        continuous_batch_sizes: list = None,
```
dbogunowicz, line 54:
```suggestion
        continuous_batch_sizes: Optional[list] = None,
```

src/deepsparse/v2/text_generation/pipeline.py
```diff
@@ -135,6 +139,26 @@ def __init__(
         compile_generated_tokens = CompileGeneratedTokens()
         join_output = JoinOutput(tokenizer=self.tokenizer)
 
+        # TODO: do we want to support lists for different engines?
+        continuous_batching_scheduler = None
```
dbogunowicz, line 143: This could be its own small function to improve readbility

src/deepsparse/v2/operators/engine_operator.py
dsikka: good catch. change I made while debugging. will switch back

src/deepsparse/v2/operators/engine_operator.py
dsikka: `model_path` isn't a kwarg needed for engine creation.  alternatively, could use the correct name and use `kwargs.get(...)` to handle the case where it isnt provided.

src/deepsparse/v2/operators/engine_operator.py
dsikka: yah sorry, the refactor wasn't expected.

src/deepsparse/v2/pipeline.py
dsikka: nothing here is creating a new engine. that would have to be defined within the `continuous_batching_scheduler`. This is just scheduling.

src/deepsparse/v2/text_generation/nl_engine_operator.py
dsikka: well we dont use the continuous_batching_engine for this case. so the batch size will always be 1

src/deepsparse/v2/text_generation/nl_engine_operator.py
dsikka: I can refactor once the case for internal kv cache is also added.

src/deepsparse/v2/pipeline.py
```diff
@@ -93,59 +98,84 @@ def _apply_split(self, inp: Any, inference_state: InferenceState):
...
-        futures = self._scheduler_group.map(
-            batches,
-            inference_state_list,
-            func=run_with_state,
-        )
-        return self.condense_inputs([x.result() for x in futures])
 
-    def _run_next_step(
+        # Start each SPLIT_ROUTE; store the Future for each.
+        for i in range(len(split_routes)):
```
bfineran, line 115: would be good to move this into a separate helper function/file that takes in the scheduler, subgraphs, etc and performs the scheduling loop

src/deepsparse/v2/schedulers/continuous_batching_scheduler.py
dsikka: this case wont ever run for internal kv cache

bfineran (approved):
LGTM - as mentioned in the MLE last weekly, we'll need to update the UX around specifying continuous batching

dbogunowicz (approved):

dbogunowicz:
Condiitonal approval -> please run the relevant integration/unit tests before landing

PR #1414, opened 2023-11-17
[Text Generation] Avoid mutating the original logits in-place when mapping from logits to token.

dbogunowicz:
## Feature Description

`TokenGenerator.generate(logits: numpy.ndarray)` takes a logits array and, if `do_sample=True`, optionally mutates them to enforce the appropriate sampling strategy. 

This results in the correct generation of tokens, but an in-place modification of the logits array. Logits array is then returned to the user in the mutated form. This can be confusing for the users who are interested in logits value, specially when returned together with prompt logits:

image
First column "before": the generated logits returned to the user are mutated (the distribution is spikier, arguably close to Dirac distribution)
Second column "after": the generated logits are the non-mutatated, original logits, consistent with the prompt logits.

### DIFF
--- a/src/deepsparse/transformers/utils/token_generator.py
+++ b/src/deepsparse/transformers/utils/token_generator.py
@@ -77,6 +77,10 @@ class TokenGenerator:
         :param logits: the logits from the model with shape (vocab_size,)
         :return: the sampled token
         """
+        # make a copy of logits to avoid modifying the original
+        # logits distribution in-place
+        logits = logits.copy()
+
         if self.deterministic:
             token = numpy.argmax(logits)
             self.tokens.append(token)


src/deepsparse/transformers/utils/token_generator.py
```diff
@@ -77,6 +77,10 @@ def generate(self, logits: numpy.ndarray) -> numpy.ndarray:
         :param logits: the logits from the model with shape (vocab_size,)
         :return: the sampled token
         """
+        # make a copy of logits to avoid modifying the original
+        # logits distribution in-place
+        logits = logits.copy()
```
mgoin, line 82: does this only happen when we are explicitly returning logits? this seems like it could be expensive

dbogunowicz: Good point. This will happen regardless of whether we are returning logits to the user or not. However, I do not think that it is worth making this logic more complex in favor of memory benefits. Note that those are not the logits for the full sequence being copied, but only the last logits array of shape `vocab_size`.

mgoin (approved):

PR #1420, opened 2023-11-20
[Pipeline Refactor] Operator Registry

dsikka:
Example using `Operator.create(...)` for the registered text generation pipeline operator 

```python

from deepsparse.transformers.pipelines.text_generation import TextGenerationInput


pipeline = Operator.create(
    task="text_generation",
    model_path=model_path,
    prompt_sequence_length=3,
    engine_kwargs={"engine_type": "onnxruntime"},
)


def run_requests():
    prompts = [["Hello there!", "How are you?"]]
    outputs = []
    for i in range(len(prompts)):
        input_value = TextGenerationInput(
            prompt=prompts[i],
            generation_kwargs={
                "do_sample": False,
                "max_length": 20,
            },
        )
        output = pipeline(input_value)
        yield output


output = run_requests()
for x in output:
    for g in x.generations:
        print("\n")
        print(g)



```

### DIFF
--- a/src/deepsparse/v2/operators/registry.py
+++ b/src/deepsparse/v2/operators/registry.py
@@ -12,67 +12,40 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Optional, Type
+from typing import Type
 
 from deepsparse.v2.task import SupportedTasks, dynamic_import_task
+from sparsezoo.utils.registry import (
+    RegistryMixin,
+    get_from_registry,
+    register,
+    registered_names,
+)
 
-
-_REGISTERED_OPERATORS = {}
 
 __all__ = ["OperatorRegistry"]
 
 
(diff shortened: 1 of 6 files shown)


src/deepsparse/v2/operators/registry.py
```diff
@@ -0,0 +1,108 @@
...
+
+from deepsparse.v2.task import SupportedTasks, dynamic_import_task
+
+
+_REGISTERED_OPERATORS = {}
+
+__all__ = ["OperatorRegistry"]
+
+
+class OperatorRegistry:
```
bfineran, line 25: take a look at registry mixin https://github.com/neuralmagic/sparsezoo/blob/fe72195581966f460680b222b1fd0caef5874e71/src/sparsezoo/utils/registry.py#L36

bfineran (approved):

PR #1425, opened 2023-11-21
[Fix] deepsparse – tokenizer stride size issue with original transformers

dbogunowicz:
Fix for: https://app.asana.com/0/1205724081986514/1205993362080974
(the exhaustive explanation for this fix is in the ticket)
@dbarbuzzi this fix works for me, please confirm whether it also solves your issue)
Will cherry-pick once approved.

### DIFF
--- a/src/deepsparse/transformers/pipelines/question_answering.py
+++ b/src/deepsparse/transformers/pipelines/question_answering.py
@@ -37,6 +37,7 @@ import collections
 import json
 import logging
 import os
+import warnings
 from typing import Any, Dict, List, Optional, Tuple, Type
 
 import numpy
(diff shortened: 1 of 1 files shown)


dbarbuzzi:
The changes work with the minimal example reported. Quick question, though, is it expected that if the consumer uses a non-default sequence_length that they should also be manually defining the doc_stride? I ask because now if I e.g. set a sequence_length of 32 in the constructor with no other kwargs, it will fail the new check in the PR as the doc_stride simply defaults to `62` regardless.

dbogunowicz:
@dbarbuzzi I understand your concerns. Let me take a look at what are HF default settings.

dbogunowicz:
@dbarbuzzi 
https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/question_answering.py#L408
It's pretty explicit here.

dbogunowicz:
@dbarbuzzi let's set doc_stride to None by default and compute it dynamically to avoid the issue. I think this is fine.

src/deepsparse/transformers/pipelines/question_answering.py
```diff
@@ -160,6 +160,16 @@ def __init__(
 
         super().__init__(**kwargs)
 
+        if doc_stride is None:
```
bfineran, line 163: let's see if we can read the number of special tokens from the tokenizer or tokenizer config and cap doc stride at `min(doc_stride, sequence_len - num_special_tokens)`

bfineran: and then let's add short warning

bfineran (changes requested):

bfineran (approved):

PR #1436, opened 2023-11-29
[DeepSparse Evaluator API][UI] Simplify serialization

dbogunowicz:
## Feature Description

Introduces a few UI improvements to the `Result` serialization functionality. This includes:
- decoupling printing to CLI and saving the serialized results (printing to CLI will be covered in the separate PR)
- serializing the full `Result` object, not only `Result.formatted`
- turning `save_evaluations` into `save_result` and making sure it has only one responsibility (only saving, not returning data)
- updating and simultaneously simplifying the appropriate tests


src/deepsparse/evaluation/cli.py
```diff
@@ -210,7 +210,7 @@ def main(
         **integration_args,
     )
 
-    _LOGGER.info(f"Evaluation done. Results:\n{result}")
+    _LOGGER.info(f"Evaluation done. Result:\n{result.formatted}")
```
dbogunowicz, line 213: For now, we just print `result.formatted`. In the next PR, I will change it, so we print a pretty table to the user.

src/deepsparse/evaluation/integrations/llm_evaluation_harness.py
```diff
@@ -100,8 +100,10 @@ def integration_eval(
 
     results_raw = evaluator.simple_evaluate(**evaluator_input.dict())
 
+    inputs_dict = evaluator_input.dict()
+    del inputs_dict["model"]
```
dbogunowicz, line 104: We need explicitly to drop `model` key from inputs. `evaluator_input.model` is a complex `DeepsparseLM` object that should be excluded from the inputs (its a big wrapper around `Pipeline`, so not only its structure is complex and ugly, but also the object is very difficult to serialize)

bfineran (dismissed):

bfineran (approved):

Satrat (approved):

PR #1437, opened 2023-11-29
initial middleware manager

horheynm:
initial middleware manager implementation

Each middleware should inherit `Middlware`, and should define `__name__`, `start_event` and `end_event`


src/deepsparse/v2/middleware/middleware_manager.py
```diff
@@ -0,0 +1,54 @@
...
+    def _init(self, middleware: List):
+        for middleware in middleware:
+            init_middleware = middleware()
+            self.middlewares[init_middleware.__name__] = init_middleware
+
+    def start_event(self, name: str, *args, **kwargs):
+        self.middlewares[name].start_event(*args, **kwargs)
+
+    def end_event(self, name: str, *args, **kwargs):
+        self.middlewares[name].end_event(*args, **kwargs)
```
bfineran, line 33: why don't we just call start and end for every single middleware?

bfineran (commented):
to land this - let's 
1. get `MiddlewareManager` integrated into `Pipeline`
2. get a simple test up for interacting with `MiddlewareManager` in the pipeline

src/deepsparse/v2/middleware/abstract_middleware.py
```diff
@@ -0,0 +1,26 @@
...
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from abc import ABC, abstractmethod
+
+
+class AbstractMiddleware(ABC):
```
bfineran, line 19: let's just call this `Middleware` or `BaseMiddleware`
```suggestion
class Middleware(ABC):
```

src/deepsparse/v2/middleware/middleware_manager.py
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from typing import List, Optional
+
+from .abstract_middleware import AbstractMiddleware
```
bfineran, line 18: need to use absolute path to conform to repo ie `from deepsparse.v2.`

src/deepsparse/v2/middleware/abstract_middleware.py
```diff
@@ -0,0 +1,26 @@
...
+    @abstractmethod
+    def start_event(self, *args, **kwargs):
```
bfineran, line 21: we should have some required args here (and for end) as discussed such as `name, inputs, inference_state`
for end should also have `outputs`

src/deepsparse/v2/middleware/middleware_manager.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class MiddlewareManager:
+    def __init__(self, middleware: Optional[List[AbstractMiddleware]] = None):
+        self.middleware = middleware
+
```
bfineran, line 24: let's add an `add` method that just wraps `self.middleware.append`

src/deepsparse/v2/operators/operator.py
```diff
@@ -71,6 +71,16 @@ def __call__(
         :param kwargs: kwargs when not initializing from an instantiated schema
         :return: operator output
         """
+        middleware = inference_state.get_state("middleware")
+        if middleware is not None:
+            if "name" not in kwargs:
+                kwargs["name"] = self.__class__.__name__
+            middleware.start_event(
+                inferece_state=inference_state,
```
bfineran, line 79: named args need to go after inline`*args`

src/deepsparse/v2/operators/operator.py
```diff
@@ -97,8 +107,19 @@ def __call__(
                 inference_state=inference_state,
                 **kwargs,
             )
+
         if self.has_output_schema():
-            return self.output_schema(**run_output)
+            rtn = self.output_schema(**run_output)
+            if middleware is not None:
```
bfineran, line 113: we should also run this in the other case where there's no schema, using `run_output` as the `outputs`

src/deepsparse/v2/middleware/middleware_manager.py
```diff
@@ -0,0 +1,33 @@
...
+
+from typing import List, Optional
+
+from .abstract_middleware import AbstractMiddleware
+
+
+class MiddlewareManager:
+    def __init__(self, middleware: Optional[List[AbstractMiddleware]] = None):
+        self.middleware = middleware
```
bfineran, line 23: just for convienence, if we don't get a list, let's wrap it as a list for the user
```
self.middleware = middleware if isinstance(middleware, list) else [middleware]
```

src/deepsparse/v2/pipeline.py
```diff
@@ -258,11 +261,13 @@ def __call__(self, *args, **kwargs):
             inference_state = kwargs.pop("inference_state")
         else:
             inference_state = InferenceState()
-            inference_state.create_state({})
+            inference_state.create_state({"middleware": self._middleware})
 
         kwargs["inference_state"] = inference_state
 
-        return self.run(*args, **kwargs)
+        rtn = self.run(*args, **kwargs)
```
bfineran, line 268: why?

src/deepsparse/v2/pipeline.py
```diff
@@ -27,6 +27,7 @@
 from deepsparse.v2.utils import InferenceState, PipelineState
 from deepsparse.v2.utils.data import SubGraph
 from deepsparse.v2.utils.helpers import run_func
+from src.deepsparse.v2.middleware.abstract_middleware import AbstractMiddleware
 
 
 __all__ = ["Pipeline"]
```
bfineran, line 33: I think we'll also need to add `start_event` and `end_event` in the `__call__` method of pipeline since it doesn't call into the super method of `Operator`

additionally should think through if there's a way to mark the overall start and end of inference without worrying about nested pipelines

src/deepsparse/v2/utils/state.py
```diff
@@ -62,3 +63,21 @@ def update_value(self, attribute: str, value: Union[str, int, list]):
 
     def update_state(self, value: Any):
         self._current_state.update(value)
+
+    def get_state(self, key):
```
bfineran, line 67:
```suggestion
    def get(self, key):
```

tests/deepsparse/v2/middleware/test_middleware_manager.py
```diff
@@ -0,0 +1,52 @@
...
+    ].start_order
+    actual_middleware_end_order = AddThreePipeline._middleware.middleware[0].start_order
+
+    for expected, actual_start, actual_end in zip(
+        expected_middleware_start_order,
+        actual_middleware_start_order,
+        actual_middleware_end_order,
+    ):
+        assert expected == actual_start
+        assert expected == actual_end
```
bfineran, line 52: why not:
```suggestion
    assert (
        expected_middleware_start_order
        == actual_middleware_start_order
        == actual_middleware_end_order
    )
```

tests/deepsparse/v2/middleware/test_middleware_manager.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_middleware_triggered_in_expected_order_for_ops():
+    """Check that when ops gets triggered, middleware behaves as expected"""
+    AddThreePipeline = Pipeline(
+        ops=[AddOneOperator(), AddTwoOperator()],
+        router=LinearRouter(end_route=2),
+        schedulers=[OperatorScheduler()],
+        continuous_batching_scheduler=ContinuousBatchingScheduler,
+        middleware=[OpsTrackerMiddleware(), CounterMiddleware()],
```
bfineran, line 34: let's define these two middleware as objects before the constructor so we can reference them directly in the asserts (reads cleaner and verifies that pass by reference is used)

bfineran, line 34: where are we testing the results of `CounterMiddleware`?

tests/deepsparse/v2/test_basic_pipeline.py
```diff
@@ -23,7 +23,7 @@
 from deepsparse.v2 import Pipeline
 from deepsparse.v2.operators import Operator
 from deepsparse.v2.routers import LinearRouter
-from deepsparse.v2.schedulers import OperatorScheduler
+from deepsparse.v2.schedulers import ContinuousBatchingScheduler, OperatorScheduler
```
bfineran, line 26: is this your change?

bfineran (changes requested):

src/deepsparse/v2/middleware/middleware_manager.py
horheynm: I think looking at type annotations, the users should provide the list of middleware

src/deepsparse/v2/utils/state.py
horheynm: isnt get_state more consistent with the definitions of this class?

tests/deepsparse/v2/test_basic_pipeline.py
horheynm: Yes

src/deepsparse/v2/pipeline.py
```diff
@@ -57,13 +58,15 @@ def __init__(
...
         pipeline_state: PipelineState = None,
+        middleware: List[BaseMiddleware] = [],  # Initialized List of middlewares
     ):
 
         self.ops = ops
         self.router = router
         self.schedulers = schedulers
         self.pipeline_state = pipeline_state
         self._continuous_batching_scheduler = continuous_batching_scheduler
+        self._middleware = MiddlewareManager(middleware)
```
bfineran, line 69: let's make this public so users can call into `.add` easily

bfineran (approved):
LGTM, see final comment

PR #1443, opened 2023-11-30
V2 middleware timer

horheynm:
Add timer into middleware

examples in tests

Blocked on https://github.com/neuralmagic/deepsparse/pull/1437


src/deepsparse/v2/middleware/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# flake8: noqa
+
+from deepsparse.v2.middleware.base_middleware import BaseMiddleware
```
dsikka, line 17: can we update thse imports to use relative paths instead of absolute? the current imports cause a problem

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,52 @@
...
+import threading
+import time
+from typing import Dict, Optional
+
+from deepsparse.v2.utils.state import InferenceState
+
+from .base_middleware import BaseMiddleware
+
+
+class TimerMiddleware(BaseMiddleware):
```
dsikka, line 24: We have multiple operators being called multiple times. Would it be possible to store the average of these times instead of overwriting the times? This is needed to identify bottlenecks

src/deepsparse/v2/operators/operator.py
```diff
@@ -71,6 +71,18 @@ def __call__(
         :param kwargs: kwargs when not initializing from an instantiated schema
         :return: operator output
         """
+        name = None
```
dsikka, line 74: When we have multiple prompts, we copy the inference state for each prompt/subgraph execution. This allows each subgraph to update its specific subgraph, independent of the other subgraphs. We use the original inference state when we join the prompts and return to the normal single graph execution (i.e. not the sub-graph specific inference states) Since the middleware is stored in inference state, it seems like the individual operator timing will be off for this case?

dsikka (changes requested):

src/deepsparse/v2/middleware/__init__.py
horheynm: For which code?

PR #1446, opened 2023-12-01
[Text Generation] Terminate the inference when kv cache is full

dbogunowicz:
## Feature Description

Once the KV cache is full, instead of continuing the inference by removing the old cache entries to make place for the new ones, we now terminate the inference with the finish reason "capacity"

### Manual Testing

```python
from deepsparse import Pipeline
prompt = "James decides to run 3 sprints 3 times a week. He runs 60 meters each sprint. How many total meters does he run a week?"
model_path = "zoo:llama2-7b-gsm8k_llama2_pretrain-pruned80_quantized"
pipeline = Pipeline.create(task="text-generation", model_path=model_path, sequence_length=64)
out = pipeline(prompt=prompt)
```
#### Before:
displaying `out.generations[0].text` and `out.generations[0].finished_reason`:

```bash
text='He runs 60*3=<<60*3=180>>180 meters in total per sprint,  Comays 5  \nen3 was 2= a en6ound'
finished=True, finished_reason='max_new_tokens'
```

#### Now:
```bash
ext='He runs 60*3=<<60*3=180>>180 meters in total per s'
finished=True, finished_reason='capacity'
```

PR #1447, opened 2023-12-01
[Cherry-Pick][Text Generation] Terminate the inference when kv cache is full

dbogunowicz:
Cherry-pick for #1446


src/deepsparse/transformers/pipelines/text_generation.py
```diff
@@ -829,6 +829,11 @@ def engine_forward(
                     generated_tokens.append(token)
                     generated_logits.append(logits)
 
+                    if session.total_num_processed_tokens > session.capacity:
+                        # if the kv cache is full, stop generation
+                        finished_reason.append(FinishReason.CAPACITY)
+                        break
```
tlrmchlsmth, line 835: I patched this PR and there's an off-by-one error somewhere. A 65th token is being inserted into the KV cache.
Should this be 
```if session.total_num_processed_tokens >= session.capacity:```?

tlrmchlsmth (changes requested):

src/deepsparse/transformers/pipelines/text_generation.py
tlrmchlsmth: @SageMoore could you double check this on our end as well, any possibility that we're walking the cache an extra time or something similar?

dbogunowicz: No, you are right @tlrmchlsmth . It needs to be `>=`, just checked

dsikka (changes requested):
Could we add a test case to catch this case?

tlrmchlsmth:
> Could we add a test case to catch this case?

+1

PR #1450, opened 2023-12-03
Non-invasive middleware implementation for Ops and Pipeline

horheynm:
Timer middleware stuff will be created in a separated PR

```python3
(.venv) george@quad-mle-1:~/deepsparse$ pytest '/home/george/deepsparse/tests/deepsparse/v2/middleware/test_middleware.py'
=========================== test session starts ===========================
platform linux -- Python 3.10.13, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/george/deepsparse
configfile: pyproject.toml
plugins: flaky-3.7.0
collected 4 items                                                         

tests/deepsparse/v2/middleware/test_middleware.py ....              [100%]

============================ 4 passed in 0.86s ============================
(.venv) george@quad-mle-1:~/deepsparse$ 
```


src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, call_next: MiddlewareCallable, identifier: str):
+        self.identifier: str = identifier
+        self.call_next: MiddlewareCallable = call_next
+
+    def __call__(self, *args, **kwargs) -> Any:
+        start_time = time.time()
+        result = self.call_next(*args, **kwargs)
+        measurement = time.time() - start_time
+
+        self.send({"measurements": measurement})
```
horheynm, line 31: This is added here:
https://github.com/neuralmagic/deepsparse/pull/1450/files#diff-fef6bcdaa9b28ca7529b2c51a5c63196da4aac77972d9ff3690f2f24a251a1cdR61

dbogunowicz (approved):

src/deepsparse/v2/middleware/middlewares.py
```diff
@@ -0,0 +1,69 @@
...
+            for next_middleware, init_args in middleware:
+                next_middleware.send = self.recieve
+                self.middleware.append(MiddlewareSpec(next_middleware, **init_args))
+
+    def recieve(self, state: Dict):
+        with self._lock():
+            for key in state.keys():
+                if key not in self.state:
+                    self.state[key] = []
+                self.state[key].append(state[key])
```
tdg5, line 69: What do you think about approaching this like a redux reducer to make it so the state can be less rigid?

You'd still take a lock, but I guess the caller of `send` would call `send` with a callable instead of a `Dict` and then the `receive` function could call the callable w/ the current state
```suggestion
    def recieve(self, reducer: Callable[[Dict], Dict]):
        with self._lock():
            self.state = reducer(self.state)
```
That would also give the caller of `send` a chance to evaluate the data in `state`, in case that was helpful to it

src/deepsparse/v2/middleware/timer_middleware.py
```diff
@@ -0,0 +1,33 @@
...
+from deepsparse.v2.middleware.middlewares import MiddlewareCallable
+
+
+class TimerMiddleware(MiddlewareCallable):
+    def __init__(self, call_next: MiddlewareCallable, identifier: str):
+        self.identifier: str = identifier
+        self.call_next: MiddlewareCallable = call_next
+
+    def __call__(self, *args, **kwargs) -> Any:
+        start_time = time.time()
```
bfineran, line 27: let's rewrite a little to make it easier to get the fine grained timings we disussed,

```python
class TimerMiddleware(MiddlewareCallable):
    @contextmanager
    def time(self, id: str):
         start time ...
         yield
         end time ...
         self.send({id: time})

    def __call__(self, *args, **kwargs) -> Any:
        with self.time(id=kwargs["inferenence_state"].get("name")):
            result = self.call_next(*args, **kwargs)
        return result
```

then in middeware manager:

```python
class MiddlewareManager:
     @property
     def timer() -> Optional[TimerMiddleware]:
        ...
    @contextmanager
    def time(self, id):
       if self.timer:
            # do timer middleware time()
        else:
            # do nothing
```

that way in an operator we can do (may move middleware manager out of self

```python
class MyOperator():

    def run():
         with self.middleware_manager.time(id="specific_timing"):
              # do something
```

bfineran: let's move timer into a separate PR with this

src/deepsparse/v2/middleware/middlewares.py
horheynm: Thanks Danny!

PR #1451, opened 2023-12-04
[DeepSparse Evaluator API][UI] Rename the alias for lm-eval

dbogunowicz:


src/deepsparse/evaluation/registry.py
```diff
@@ -33,7 +33,12 @@ class EvaluationRegistry(RegistryMixin):
 
     @classmethod
     def load_from_registry(cls, name: str) -> Callable[..., "Result"]:  # noqa: F821
-        return cls.get_value_from_registry(name=name)
```
dbogunowicz, line 36: @bfineran let's discuss whether the `RegistryMixin` should be robust against hyphens and underscores so that we could do something like

```python
@Registry.register(name=["bat_man", "cool-cat"])
class A

assert registry.load_from_registry("bat-man")
assert registry.load_from_registry("cool_cat")
```

bfineran (commented):
LGTM pending base registry changes

bfineran (dismissed):

bfineran (approved):

rahul-tuli (approved):
Let's Go!

PR #1455, opened 2023-12-05
[Fix] Adapt to the newest SparseZoo interface (and get GHA green again) 

dbogunowicz:
Due to those changes: https://github.com/neuralmagic/sparsezoo/pull/389 we need to update the relevant pathways.

### DIFF
--- a/src/deepsparse/utils/onnx.py
+++ b/src/deepsparse/utils/onnx.py
@@ -129,7 +129,7 @@ def model_to_path(model: Union[str, Model, File]) -> str:
 
     if Model is not object and isinstance(model, Model):
         # trigger download and unzipping of deployment directory if not cached
-        model.deployment.download()
+        model.deployment.path
 
         # default to the main onnx file for the model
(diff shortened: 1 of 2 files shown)


mgoin (dismissed):

src/deepsparse/utils/onnx.py
```diff
@@ -129,7 +129,7 @@ def model_to_path(model: Union[str, Model, File]) -> str:
 
     if Model is not object and isinstance(model, Model):
         # trigger download and unzipping of deployment directory if not cached
-        model.deployment_directory_path
+        model.deployment.download()
```
bfineran, line 132: Why do we want download instead of path? Would this trigger redownload?

dbogunowicz: Yeah, I agree

dbogunowicz, line 132:
```suggestion
        model.deployment.path
```

Satrat (approved):

bfineran (approved):

PR #1463, opened 2023-12-08
dummy injection

horheynm:
python3 scratch/dep.py


src/deepsparse/dependency_injector/services.py
```diff
@@ -0,0 +1,41 @@
...
+    
+    def measurements(self):
+        return self.timer.measurements
+    
+    def start_times(self):
+        return seld.timer.start_times
+    
+    @contextmanager
+    def record(self, id: str):
+        self.start(id)
```
horheynm, line 39: take the timings out

PR #1469, opened 2023-12-12
[V2  Pipeline] Fine-grained timer from inference state

horheynm:
## Description
Adds ability to record run times in Pipeline and Ops, given inference state

## Problem
Currently, we dont have a way to get the run time info in the pipeline and in the operator

## Solution 
We add timer manager as an attribute to the Pipeline, and add Timer logic into the InferenceState as an attribute (see tests)

## Design
We re use the timer attribute in the inference state. All run-times will be saved there and will be written to the timer manager once the Pipeline run ends.

## Usage
Given access to inference_state
```python3
with inference_state.time(id="foo"):
   time.sleep(1)
```

## Testing
1. Checks that the runtime from ops and Pipeline is saved into the timer manager
2. multi-threaded  check
The runtime without saving the measurements and the runtime saving the measurements are checked to be same
3. The expected time that is saved is correct
```
no timer, no wait : 2.95375657081604
no timer, 0.2 wait: 146.53509402275085
pipeline timer, no wait : 3.020995855331421, 2.8153765201568604 
pipeline timer, 0.2 wait: 150.66221618652344
pipeline + ops timer   ,    no wait : 2.9319729804992676 3.0496866703033447 3.1929404735565186
pipeline + ops timer,       0.2 wait: 154.01117300987244
```
```python3
model_path = "hf:neuralmagic/TinyLlama-1.1B-Chat-v0.4-pruned50-quant-ds"

pipeline = Pipeline.create(
    task="text_generation",
    model_path=model_path,
    engine_type="deepsparse", # onnxruntime
    internal_kv_cache=True, # False
    #
)
import time 
start =time.time()
output = pipeline(
    # prompt= ["The sun shined"],
    prompt= ["The sun shined", "Oh hello!"],
    generation_kwargs= {
        "num_return_sequences": 4,
        "do_sample": True,
        "max_length": 20,
    },
)
rt = time.time() - start
print(rt) 
breakpoint()
```


src/deepsparse/pipeline.py
```diff
@@ -328,6 +331,9 @@ def run(
                 inference_state = graph.inf
                 next_step = graph.step
 
+        timer = inference_state.get_state("timer")
```
bfineran, line 334: let's add a timer property to inference state as discussed to handle the optional timer

src/deepsparse/pipeline.py
```diff
@@ -328,6 +331,9 @@ def run(
...
+        if timer is not None:
+            self.timer_manager.update(timer.measurements)
```
bfineran, line 336: do we need this if we update in line 356 anyways?

src/deepsparse/pipeline.py
```diff
@@ -342,11 +348,17 @@ def __call__(self, *args, **kwargs):
             inference_state = kwargs.pop("inference_state")
         else:
             inference_state = InferenceState()
-            inference_state.create_state({})
+            inference_state.create_state({**{"timer": self.timer_manager.get_new_timer()}})
```
bfineran, line 351:
```suggestion
            inference_state.create_state(dict(timer=self.timer_manager.get_new_timer()))
```

src/deepsparse/utils/state.py
```diff
@@ -43,9 +46,23 @@ def create_state(self, new_state: dict):
...
+        
+class TimerState:
+    def __init__(self, timer: Optional[Timer] = None):
+        self.timer = timer
+        
+    @contextmanager
+    def record(self, id: str):
+        if self.timer is not None:
+            return self.timer.record(id=id)
+        yield
```
bfineran, line 58: add in-line comment that this is a null context

src/deepsparse/utils/state.py
```diff
@@ -43,9 +46,23 @@ def create_state(self, new_state: dict):
         if self._current_state:
             raise ValueError("State creation is only allowed during initialization.")
         self._current_state = new_state
+        
+class TimerState:
+    def __init__(self, timer: Optional[Timer] = None):
+        self.timer = timer
+        
+    @contextmanager
+    def record(self, id: str):
```
bfineran, line 55: prefer to rename this to `time()` especially since it's being called from inference state directly

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,48 @@
...
+import time
+from contextlib import contextmanager
+import threading
+from typing import Dict
+
+class Timer:
+    def __init__(self):
+        self._lock = threading.Lock()
+        self.start_time = {}
+        self.measurements = {}
```
bfineran, line 11:
```suggestion
        self.measurements = defaultdict(list)
```

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,48 @@
...
+                start_time = self.start_time[id]
+                del self.start_time[id]
+                self.measurements[id] = end_time - start_time
+
+    @contextmanager
+    def record(self, id: str):
+        start = time.time()
+        yield
+        with self._lock:
+            self.measurements[id] = time.time() - start
```
bfineran, line 36:
```suggestion
            self.measurements[id].append(time.time() - start)
```

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,48 @@
...
+    ):
+        end_time = time.time()
+        with self._lock:
+            if id in self.start_time:
+                start_time = self.start_time[id]
+                del self.start_time[id]
+                self.measurements[id] = end_time - start_time
+
+    @contextmanager
+    def record(self, id: str):
```
bfineran, line 32: let's rename this to `time` as well

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,48 @@
...
+        start = time.time()
+        yield
+        with self._lock:
+            self.measurements[id] = time.time() - start
+
+class TimerManager:
+    def __init__(self):
+        self.lock = threading.Lock()
+        self.measurements = {}
```
bfineran, line 41: this should be a list

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,48 @@
...
+        
+    def get_new_timer(self):
+        return Timer()
+    
+    def update(self, measurements: Dict[str, float]):
+        with self.lock:
+            self.measurements.update(measurements)
```
bfineran, line 48: append

src/deepsparse/operators/operator.py
```diff
@@ -71,35 +71,41 @@ def __call__(
         :param kwargs: kwargs when not initializing from an instantiated schema
         :return: operator output
         """
-        if self.has_input_schema():
-            if len(args) > 1:
-                raise ValueError(
-                    f"The operator requires an {self.input_schema}. Too many arguments"
-                    "provided."
+
+        timer = empty_context_manager
```
bfineran, line 75: let's remove this in favor of the timer middleware

bfineran (changes requested):

src/deepsparse/pipeline.py
```diff
@@ -77,6 +77,7 @@ def __init__(
         self.schedulers = schedulers
         self.pipeline_state = pipeline_state
         self._continuous_batching_scheduler = continuous_batching_scheduler
+        self.timer_manager = TimerManager()
```
bfineran, line 80: we should ideally make timer manager and timings optional

bfineran: additionally, as discussed before we should follow up with the default ability to not keep a history of timings

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def get_new_timer(self):
+        return Timer()
+
+    def update(self, measurements: Dict[str, float]):
+        with self.lock:
+            self.measurements.append(measurements)
+
+    def average_times(self):
+        ...
```
bfineran, line 51: let's remove for this PR
```suggestion
```

bfineran (approved):

bfineran (approved):

bfineran (approved):

src/deepsparse/utils/state.py
```diff
@@ -62,3 +92,24 @@ def update_value(self, attribute: str, value: Union[str, int, list]):
 
     def update_state(self, value: Any):
         self._current_state.update(value)
+
+    def get_state(self, key):
+        if key in self.current_state:
```
rahul-tuli, line 97: docstring?

src/deepsparse/utils/state.py
```diff
@@ -62,3 +92,24 @@ def update_value(self, attribute: str, value: Union[str, int, list]):
...
+            return self.current_state[key]
+
+    def copy_state(self, props=["timer"]):
+        """copy everything except the attrs in props"""
```
rahul-tuli, line 101: add args?

src/deepsparse/utils/state.py
```diff
@@ -62,3 +92,24 @@ def update_value(self, attribute: str, value: Union[str, int, list]):
 
     def update_state(self, value: Any):
         self._current_state.update(value)
+
+    def get_state(self, key):
+        if key in self.current_state:
+            return self.current_state[key]
+
+    def copy_state(self, props=["timer"]):
```
rahul-tuli, line 100: Won't a shallow copy have the same effect as this function?

src/deepsparse/utils/time.py
```diff
@@ -0,0 +1,48 @@
...
+        start = time.time()
+        yield
+        with self._lock:
+            self.measurements[id].append(time.time() - start)
+
+
+class TimerManager:
+    def __init__(self):
+        self.lock = threading.RLock()
+        self.measurements = []
```
rahul-tuli, line 41: measurements is a list of dicts with timings right? could add typing to this variable to make it clearer?

bfineran:
LGTM pending @rahul-tuli's comments

src/deepsparse/utils/state.py
horheynm: Doc strings are inconsistent in this repo. Here above, update_state dont even have any.
Ill address this in the stand up

src/deepsparse/utils/state.py
horheynm: Every InferenceState may have their own context, so you can't share it with other Operator.
The only thing we share is the Timer.

bfineran (approved):

dsikka (commented):
These changes LGTM but the pipeline code has changed. Can we merge in main and make sure everything runs as you'd expect? If all good, can merge in.

src/deepsparse/utils/state.py
```diff
@@ -45,14 +49,40 @@ def create_state(self, new_state: dict):
         self._current_state = new_state
 
 
-class InferenceState(State):
+class TimerState:
+    """TimerState shared among all InferenceState"""
+
+    _timer = None
```
bfineran, line 55: why is this static and not a class property?

bfineran: we don't want to share the timer between inference states. each new inference state (outside of copies ofc) should have their own timer so that the timer can hold the specific timings for a single inference

bfineran (approved):

PR #1472, opened 2023-12-13
V2 Timer Manager

horheynm:
Blocked on https://github.com/neuralmagic/deepsparse/pull/1471

## Description
Add Timer to middleware, which will record the runtime of __call__ in Pipeline and Operator

## Problem
No existing way to get the run time of the callables

## Solution 
Add TimerMiddleware to the existing middleware

## Design
General middleware implementation. Record start time, run the callable, get the end time and the overall run-time

## Usage
Check tests (`/tests/deepsparse/middlewares/test_timer_middleware_pipeline.py`)
```python3

 middlewares = [
        MiddlewareSpec(TimerMiddleware),  # for timer
    ]

    ops = [AddOneOperator(), AddTwoOperator()]

    AddThreePipeline = Pipeline(
        ops=ops,
        router=LinearRouter(end_route=2),
        continuous_batching_scheduler=ContinuousBatchingScheduler,
        middleware_manager=MiddlewareManager(middlewares),
    )

    pipeline_input = IntSchema(value=5)
    pipeline_output = AddThreePipeline(pipeline_input) # output is 8

    pipeline_measurements: List[
        defaultdict
    ] = AddThreePipeline.timer_manager.measurements
    measurements = pipeline_measurements[0]


>> measurements
defaultdict(<class 'list'>, {'AddOneOperator': [9.202957153320312e-05], 'AddTwoOperator': [3.9577484130859375e-05], 'total': [0.0009074211120605469]})
```

## Testing
Checks if timings are measured, and exists as expected


src/deepsparse/middlewares/timer_middleware.py
```diff
@@ -0,0 +1,40 @@
...
+
+from deepsparse.middlewares.middleware import MiddlewareCallable
+
+
+class TimerMiddleware(MiddlewareCallable):
+    def __init__(self, call_next: MiddlewareCallable, identifier: str):
+        self.identifier: str = identifier
+        self.call_next: MiddlewareCallable = call_next
+
+    def __call__(self, *args, **kwargs) -> Any:
```
bfineran, line 26: we should be grabbing the `timer` from inference state, using that to time, and then using those results that are already written for us to timer manager.

no need for send here

bfineran (changes requested):

src/deepsparse/middlewares/timer_middleware.py
```diff
@@ -0,0 +1,32 @@
...
+class TimerMiddleware(MiddlewareCallable):
+    def __init__(
+        self, call_next: MiddlewareCallable, identifier: str = "TimerMiddleware"
+    ):
+        self.identifier: str = identifier
+        self.call_next: MiddlewareCallable = call_next
+
+    def __call__(self, *args, **kwargs) -> Any:
+        name = kwargs.get("name")
+        inference_state = kwargs.get("inference_state")
```
bfineran, line 29: need to return if inference state is not found

src/deepsparse/pipeline.py
```diff
@@ -405,13 +413,25 @@ def __call__(self, *args, **kwargs):
...
 
         next_call = self.run
         if self.middleware_manager is not None:
+            # make next calls to be middlewares if any
             next_call = self.middleware_manager.build_middleware_stack(next_call)
 
-        return next_call(*args, **kwargs)
+        kwargs["inference_state"] = inference_state
+        # add name for timer measurements key
+        kwargs["name"] = "total"
```
bfineran, line 426: if a pipeline is nested we will get conflicts on the `"total"` name - need a way to resolve this (ie if timer already exists in inference state we know its a nested pipeline)

src/deepsparse/pipeline.py
```diff
@@ -405,13 +413,25 @@ def __call__(self, *args, **kwargs):
             inference_state = InferenceState()
             inference_state.create_state({})
 
-        kwargs["inference_state"] = inference_state
+            timer = self.timer_manager.get_new_timer()
```
bfineran, line 416: if we have a nested pipeline, then we don't want to make a new timer here

src/deepsparse/utils/time.py
```diff
@@ -49,3 +55,18 @@ def get_new_timer(self):
     def update(self, measurements: Dict[str, float]):
         with self.lock:
             self.measurements.append(measurements)
+
+    def average(self):
+        """Get the average time for each key in every element inside measurements"""
+
+        summary = {"time": defaultdict(float), "execution": defaultdict(int)}
```
bfineran, line 62: need a better name for "execution" to make it clear that it's the number of times an operator is executed per inference. also let's add an `average_times` function as well that just returns the "time" dict

src/deepsparse/utils/time.py
```diff
@@ -49,3 +55,18 @@ def get_new_timer(self):
...
+
+        for measurement in self.measurements:
+            for key, values in measurement.items():
+                summary["time"][key] += sum(values)
+                summary["execution"][key] += len(values)
+
+        for key in summary["time"]:
+            summary["time"][key] /= summary["execution"][key]
```
bfineran, line 70: shouldn't execution be averaged as well by the number of inferences?

src/deepsparse/utils/time.py
```diff
@@ -49,3 +55,18 @@ def get_new_timer(self):
...
+
+        return {metric: dict(values) for metric, values in summary.items()}
```
bfineran, line 72: why do we need to wrap values in a dict call

mgoin:
could you please make the TimerManager in time.py have a `__repr__()` defined like [the one in timer.py](https://github.com/neuralmagic/deepsparse/blob/9ddeb61d0c80dcdab7b4737ae8aff14fbaf5ed53/src/deepsparse/utils/timer.py#L258-L264)? 
this will let it have a useful string representation since currently it just prints the raw object:
```
(Pdb) model.timer_manager
<deepsparse.utils.time.TimerManager object at 0x7fd9c6987be0>
```

bfineran (approved):

tests/deepsparse/middlewares/test_timer_middleware_pipeline.py
```diff
@@ -0,0 +1,304 @@
...
+        MiddlewareSpec(TimerMiddleware),  # for timer
+    ]
+
+    model = TextGeneration(
+        model_path="hf:mgoin/TinyStories-1M-ds",
+        middleware_manager=MiddlewareManager(middlewares),
+    )
+
+    prompt = "How to make banana bread?"
+    generation_config = {"max_new_tokens": 100}
```
mgoin, line 222:
```suggestion
    generation_config = {"max_new_tokens": 10}
```

mgoin (approved):
works with my tests, thanks!

bfineran (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -90,6 +90,7 @@ def __init__(
         self.schedulers = schedulers
         self.pipeline_state = pipeline_state
         self._continuous_batching_scheduler = continuous_batching_scheduler
+        self.timer_manager = TimerManager()
```
dsikka, line 93: Im seeing two time_managers here? Line 93 and line 95.

src/deepsparse/pipeline.py
```diff
@@ -178,63 +179,73 @@ async def run_async(self, *args, inference_state: InferenceState, **kwargs):
...
-                )
-                next_step = self.router.JOIN_ROUTE
-
-            else:
-                if next_step == self.router.START_ROUTE:
-                    outputs = self.run_func(
-                        *args,
-                        func=self._scheduler_group.submit,
-                        operator=self.ops[next_step],
+        with inference_state.time(id="total"):
```
dsikka, line 189: would it be possible for us to be consistent with the existing timer? I believe we store `total_inference` with that one but you'd have to double check. Otherwise, the benchmark pipeline will have a lot of custom code for v2.

src/deepsparse/pipeline.py
```diff
@@ -406,17 +419,29 @@ def __call__(self, *args, **kwargs):
...
+        kwargs["inference_state"] = inference_state
+        # add name for timer measurements key
+        kwargs["name"] = "total"
+        kwargs["is_nested"] = is_nested
 
         # timer shared across all operators, has all measurements
         timer = inference_state.timer
 
+        next_call = self.run
+        if self.middleware_manager is not None:
```
dsikka, line 440: these lines repeat what we have on line 427 - is that intended?
Also, it seems like some of the steps being done here are not done in the `aysnc` pathway?

dsikka (changes requested):

src/deepsparse/pipeline.py
horheynm: Good catch. One is a duplicate.

Yes async pathway does not have a `total`. The entrypoint is run_async -> trigger operators, whereas regular entrypoint is __call__ -> run -> trigger ops, so can wrap in __call__.

src/deepsparse/pipeline.py
horheynm: THis is from the design doc, feel free to change if youd like. shouldnt be blocking merge

src/deepsparse/pipeline.py
horheynm: Yes good catch

src/deepsparse/pipeline.py
dsikka: Does that mean the async pathway wont be able store timings?

src/deepsparse/pipeline.py
dsikka: We'll have to make it clear then that this isn't compatible with the `benchmark_pipeline`

bfineran: `total_inference` is fine

bfineran (approved):

dsikka (approved):

PR #1473, opened 2023-12-13
[Text Generation][Pipeline Refactor] Add kv_cache session full check

dsikka:
# Summary
- This commit happened while v2 was being migrated so adding it back to the new pipeline
- `test_text_generation.py` updated to use the new pipeline; skipping tests which are missing features


dbogunowicz (approved):
Thanks for that!

bfineran (approved):

PR #1476, opened 2023-12-13
[Text Generation][Pipeline Refactor] Causal Mask Check 

dsikka:
# Summary
- Add in causal mask check


dbogunowicz (approved):

bfineran (approved):
LGTM - wonder if there's a way we can keep some graph state so we don't need to re-load the model to check

PR #1487, opened 2023-12-19
Add Text Gen Alias

dsikka:


rahul-tuli (approved):

dbogunowicz:
Well...I know that this is an attempt to bring back the same functionalities as in the v1, but let's be realistic about it. I'd definitely not put bloom in there, because we do not support it, I'd much rather put MPT and llama / llama2 in the alias, since this is what the people are most likely to use.

dsikka:
I can remove bloom but if we still seem to support kv_cache injection for it?

dbogunowicz:
@dsikka but its not in sparsezoo and we do not treat it as productionized in our LLM portfolio.

bfineran (approved):

mgoin (approved):

PR #1493, opened 2023-12-19
[Text Generation] Enable continuous batching when internal kv_cache is enabled

dsikka:
# Summary
- Remove the check in `text_generation/pipeline.py` which  disables continuous batching when internal kv cache is enabled
- Update `NLEngineOperator` to use the engine provided as an input, if available, when using internal kv cache. This engine will have the model with the updated batch size, when continuous batching is being used (this is true in general for continuous batching)
- Update step when engines are created within the `continuous_batching_scheduler` specific for the `NLEngineOperator`

### DIFF
--- a/src/deepsparse/schedulers/continuous_batching_scheduler.py
+++ b/src/deepsparse/schedulers/continuous_batching_scheduler.py
@@ -17,7 +17,7 @@ from concurrent.futures import Future
 from threading import Lock
 from typing import List
 
-from deepsparse.operators import EngineOperator, Operator
+from deepsparse.operators import DEEPSPARSE_ENGINE, EngineOperator, Operator
 from deepsparse.schedulers.scheduler import OperatorScheduler
 from deepsparse.schedulers.utils import (
     ContinuousBatchingExecutorThread,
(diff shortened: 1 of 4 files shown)


src/deepsparse/schedulers/continuous_batching_scheduler.py
```diff
@@ -167,16 +167,30 @@ def add_engine_operator(
...
 
             override_model_path = None
-            # text generation/NLEngineOperator specific; could add generic method
-            # for all engine_operators, if desired
-            if hasattr(engine_operator, "override_model_inputs"):
-                override_model_path = engine_operator.override_model_inputs(
-                    model_path=engine_operator.model_path, batch_size=batch_size
+            kwargs = {}
+
+            if engine_operator.__class__.__name__ == "NLEngineOperator":
```
bfineran, line 172: can any of this logic be moved into the operators itself? just worried about tying in the operator class too much

bfineran (dismissed):

src/deepsparse/schedulers/continuous_batching_scheduler.py
dsikka: so all of this is needed for engine creation which happens in the operator's init function for batch_size 1. We could move these functions outside of the engine init, override the engine'e `create_engine` method, and add these steps to it.

src/deepsparse/schedulers/continuous_batching_scheduler.py
```diff
@@ -167,16 +167,30 @@ def add_engine_operator(
...
 
-            # will break for internal kv_cache; needs additional argument
+                if (
+                    engine_operator._engine_type == DEEPSPARSE_ENGINE
+                    and engine_operator.internal_kv_cache
+                ):
+                    output_indices_to_be_cached = additional_outputs.get(
+                        "output_indices_to_be_cached"
+                    )
+                    if any(output_indices_to_be_cached):
```
dbogunowicz, line 189: in which scenario `any(output_indices_to_be_cached)==False` given that
` (engine_operator._engine_type == DEEPSPARSE_ENGINE and engine_operator.internal_kv_cache)== True`
?

src/deepsparse/transformers/pipelines/text_generation/nl_engine_operator.py
```diff
@@ -208,14 +209,15 @@ def run(self, inp: NLEngineInputs, **kwargs) -> NLEngineOutputs:
             # (batch_size=0) to the engine. Therefore,
             # we skip the validation
 
-            # Internal kv_cache works for batch_size of 1 atm
-            out = self.engine._eng_net.execute_list_out(
-                inputs[0], kv_cache[0].engine_internal_cache
-            )
+            internal_kv_cache = [x.engine_internal_cache for x in kv_cache]
+            if inp.engine:
```
dbogunowicz, line 213: `inp.engine`? As in the engine is an attribute of the input and not the operator itself? When does this happen?

dsikka: When we introduced continuous_batching, this was added in so that we could use the engine which has the model with the correct batch size

src/deepsparse/schedulers/continuous_batching_scheduler.py
dsikka: this condition is from v1. if they're not mutually exclusive, I can combine

mgoin (dismissed):

dsikka:
I cleaned this up so let's hold on merging this in until tomorrow

src/deepsparse/schedulers/continuous_batching_scheduler.py
dsikka: Done.

bfineran (approved):

dbogunowicz (approved):

PR #1499, opened 2023-12-21
Replace Fastchat Chat Templates with HF `apply_chat_template`

mgoin:
Inspired by https://github.com/vllm-project/vllm/pull/1365 and instigated by prefill benchmarking, using the HF Chat Templates hidden within the model's `tokenizer` seems to be a more elegant approach.

From a performance perspective, fastchat's default template adds over 400 tokens as a preamble to chat conversations, **without actually using the right prompt format the model was trained on**! Example below:
```
### Human: Got any creative ideas for a 10 year old’s birthday?
### Assistant: Of course! Here are some creative ideas for a 10-year-old's birthday party:
1. Treasure Hunt: Organize a treasure hunt in your backyard or nearby park. Create clues and riddles for the kids to solve, leading them to hidden treasures and surprises.
2. Science Party: Plan a science-themed party where kids can engage in fun and interactive experiments. You can set up different stations with activities like making slime, erupting volcanoes, or creating simple chemical reactions.
3. Outdoor Movie Night: Set up a backyard movie night with a projector and a large screen or white sheet. Create a cozy seating area with blankets and pillows, and serve popcorn and snacks while the kids enjoy a favorite movie under the stars.
4. DIY Crafts Party: Arrange a craft party where kids can unleash their creativity. Provide a variety of craft supplies like beads, paints, and fabrics, and let them create their own unique masterpieces to take home as party favors.
5. Sports Olympics: Host a mini Olympics event with various sports and games. Set up different stations for activities like sack races, relay races, basketball shooting, and obstacle courses. Give out medals or certificates to the participants.
6. Cooking Party: Have a cooking-themed party where the kids can prepare their own mini pizzas, cupcakes, or cookies. Provide toppings, frosting, and decorating supplies, and let them get hands-on in the kitchen.
7. Superhero Training Camp: Create a superhero-themed party where the kids can engage in fun training activities. Set up an obstacle course, have them design their own superhero capes or masks, and organize superhero-themed games and challenges.
8. Outdoor Adventure: Plan an outdoor adventure party at a local park or nature reserve. Arrange activities like hiking, nature scavenger hunts, or a picnic with games. Encourage exploration and appreciation for the outdoors.
Remember to tailor the activities to the birthday child's interests and preferences. Have a great celebration!
### Human:
```

All of the above is just a default preamble, and then it proceeds to contain it within a `### Human: ... ### Assistant:` format is not correct for our tested `hf:neuralmagic/TinyLlama-1.1B-Chat-v0.4-pruned50-quant-ds`, which prefers `<s>[INST] ... <<SYS>> ... <</SYS>> ... [/INST]`. HF `tokenizer.apply_chat_template` has no problem generating that right format and integrating the roles properly in the same list of dictionaries format.

### DIFF
--- a/src/deepsparse/server/openai_server.py
+++ b/src/deepsparse/server/openai_server.py
@@ -102,8 +102,17 @@ class OpenAIServer(Server):
                 )
 
             try:
+                messages = request.messages
+                # For chat templating, the message needs to be formatted
+                # as a list of dictionaries of `{"role": "", "content": ""}`
+                # https://huggingface.co/docs/transformers/chat_templating
+                if isinstance(messages, str):
+                    messages = [{"role": "user", "content": messages}]
+                elif isinstance(messages, dict):
+                    messages = [messages]
+
                 prompt = pipeline.tokenizer.apply_chat_template(
-                    conversation=request.messages,
+                    conversation=messages,
                     add_generation_prompt=request.add_generation_prompt,
                     tokenize=False,
                 )
--- a/src/deepsparse/server/openai_server.py
+++ b/src/deepsparse/server/openai_server.py
@@ -96,22 +96,21 @@ class OpenAIServer(Server):
             if isinstance(request.messages, str):
                 prompt = request.messages
             else:
-                # else case assums a FastChat-compliant dictionary
+                # else case assumes a FastChat-compliant dictionary
                 # Fetch a model-specific template from FastChat
-                _LOGGER.warning(
-                    "A dictionary message was found. This dictionary must "
-                    "be fastchat compliant."
-                )
                 try:
                     from fastchat.model.model_adapter import get_conversation_template
                 except ImportError as e:
-                    return create_error_response(HTTPStatus.FAILED_DEPENDENCY, str(e))
+                    return create_error_response(
+                        HTTPStatus.FAILED_DEPENDENCY,
+                        f"{str(e)} - Please ensure `fastchat` is installed.",
+                    )
 
                 conv = get_conversation_template(request.model)
                 messages = request.messages
                 messages = messages if isinstance(messages, list) else [messages]
+                # add the model to the Conversation template, based on the given role
                 for message in messages:
-                    # add the model to the Conversation template, based on the given role
                     msg_role = message["role"]
                     if msg_role == "system":
                         conv.system_message = message["content"]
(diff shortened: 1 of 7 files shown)


rahul-tuli (approved):

src/deepsparse/server/openai_server.py
```diff
@@ -93,49 +93,26 @@ async def create_chat_completion(raw_request: Request):
...
             if not pipeline:
                 return create_error_response(
-                    HTTPStatus.BAD_REQUEST, f"{model} is not available"
+                    HTTPStatus.BAD_REQUEST,
+                    f"The model `{model}` does not exist.",
+                )
+
+            try:
+                prompt = pipeline.tokenizer.apply_chat_template(
+                    conversation=request.messages,
```
dsikka, line 106: This works for both a `dictionary` and a `string`? any difference in how they're both treated?

src/deepsparse/server/openai_server.py
```diff
@@ -93,49 +93,26 @@ async def create_chat_completion(raw_request: Request):
...
+                    add_generation_prompt=request.add_generation_prompt,
+                    tokenize=False,
```
dsikka, line 108: what does this do?

src/deepsparse/server/openai_server.py
mgoin: good catch, it definitely does not deal with `str` properly, so vLLM has a bug there :)

src/deepsparse/server/openai_server.py
mgoin: It specifies whether the chat template application should add the start of an assistant message to the end of the prompt. vLLM does it by default and this is usually how chat templates end in our examples, so i thought it'd make sense to continue that

for instance here is a tinyllama example from its model card where you can see the `<|im_start|>assistant\n` ending:
```python
from deepsparse import TextGeneration

prompt = "How to make banana bread?"
formatted_prompt =  f"<|im_start|>user\n{prompt}<|im_end|>\n<|im_start|>assistant\n"

model = TextGeneration(model="hf:neuralmagic/TinyLlama-1.1B-Chat-v0.4-pruned50-quant-ds")
print(model(formatted_prompt, max_new_tokens=500).generations[0].text)
```

dsikka (dismissed):
LGTM pending merge conflicts fixed

bfineran (approved):

PR #1500, opened 2023-12-27
Add CLIP model to enable test_clip.py

mgoin:
Since we've made a quantized CLIP (https://huggingface.co/neuralmagic/CLIP-ViT-B-32-256x256-DataComp-s34B-b86K-quant-ds), let's use it to test our pipeline!

### DIFF
--- a/tests/deepsparse/pipelines/test_clip.py
+++ b/tests/deepsparse/pipelines/test_clip.py
@@ -43,8 +43,9 @@ def custom_process_inputs(self, inputs):
     return [tokens, tokens_lengths]
 
 
-# This overrides the process_inputs function globally for all CLIPTextPipeline classes
-# This is needed for CLIP-ViT-B-32-256x256-DataComp-s34B-b86K
+# This overrides the process_inputs function globally for all CLIPTextPipeline classes.
+# This is needed for CLIP-ViT-B-32-256x256-DataComp-s34B-b86K since it has a second input
+# that specifies how many tokens are present.
 CLIPTextPipeline.process_inputs = custom_process_inputs
 
 
--- a/tests/deepsparse/pipelines/test_clip.py
+++ b/tests/deepsparse/pipelines/test_clip.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import pytest
+import numpy as np
 from deepsparse.clip import (
     CLIPCaptionInput,
     CLIPCaptionPipeline,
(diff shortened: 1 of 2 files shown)


dsikka (changes requested):
Will this model work just for classification or for captioning as well? From the OpenCLIP models available, only a small fraction worked for captioning. 

Can we add a description of the available models on sparsezoo somewhere? Either in a read me or docstring? Should make it clear how/if this differs from OpenCLIP

mgoin:
@dsikka it is just made for image/text retrieval and zero-shot image classification. I left in the pytest.skip for the captioning test as a result. 
This model was made from [OpenCLIP using a special branch](https://github.com/neuralmagic/open_clip/tree/jen/clip) that @ohaijen worked on. This was made as a collaboration so we're focused on getting results and pipelines working quickly - this is why we pushed to HF and everything to run the model is pretty much self-contained within that model card and notebooks.
My opportunistic thinking here was that we could use the model we're using on HF for some active testing on DeepSparse. I don't think there are plans on pushing CLIP models to SparseZoo at the moment, but if you think it's necessary for testing we could try to get a model up there.

dbogunowicz (approved):

tests/deepsparse/pipelines/test_clip.py
```diff
@@ -30,24 +32,48 @@
...
+    model_path = model_folder + "/textual.onnx"
     text = ["a building", "a dog", "a cat"]
     return CLIPTextInput(text=text), model_path
 
 
-@pytest.mark.skip(reason="No CLIP models currently available to run tests")
 @mock_engine(rng_seed=0)
 def test_visual_clip(engine, visual_input):
-    from deepsparse import Pipeline
+    from deepsparse.legacy import Pipeline
```
dsikka, line 76: shouldnt be needed?

tests/deepsparse/pipelines/test_clip.py
```diff
@@ -57,10 +83,9 @@ def test_visual_clip(engine, visual_input):
     assert len(output.image_embeddings) == 1
 
 
-@pytest.mark.skip(reason="No CLIP models curently available to run tests")
 @mock_engine(rng_seed=0)
 def test_text_clip(engine, text_input):
-    from deepsparse import Pipeline
+    from deepsparse.legacy import Pipeline
```
dsikka, line 88: same comment as above.

tests/deepsparse/pipelines/test_clip.py
```diff
@@ -30,24 +32,48 @@
...
+# that specifies how many tokens are present.
+CLIPTextPipeline.process_inputs = custom_process_inputs
+
+
 @pytest.fixture
-def visual_input():
+def model_folder():
+    from huggingface_hub import snapshot_download
+
+    model_id = "neuralmagic/CLIP-ViT-B-32-256x256-DataComp-s34B-b86K-quant-ds"
```
dsikka, line 56: Could we add a quick comment/note indicating that this model is not from OpenCLIP and only use for zero-shot classification?

dsikka (changes requested):
A few small comments otherwise LGTM

jeanniefinks:

PR #1504, opened 2024-01-04
[TextGeneration] Set scary logs to `debug`

dsikka:
- Two logs that were initially set to `info` are now changed to be `debug`
This will address the following ticket: https://app.asana.com/0/1205229323407165/1206269382548635/f


rahul-tuli (approved):
GG!

bfineran (approved):

PR #1508, opened 2024-01-04
[Server] Use the integration field when starting a server

dsikka:
# Summary
- The integration value defined in the server config was being ignored and the cli argument was only being considered
- Update to use both
- Also update tests and small clean-up
- For this ticket: https://app.asana.com/0/1205229323407165/1206280777419100/f

# Testing
- All server tests pass

The following config file will create an openai server:

```yaml
num_workers: 1
num_streams: 1
integration: openai
endpoints:
  - task: text_generation
    model: "hf:mgoin/TinyStories-1M-ds"
    kwargs:
      {"continuous_batch_sizes": [4, 8, 16], "force_max_tokens": True}
```

### DIFF
--- a/src/deepsparse/server/config.py
+++ b/src/deepsparse/server/config.py
@@ -166,7 +166,7 @@ class ServerConfig(BaseModel):
 
     integration: str = Field(
         default=None,
-        description="The kind of integration to use. local|sagemaker|openai",
+        description=f"The kind of integration to use. {INTEGRATIONS}",
     )
 
     engine_thread_pinning: str = Field(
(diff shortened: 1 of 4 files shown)


src/deepsparse/server/config.py
```diff
@@ -164,8 +165,8 @@ class ServerConfig(BaseModel):
     )
 
     integration: str = Field(
-        default=INTEGRATION_LOCAL,
-        description="The kind of integration to use. local|sagemaker",
+        default=None,
+        description="The kind of integration to use. local|sagemaker|openai",
```
mgoin, line 169: nit: could directly put INTEGRATIONS in the description
```suggestion
        description=f"The kind of integration to use. {INTEGRATIONS}",
```

mgoin (approved):
nice fix and cleanup, it works with my test, thanks!

mgoin (approved):

dbogunowicz (approved):

PR #1509, opened 2024-01-05
[Benchmark] Fix number of iterations reported

dsikka:
# Summary

- While using `benchmark_pipeline`, found a small bug in reported stats where we were incorrectly reporting the number of keys in the dictionary, not the number of iterations for which inference ran


dbogunowicz (approved):
Oh, pesky bug

mgoin (approved):

PR #1510, opened 2024-01-05
[Text Generation] Consolidate `TextGenerationPipeline` and `TextGenerationPipelineNoKVCache`

dbogunowicz:


bfineran (dismissed):

anmarques (commented):
There is a discrepancy in the interface between cached and non-cached models. For cached models it takes sequences, whereas for cached models it takes promt

dsikka (changes requested):
Is there anyway for us to consolidate some of the repeat code here? We set-up the model twice. We can just do it once, decide if kv_cache is present, and then pass the correct paths

src/deepsparse/pipeline.py
```diff
@@ -140,7 +140,6 @@ def create(cls, task: str, **kwargs) -> "Pipeline":
                     "provided task should be registered using the OperatorRegistry"
                 )
         except Exception:
-            _LOGGER.warning(f"Could not create v2 '{task}' pipeline, trying legacy")
```
dsikka, line 143: Where'd it go

dbogunowicz: blimey

dsikka (approved):
This LGTM. 

Can we just do a quick audit that the properties that were available before still work?
Specifically: model_path, tokenizer, input/output_schemas?
Also, can we confirm this PR didn't cause any issues: https://github.com/neuralmagic/deepsparse/pull/1523

If the above are true, can merge.

bfineran (approved):

dbogunowicz:
@dsikka all units/integrations tests pass with #1523 included.
I confirm that all the aforementioned properties are still present and functional in both "incarnations" of TextGenerationPipeline

PR #1520, opened 2024-01-09
[Move] Common Evaluation Modules to `Sparsezoo``

rahul-tuli:
# Evaluator Move

This PR moves a few modules from `deepsparse.evaluation` to
`sparsezoo.evaluation`

## Motivation and Context

The moved modules provide a common interface for evaluating models.
This interface can be leveraged and used by both `SparseML`
and `DeepSparse` to evaluate models.

Since `SparseZoo` is a common dependency for both repositories, it
makes sense to move these modules to `SparseZoo`.

## How Has This Been Tested?

The tests for moved modules in `deepsparse.evaluation` are moved to `sparsezoo.evaluation` and other deepsparse specific `deepsparse.evaluation` tests were updated to use the moved modules.

Both `deepsparse` and `sparsezoo` tests were run to ensure that the move did not break anything.

Also ran the following eval command:
```bash
$ deepsparse.eval --target zoo:mpt-7b-mpt_pretrain-base_quantized \                                                             (move-evaluator-to-sparsezoo|●1✚2…4)
                --dataset hellaswag \
                --dataset gsm8k \
                --integration lm-evaluation-harness \
                --limit 2
```

Note: This diff relies on sparsezoo diff https://github.com/neuralmagic/sparsezoo/pull/411

The failing tests should pass after sparsezoo side changes are merged!


src/deepsparse/evaluation/registry.py
```diff
@@ -17,23 +17,23 @@
...
 
 
-__all__ = ["EvaluationRegistry"]
+__all__ = ["DeepSparseEvaluationRegistry"]
 
 _LOGGER = logging.getLogger(__name__)
 
 
-class EvaluationRegistry(RegistryMixin):
+class DeepSparseEvaluationRegistry(EvaluationRegistry):
```
Satrat, line 28: What in this class is deepsparse specific? Could we move this to the SparseZoo base class?

Satrat (commented):
Looks good to me overall, see my comment on the DeepSparseEvaluationRegistry, was wondering if any of that can be moved to the SparseZoo base class

src/deepsparse/evaluation/registry.py
rahul-tuli: This class has a resolve method, that imports certain modules from deepsparse; I would like to keep importing deepsparse modules in deepsparse and not in sparsezoo unless there's a strong reasoning behind that; the idea being sparsezoo would just define the contract and contain common utilities whereas deepsparse and sparseml would add things relevant to just them.

In this case the partially_check_dependency method is coupled with the integrations. As a future step, we might add a small config file, mapping integration name --> their corresponding files, that could potentially remove the coupling and we can unify more code then?

tests/deepsparse/evaluation/test_registry.py
```diff
@@ -1,51 +0,0 @@
-# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
dbogunowicz, line 1: You are moving those tests to sparsezoo right?

src/deepsparse/evaluation/registry.py
dbogunowicz: I think @rahul-tuli's idea is good. So we have a "chain of inheritance":
1. `sparsezoo.RegistryMixin`
2. `sparsezoo.EvaluationRegistry`
3. `deepsparse.DeepSparseEvaluationRegistry` / `sparseml.SparseMLEvaluationRegistry`
right?

dbogunowicz (approved):

src/deepsparse/evaluation/registry.py
rahul-tuli: Yes!

tests/deepsparse/evaluation/test_registry.py
rahul-tuli: Yes they are there!

src/deepsparse/evaluation/cli.py
```diff
@@ -59,7 +59,7 @@
 ##########
 Example command for evaluating a quantized MPT model from SparseZoo using the Deepsparse Engine.
 The evaluation will be run using `lm-evaluation-harness` on `hellaswag` and `gsm8k` datasets:
-deepsparse.eval zoo:mpt-7b-mpt_pretrain-base_quantized \
+deepsparse.eval --target zoo:mpt-7b-mpt_pretrain-base_quantized \
```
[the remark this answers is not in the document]
rahul-tuli: I just fixed the usage command, not specifying `--target` in the current state of the cli throws an error.

bfineran:
base tests failing - LGTM otherwise

PR #1526, opened 2024-01-12
Add support for tokenized prompt input for CompletionRequest

mgoin:
Found through lm-evaluation-harness since for the OpenAI Completion interface they make requests using raw token_ids

OAI docs reference: https://platform.openai.com/docs/api-reference/completions
> prompt: The prompt(s) to generate completions for, encoded as a string, array of strings, array of tokens, or array of token arrays.

Example request during `gsm8k` evaluation:
```
{'model': 'hf:mgoin/llama-2-7b-gsm8k-pruned60-quant-ds', 'prompt': [[1, 894, 29901, 21828, 18577, 29871, 29929, 29900, 9814, 273, 21425, 322, 29871, 29946, 29900, 28145, 5697, 348, 3173, 393, 9814, 273, 21425, 29889, 1128, 1784, 18281, 947, 540, 8024, 3001, 29973, 13, 22550, 29901]], 'max_tokens': 256, 'stop': ['<|endoftext|>'], 'seed': 1234, 'temperature': 0.0}
```

Server command:
```
deepsparse.server --integration openai --task text-generation --model_path hf:mgoin/llama-2-7b-gsm8k-pruned60-quant-ds
```

Eval command (using this branch https://github.com/EleutherAI/lm-evaluation-harness/pull/1277):
```
lm_eval --model local-completions --model_args base_url=http://localhost:5543/v1,model=hf:mgoin/llama-2-7b-gsm8k-pruned60-quant-ds,tokenizer_backend=huggingface,tokenizer=mgoin/llama-2-7b-gsm8k-pruned60-quant-ds --tasks gsm8k --num_fewshot 0
```

### DIFF
--- a/src/deepsparse/server/openai_server.py
+++ b/src/deepsparse/server/openai_server.py
@@ -232,7 +232,28 @@ class OpenAIServer(Server):
 
             request_id = f"cmpl-{random_uuid()}"
             created_time = int(time.time())
-            prompt = request.prompt
+
+            if isinstance(request.prompt, list):
+                if len(request.prompt) == 0:
+                    return create_error_response(
+                        HTTPStatus.BAD_REQUEST, "please provide at least one prompt"
+                    )
+                first_element = request.prompt[0]
+                if isinstance(first_element, int):
+                    # There is just a single tokenized prompt
+                    prompt = pipeline.tokenizer.decode(request.prompt)
+                elif isinstance(first_element, (str, list)):
+                    if len(request.prompt) > 1:
+                        return create_error_response(
+                            HTTPStatus.BAD_REQUEST,
+                            "multiple prompts in a batch is not currently supported",
+                        )
+                    if isinstance(first_element[0], int):
+                        prompt = pipeline.tokenizer.decode(first_element)
+                    else:
+                        prompt = first_element
+            else:
+                prompt = request.prompt
 
             try:
                 sampling_params = dict(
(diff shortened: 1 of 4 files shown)


dbogunowicz (approved):

src/deepsparse/server/openai_server.py
```diff
@@ -232,7 +232,28 @@ async def create_completion(raw_request: Request):
...
+                    # There is just a single tokenized prompt
+                    prompt = pipeline.tokenizer.decode(request.prompt)
+                elif isinstance(first_element, (str, list)):
+                    if len(request.prompt) > 1:
+                        return create_error_response(
+                            HTTPStatus.BAD_REQUEST,
+                            "multiple prompts in a batch is not currently supported",
+                        )
+                    if isinstance(first_element[0], int):
+                        prompt = pipeline.tokenizer.decode(first_element)
```
bfineran, line 252: how does decode work when decoding a single token at a time? it won't be 1:1 with the plaintext prompt right?

mgoin: this is decoding all the tokens at once - first_element is a `List[int]` after this conditional checks

dsikka: any consideration for special characters? I believe we have a specific param in the pipeline

dsikka (changes requested):
please add a test for this input case

mgoin:
> please add a test for this input case

very good call, missed that file

src/deepsparse/server/openai_server.py
mgoin: i think we do want to preserve special tokens, and `skip_special_tokens` is False by default

dsikka (approved):

bfineran (approved):

PR #1528, opened 2024-01-15
[TextGeneration] Update defaults to always start with deepsparse defaults when using the `generation_config` arg

dsikka:
# Summary
- Previously, if the user provided anything for the `generation_config` argument, we always returned a `GenerationConfig` object for this case
- Now if the user provides the `GenerationConfig` object, that is the only case where we return the object without using deepsparse defaults
- If the user provides a dictionary, we create a GenerationConfig object, override with deepsparse defaults, and then set any attributes provided by the user in the dictionary
- This is aided through the conversion of the defaults into a dataclass
- All other workflows remain the same (i.e providing `generation_kwargs` or cli arguments which have always used deepsparse defaults)

Also takes care of this discussion: https://app.asana.com/0/1205229323407165/1206280624568198/f

### DIFF
--- a/src/deepsparse/transformers/schemas/text_generation_schemas.py
+++ b/src/deepsparse/transformers/schemas/text_generation_schemas.py
@@ -14,6 +14,7 @@
 
 import datetime
 import pathlib
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import Any, Callable, Dict, List, Optional, Sequence, Union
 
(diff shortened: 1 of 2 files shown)


dbogunowicz (approved):

src/deepsparse/transformers/schemas/text_generation_schemas.py
```diff
@@ -22,20 +23,21 @@
 
 
 # Based off of https://huggingface.co/docs/transformers/main_classes/text_generation#transformers.GenerationConfig # noqa E501
+@dataclass
```
rahul-tuli, line 26: Any reason to use data classes over Pydantic BaseModels? Specially because we're already using BaseModel throughout the codebase?

src/deepsparse/transformers/schemas/text_generation_schemas.py
```diff
@@ -22,20 +23,21 @@
...
     # Parameters that control the generation strategy used
-    do_sample = False
+    do_sample: bool = field(default=False)
     # Parameters for manipulation of the model output logits
-    temperature = 1.0
-    top_k = 50
-    top_p = 1.0
-    repetition_penalty = 1.0
+    temperature: float = field(default=1.0)
+    top_k: int = field(default=50)
```
rahul-tuli, line 35: Pydantic Fields would go in nicely here, specially cause we can provide a description of what each argument is

rahul-tuli (changes requested):

src/deepsparse/transformers/schemas/text_generation_schemas.py
dsikka: doesnt seem like it needs to be a pydantic model if only used internally and just holds default data

src/deepsparse/transformers/schemas/text_generation_schemas.py
dsikka: doesnt seem like it needs to be a pydantic model if only used internally and just holds default data

bfineran (approved):

mgoin (approved):

PR #1537, opened 2024-01-17
[V2 Logger]  factory

horheynm:
Screenshot 2024-01-17 at 1 53 24 PM

PR #1540, opened 2024-01-17
[V2 Logger] Filters

horheynm:
Screenshot 2024-01-17 at 1 49 11 PM

Please review https://github.com/neuralmagic/deepsparse/pull/1533 and https://github.com/neuralmagic/deepsparse/pull/1537 first 

# Desctiption:
This is the juiciest component of the loggers. The PR is responsible for filtering incoming value to log or ignore based on the rules. 

Rule 1. Filtering by tag. The incoming `log(value, tag, log_type)`, will be filtered if the tag from arg does not match with any of the tag from the config (inside config.yaml) inside the root logger. The tag in the config can be a regex.
Rule 2. Filtering by the number of encounters of {log_type}.{tag}.{func}  for each system, metric, performance root logger after rule 1 passes. The incoming `log(value, tag, log_type)`, may be called many times. We dont want to log every logger call. Based on the `freq` value in the config file, ignore any call number that is not a multiple of `freq`.  

# Usage:
Check tests


src/deepsparse/loggers_v2/filters/frequency_filter.py
```diff
@@ -0,0 +1,49 @@
...
+        stub_frequency = f"{tag}.{func}"
+        with self._lock:
+            if stub not in self.counter:
+                self.counter[stub] = 0
+
+            for key, value in self.frequency.items():
+
+                if is_match_found(key, stub_frequency):
+                    frequency = value
+                    self.counter[stub] = (self.counter[stub] + 1) % frequency
```
horheynm, line 46: avoid creating big int type.

tests/logger_v2/config.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2021 - present / Neuralmagic, Inc. All Rights Reserved.
```
horheynm, line 1: delete

src/deepsparse/loggers_v2/filters/frequency_filter.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class FrequencyFilter:
+    def __init__(self):
+        self._lock = Lock()
+        self.frequency = {}
+        self.counter = {}
+
+    def add_template_to_frequency(self, tag: str, func: str, rate: int = 1):
+        with self._lock:
```
bfineran, line 28: depending on how many logs we end up emitting and at what scale, we might want to find a way to either relax the lock, or have a different lock for keys if this ends up becoming blocking

src/deepsparse/loggers_v2/filters/frequency_filter.py
```diff
@@ -0,0 +1,48 @@
...
+            self.frequency[f".*{tag}.*{func}.*"] = rate
+
+    def should_execute_on_frequency(
+        self, tag: str, log_type: str, func: Callable
+    ) -> bool:
+
+        stub = f"{log_type}.{tag}.{func}"
+        stub_frequency = f"{tag}.{func}"
+        with self._lock:
```
bfineran, line 37: we should only lock when overwriting an object we want to protect. locking during the loop / `is_match_found` holds the lock longer than necessary

src/deepsparse/loggers_v2/filters/frequency_filter.py
```diff
@@ -0,0 +1,48 @@
...
+            if stub not in self.counter:
+                self.counter[stub] = 0
```
bfineran, line 39: why not initialize in `add_template_to_frequency` or use a `defaultdict`?

src/deepsparse/loggers_v2/filters/pattern.py
```diff
@@ -0,0 +1,27 @@
...
+# software distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+from typing import Optional
+
+
+def is_match_found(
```
bfineran, line 19: note that treating all targets/patterns as pure regex may have unintended side effects

for example, `.` is a wildcard, so `fooXbar` will also match with`foo.bar`

we might want to think through if we want to protect regex use a bit more

bfineran (approved):

PR #1542, opened 2024-01-17
[V2 Logger] root logger

horheynm:
Screenshot 2024-01-18 at 11 31 23 AM

Three types of root loggers: system, performance and metric. All root loggers have shared leaf loggers (bc they are singleton). Each root loggers have their own FrequencyFilter. Calling `.log(...)` will call leaf loggers after the filtering rules passes.

### DIFF
--- a/src/deepsparse/loggers_v2/root_logger.py
+++ b/src/deepsparse/loggers_v2/root_logger.py
@@ -15,7 +15,7 @@
 
 from collections import defaultdict
 from enum import Enum
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from deepsparse.loggers_v2.filters.frequency_filter import FrequencyFilter
 from deepsparse.loggers_v2.filters.pattern import (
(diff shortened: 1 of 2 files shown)


src/deepsparse/loggers_v2/root_logger.py
```diff
@@ -0,0 +1,127 @@
...
+from .utils import import_from_registry
+
+
+class LogType(Enum):
+    SYSTEM = "SYSTEM"
+    PERFORMANCE = "PERFORMANCE"
+    METRIC = "METRIC"
+
+
+class RootLogger(FrequencyFilter):
```
bfineran, line 32: what do we get out of having `FrequencyFilter` as a parent class instead of a property that is used as a helper?

horheynm: Depends on how deep the question is. 
Its a class bc you need a state to keep in track of the counter per tag.func. If you mean just a @property in root loggers, yes we can do that. 
If you mean to get rid of FF and combine to RootLogger, we can do that, but better to keep logic isolated

src/deepsparse/loggers_v2/root_logger.py
```diff
@@ -0,0 +1,227 @@
...
+        leaf loggers
+
+        :param value: Any value to log, may be dimentionally reduced by func
+        :param log_type: String representing the root logger level
+        :param tag: Candidate id that will be used to filter out only the wanted log
+        :param capture: The property or dict key to record if match exists. If set to None
+            no capture filter will be applied even if set in config
+
+        """
+        for tag_from_config, tag_run_args in self.run_args.items():
```
bfineran, line 115: let's work on breaking this down into helper functions

PR #1554, opened 2024-01-23
Fix documentation about TextGeneration

luis5tb:
TextGeneration takes model_path as parameter instead of model

### DIFF
--- a/docs/llms/text-generation-pipeline.md
+++ b/docs/llms/text-generation-pipeline.md
@@ -42,7 +42,7 @@ from deepsparse import TextGeneration
 
 # construct a pipeline
 model_path = "zoo:mpt-7b-dolly_mpt_pretrain-pruned50_quantized"
-pipeline = TextGeneration(model=model_path)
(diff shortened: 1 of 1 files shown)


mgoin:
> TextGeneration takes model_path as parameter instead of model

Hey @luis5tb `model=` should work for nightly and/or the upcoming 1.7, are you using the 1.6 release possibly?

luis5tb:
> > TextGeneration takes model_path as parameter instead of model
> 
> Hey @luis5tb `model=` should work for nightly and/or the upcoming 1.7, are you using the 1.6 release possibly?

Yep using 1.6.1, following that documentation I hit it and saw that it was solved by adding model_path. Perhaps the code should take both then to keep backwards compatibility or add a note in the docs

luis5tb:
Should I close this then?

mgoin:
@luis5tb the nightly should both accept `model_path` and `model`, we just added `model` as a new alias that we'll be using going forward so backwards compatibility is good. it didn't make it into 1.6.1 though

PR #1556, opened 2024-01-23
sklearn patch

andy-neuma:
patch import of `sklearn`


mgoin:
sklearn should be removed now

PR #1562, opened 2024-01-25
[TextGeneration] Split up `prep_for_generation` operator, handle edge cases, handle kv_cache full during prefill

dsikka:
# Summary

**A series of improvements**: 

- Split up the `prep_for_generation` operator such that we aren't setting up the tokenizer and generating tokens in the same operator, outside of the `generation loop`
- With this split, make sure we properly handle edge cases, such as `max_length=1, max_new_tokens=0, max_new_tokens=1`. Also added the condition such that `max_length` has to be greater than 0 or we throw a `ValueError`
- With splitting this up, we no longer have to import the `process_output` operator in the `prep_for_generation` operator and can remove the custom streaming code in the operator . Also, the non-kv cache and kv_cache pipelines now both share the same `output_schema`, simplifying generation logic
- Also added a condition to check kv_cache capacity during prompt inference and exit if the capacity is reached before generation

# Testing
- Added new tests to validate the edge cases
- Added a test to check if we throw an error when the cache capacity is filled during prefill
- Updated the non-kv cache unit test to also compare logits
- All other test cases pass

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation/generate_new_token.py
+++ b/src/deepsparse/transformers/pipelines/text_generation/generate_new_token.py
@@ -52,20 +52,29 @@ class GenerateNewTokenOperator(Operator):
             if isinstance(inp, NLEngineOutputs)
             else inp.prompt_logits
         )
-        kv_cache = inp.kv_cache if isinstance(inp, NLEngineOutputs) else None
+        kv_cache = (
+            inp.kv_cache
+            if not isinstance(inp, PromptLogitsNoKVCacheInference)
+            else None
+        )
+
+        max_tokens = inference_state.current_state.get("max_tokens")
+        length_finish_reason = inference_state.current_state.get("length_finish_reason")
+        generated_tokens = inference_state.current_state.get("generated_tokens")
+        num_generated_tokens = len(generated_tokens)
 
         token_generator = inference_state.current_state.get("token_generator")
         token = token_generator.generate(logits=logits[0, -1, :])
         finish_reason = None
-
-        callback = inference_state.current_state.get("callback")
-        stop = inference_state.current_state.get("stop")
 
         if (
             kv_cache is not None
             and kv_cache.total_num_processed_tokens >= kv_cache.capacity
         ):
             finish_reason = FinishReason.CAPACITY
+
+        callback = inference_state.current_state.get("callback")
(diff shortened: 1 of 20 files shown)


src/deepsparse/transformers/pipelines/text_generation/generate_new_token.py
```diff
@@ -84,9 +88,8 @@ def run(
             )
             finish_reason = FinishReason.CALLBACK
 
-        max_tokens = inference_state.current_state.get("max_tokens")
-        if len(inference_state.current_state.get("generated_tokens")) + 1 >= max_tokens:
-            finish_reason = inference_state.current_state.get("length_finish_reason")
+        if num_generated_tokens + 1 == max_tokens:
```
dbogunowicz, line 91: I'd add a short comment explaining the logic behind `+1`

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -29,7 +29,7 @@ def pipeline():
     return Pipeline.create(
         task="text_generation",
         model_path="hf:mgoin/TinyStories-1M-deepsparse",
-        engine_type="onnxruntime",
```
dbogunowicz, line 32: in general using onnxruntime is preferred in tests, since it is faster to compile.

tests/deepsparse/transformers/pipelines/test_text_generation.py
```diff
@@ -282,3 +284,45 @@ def test_streaming_non_streaming_generate_same_tokens(pipeline, prompt):
         tokens.append(g.generations[0].text)
     output_2 = "".join(tokens)
     assert output_1 == output_2
+
+
+def test_edge_cases(pipeline, prompt):
+    output = pipeline(prompt=prompt, max_length=1, output_scores=True)
```
dbogunowicz, line 290: please add small comments to each test case to flag which type of behavior it tests

dbogunowicz (approved):

tests/deepsparse/transformers/pipelines/test_text_generation.py
dsikka: ah I changed it locally and forgot to change it back. Good catch!

dbogunowicz (approved):

bfineran (approved):

PR #1566, opened 2024-01-29
Debug pipeline

ProExpertProg:
This is currently a hack but it would be great to get a version of this into production so that we can use debug_analysis on the pipeline and pass real inputs.


jeanniefinks:

PR #1570, opened 2024-01-30
Better legacy failure warning

ProExpertProg:
In case someone fails to create the V2 pipeline, I think it's helpful to print the error.

### DIFF
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -140,8 +140,7 @@ class Pipeline(Operator):
                     "provided task should be registered using the OperatorRegistry"
                 )
         except Exception as e:
-            _LOGGER.warning(f"Could not create v2 '{task}' pipeline, with error: {e}")
-            _LOGGER.warning(f"Attempting to create the legacy pipeline")
+            _LOGGER.warning(f"Could not create v2 '{task}' pipeline, attempting to create the legacy pipeline. Reason: {e}")
             from deepsparse.legacy import Pipeline
 
             pipeline = Pipeline.create(task=task, **kwargs)
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -139,8 +139,9 @@ class Pipeline(Operator):
                     "Pipeline was not created for the given task. The "
                     "provided task should be registered using the OperatorRegistry"
                 )
-        except Exception:
-            _LOGGER.warning(f"Could not create v2 '{task}' pipeline, trying legacy")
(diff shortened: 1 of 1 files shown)


horheynm (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -139,8 +139,9 @@ def create(cls, task: str, **kwargs) -> "Pipeline":
                     "Pipeline was not created for the given task. The "
                     "provided task should be registered using the OperatorRegistry"
                 )
-        except Exception:
-            _LOGGER.warning(f"Could not create v2 '{task}' pipeline, trying legacy")
+        except Exception as e:
+            _LOGGER.warning(f"Could not create v2 '{task}' pipeline, with error: {e}")
+            _LOGGER.warning(f"Attempting to create the legacy pipeline")
```
bfineran, line 144: could we combine these to one string so only one message is logged?

rahul-tuli (dismissed):

src/deepsparse/pipeline.py
ProExpertProg: Yeah I just figured the error might be long. But I'll put it at the end, something like:
```python
_LOGGER.warning(f"Could not create v2 '{task}' pipeline, attempting to create the legacy pipeline. V2 pipeline error: {e}")
```

dbogunowicz: You can just do:

```
_LOGGER.warning(f"Could not create v2 '{task}' pipeline, with error: {e}. "
                                 "Attempting to create the legacy pipeline")
```

dbogunowicz (dismissed):

src/deepsparse/pipeline.py
rahul-tuli: I think Luka meant the error will be long so the sentence might look weird in the console

dsikka:
You can't have the word `error` in your warning

src/deepsparse/pipeline.py
ProExpertProg: Yeah just in case

ProExpertProg:
Oh, I see. What words are allowed? Can I say `"failed with"`?

dsikka:
@ProExpertProg you're currently failing this test: https://github.com/neuralmagic/deepsparse/blob/e29211c95b75ff19e0bf0c65a26b6044ba51bf80/tests/test_pipeline_benchmark.py#L89

Which checks for `error` and `fail`

jeanniefinks:

PR #1583, opened 2024-02-05
[TextGeneration] Add helper function to parse model path from args

dsikka:
# Summary
- Add a quick helper function which supports non-kwarg for model_path when using the `TextGeneration` pathway
- For this ticket: https://app.asana.com/0/1206109050183159/1206524727025314/f

The following now works:

```python
from deepsparse import TextGeneration
model_path = "hf:mgoin/TinyStories-1M-ds"

pipe = TextGeneration(model_path)

```

This still works:

```python

model_path  = TextGeneration(model_path=model_path)
```

### DIFF
--- a/src/deepsparse/pipeline.py
+++ b/src/deepsparse/pipeline.py
@@ -702,6 +702,7 @@ def text_generation_pipeline(*args, **kwargs) -> "Pipeline":
     :return: text generation pipeline with the given args and
         kwargs passed to Pipeline.create
     """
+    args, kwargs = _check_model_path_arg(*args, **kwargs)
     return Pipeline.create("text_generation", *args, **kwargs)
 
 
@@ -710,6 +711,7 @@ def code_generation_pipeline(*args, **kwargs) -> "Pipeline":
     :return: text generation pipeline with the given args and
         kwargs passed to Pipeline.create
     """
+    args, kwargs = _check_model_path_arg(*args, **kwargs)
     return Pipeline.create("code_generation", *args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


bfineran (dismissed):

src/deepsparse/pipeline.py
```diff
@@ -802,3 +805,14 @@ def zero_shot_text_classification_pipeline(*args, **kwargs) -> "Pipeline":
...
+
+
+def _check_model_path_arg(*args, **kwargs):
+    if args:
+        if len(args) > 1 or "model_path" in kwargs or "model" in kwargs:
+            raise ValueError(
+                "Only the model path can be provided as a non-kwarg argument"
+            )
+        kwargs["model_path"] = args[0]
+        args = args[1:]
```
rahul-tuli, line 817: Won't args be an empty list here?

src/deepsparse/pipeline.py
```diff
@@ -802,3 +805,14 @@ def zero_shot_text_classification_pipeline(*args, **kwargs) -> "Pipeline":
     is returned depends on the value of the passed model_scheme argument.
     """
     return Pipeline.create("zero_shot_text_classification", *args, **kwargs)
+
+
+def _check_model_path_arg(*args, **kwargs):
+    if args:
+        if len(args) > 1 or "model_path" in kwargs or "model" in kwargs:
```
rahul-tuli, line 812: why not have the functions accept `model_path` as an arg directly? then python would raise an error automatically if kwargs also have model_path, we'll still have to check for `model` in kwargs though? Thoughts?


for example changing:

```python
def chat_pipeline(*args, **kwargs) -> "Pipeline":
```

To
```python
def chat_pipeline(model_path, **kwargs) -> "Pipeline":
```

dsikka: Can't just have `model_path` as we also supprt `model` so seems cleaner to just accept args and parse?

src/deepsparse/pipeline.py
dsikka: Yah good catch. Shouldnt be any other args.

mgoin (approved):

PR #1584, opened 2024-02-06
[server] Add `model` argument to server cli

dsikka:
# Summary:
- With our current `click` set-up, to get the use case described in this ticket (shown below): https://app.asana.com/0/1206109050183159/1206524727025314/f 

```bash
deepsparse.server \
 "zoo:llama2-7b-ultrachat200k_llama2_pretrain-pruned50_quantized" \
 --integration openai
```

# PR Update 
This PR allows us to get the following cli command:

```bash
deepsparse.server --integration openai "hf:mgoin/TinyStories-1M-ds"
```

We can run the following commands with the current set-up:
```bash
deepsparse.server --task text_generation --model_path "hf:mgoin/TinyStories-1M-ds"
deepsparse.server --model_path "hf:mgoin/TinyStories-1M-ds" --integration openai
deepsparse.server --task text_generation "hf:mgoin/TinyStories-1M-ds"
deepsparse.server --intergration openai --task text_generation  "hf:mgoin/TinyStories-1M-ds"
deepsparse.server --config_file ~/debugging/sample_config.yaml
```

Caveats (@bfineran):
- This also does not let us run the cli command by providing the model path first (i.e will not directly follow what is shown in the ticket/UX docs; the integration or task option would have to be first)

Shoutout to @rahul-tuli  for his click knowledge and help

### DIFF
--- a/src/deepsparse/server/cli.py
+++ b/src/deepsparse/server/cli.py
@@ -79,7 +79,17 @@ HOT_RELOAD_OPTION = click.option(
     ),
 )
 
-MODEL_OPTION = click.argument("model_path", type=str, default="default")
+MODEL_ARG = click.argument("model", type=str, default=None, required=False)
+MODEL_OPTION = click.option(
+    "--model_path",
+    type=str,
+    default="default",
+    help=(
+        "The path to a model.onnx file, a model folder containing the model.onnx "
+        "and supporting files, or a SparseZoo model stub. "
+        "If not specified, the default model for the task is used."
+    ),
+)
 
 BATCH_OPTION = click.option(
     "--batch_size",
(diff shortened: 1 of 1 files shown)


bfineran (dismissed):

mgoin:
is this a breaking change if we were using `--model_path` first? this seems fairly important for all server flows and we should hopefully be able to deprecate rather than removing

dsikka:
@mgoin yes. If we make `model_path` an argument to support the UX docs, we can't use it as an option as well `---model_path`

We could add a separate model path entry point

bfineran:
> @mgoin yes. If we make `model_path` an argument to support the UX docs, we can't use it as an option as well `---model_path`
> 
> 
> 
> We could add a separate model path entry point but click can't support both AFAIK

@dsikka per @markurtz lets add a model_path kwarg back in (can rename the positional arg to something else) and allow it to override the positional if given. (Would need to make the positional op timbale in this case I guess)

bfineran (approved):

mgoin (approved):

PR #1585, opened 2024-02-06
[Continuous Batching][Text Generation] Add back continuous batching tests

dsikka:
# Summary
- Temporarily disabled due to potential source for jenkins seg fault
- As those tests are now passing/not running into seg fault errors, update to add back these tests


bfineran (approved):

mgoin (approved):
nice

PR #1605, opened 2024-02-13
[TextGeneration] Update doc to be reflective of updated deepsparse defaults

dsikka:
# Summary
- Update our docs to be reflective of some of the defaults which have been updated for text generation 
- Not a requirement for 1.7 but for completeness, should be up-to-date

New table:
Screenshot 2024-02-13 at 4 00 49 PM


mgoin (approved):
good call, thanks!

PR #1624, opened 2024-02-26
Pass arguments for pipeline creation properly to deepsparse.evaluate

dbogunowicz:


bfineran (approved):

src/deepsparse/evaluation/utils.py
```diff
@@ -202,11 +202,13 @@ def create_pipeline(
     :return: The initialized pipeline
```
horheynm, line 202: Add a note saying that kwargs dict will be mutated

src/deepsparse/evaluation/evaluator.py
```diff
@@ -56,9 +56,10 @@ def evaluate(
 
     # if target is a string, turn it into an appropriate pipeline
     # otherwise assume it is a pipeline
-    pipeline = (
-        create_pipeline(model, engine_type) if isinstance(model, (Path, str)) else model
-    )
+    if isinstance(model, (Path, str)):
+        pipeline, kwargs = create_pipeline(model, engine_type, **kwargs)
```
horheynm, line 60: nit: The rules for passing in kwargs here in create_pipeline and later in eval_integration is pretty loose.
Instead of mutating i would have two dicts instead of using one and mutating it

dbogunowicz: @horheynm you are right. For the reasons that we will discuss internally, the design here is not the best and some refactoring will be necessary. But for the sake of landing it for 1.7, i'd ask you to turn a blind eye for now.

horheynm (approved):

bfineran (approved):

PR #1635, opened 2024-03-14
[TextGeneration] Fix llama tokenizer

dsikka:
Tested code:

```python

import deepsparse

MODEL_ID = "hf:nm-testing/llama2-7B-sparse70-retrained-ultrachat200k-pruned70-smoothquant-ds"
#MODEL_ID = "zoo:mistral-7b-ultrachat200k_mistral_pretrain-pruned40_quantized"

pipe = deepsparse.Pipeline.create(
    task="text-generation",
    model_path=MODEL_ID,
    sequence_length=512,
    prompt_sequence_length=16,
)

message = "Once upon a time"

conversation = []
conversation.append({"role": "user", "content": message})
formatted_conversation = pipe.tokenizer.apply_chat_template(
    conversation, tokenize=False, add_generation_prompt=True
)

generation_config = {
    "max_new_tokens": 100,
}

inference = pipe(
    sequences=formatted_conversation,
    generation_config=generation_config,
    streaming=True,
)

for token in inference:
    print(token.generations[0].text, end="")


```

Output:

```

There was a time when the world was a different place. A time when people were more accepting of each other and didn't judge based on race, religion, or gender. A time when kindness and compassion were the norm, and hate and prejudice were unheard of.

But then something changed. The world became more divided, and people started to see each other through a

```

### DIFF
--- a/src/deepsparse/transformers/pipelines/text_generation/prep_for_generation.py
+++ b/src/deepsparse/transformers/pipelines/text_generation/prep_for_generation.py
@@ -101,6 +101,7 @@ class PrepareGeneration(Operator):
             else [],
             "finished_reason": [],
             "token_generator": token_generator,
+            "past_tokens_queue": copy.copy(tokens),
         }
 
         if kv_cache is None:
--- a/src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
+++ b/src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
@@ -97,7 +97,7 @@ class ProcessOutputs(Operator):
         if isinstance(
             self.tokenizer,
             (transformers.LlamaTokenizer, transformers.LlamaTokenizerFast),
-        ):
+        ) and inference_state.current_state.get("streaming"):
             past_tokens_queue = inference_state.current_state.get("past_tokens_queue")
             sequences = self._generate_streamed_text_from_past_tokens(
                 generated_tokens, past_tokens_queue
--- a/src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
+++ b/src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import datetime
-from typing import Optional
+from typing import List, Optional
 
 import numpy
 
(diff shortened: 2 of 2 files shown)


bfineran (approved):

src/deepsparse/transformers/pipelines/text_generation/prep_for_generation.py
```diff
@@ -101,6 +101,7 @@ def run(
             else [],
             "finished_reason": [],
             "token_generator": token_generator,
+            "past_tokens_queue": copy.copy(tokens),
```
mgoin, line 104: is this just an initial copy of the prompt tokens, or does it happen for each iteration?

src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
```diff
@@ -64,9 +91,24 @@ def run(
...
+        # TODO: make streaming a conditional input to this operator. using inference
+        # state is a quick fix.
+        if isinstance(
+            self.tokenizer,
+            (transformers.LlamaTokenizer, transformers.LlamaTokenizerFast),
+        ) and inference_state.current_state.get("streaming"):
+            past_tokens_queue = inference_state.current_state.get("past_tokens_queue")
+            sequences = self._generate_streamed_text_from_past_tokens(
+                generated_tokens, past_tokens_queue
+            )
```
mgoin, line 107: does the queue need to contain all of the previous tokens? i think vLLM or TGI only uses a window of several tokens - reference here https://github.com/huggingface/text-generation-inference/blob/v0.9.4/server/text_generation_server/models/model.py#L62C9-L62C15

src/deepsparse/transformers/pipelines/text_generation/prep_for_generation.py
dsikka: Start with the prompt tokens. Update to add the new token and remove the first token in the queue. Basically a sliding window

src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
dsikka: We also only use a window of several tokens

mgoin (approved):
Thanks for answering my questions, nice implementation

src/deepsparse/transformers/pipelines/text_generation/process_outputs.py
dbogunowicz: @dsikka is right

dbogunowicz (commented):
A test would be nice to have, but I guess the priority is to land this asap.

PR #1664, opened 2024-08-28
Add more detail to yolo bounding boxes documentation

### LINKED ISSUE #1663: What is the format of the boxes returned from a YOLO Pipeline?
It's XYXY top left,bottom right. I guess the negative values are just the model predicting that the object extends outside the image. Leaving this open because it's still a problem that this isn't documented well.

~~Is it xyxy? xywh? If xywh, is the xy the center or top-left? I ran a yolo pipeline on an image and there are negative numbers in the bounding box ouput. Are the coordinates relative to the image center or something?~~
image

No info to be found in any of these places that would make sense
- YOLOOutput class source code
image
- YOLOPipeline class source code
- user guide
- object detection use case guide (https://github.com/neuralmagic/deepsparse/blob/main/docs/use-cases/cv/object-detection-yolov5.md)

~~I would be more than happy to get a "RTFM" response if I could just get a link to where this is documented.~~

EricPedley:
Expands the description to include info about the format and units.

Closes #1663

### DIFF
--- a/src/deepsparse/yolo/schemas.py
+++ b/src/deepsparse/yolo/schemas.py
@@ -101,7 +101,7 @@ class YOLOOutput(BaseModel):
     """
 
     boxes: List[List[List[float]]] = Field(
-        description="List of bounding boxes, one for each prediction"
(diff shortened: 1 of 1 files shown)


jeanniefinks:

